PR #11, opened 2021-02-15
Rename module and main branch

SimonHeybrock:
- scipp-neutron -> scippneutron
- master -> main


SimonHeybrock:
Merging, so I can setup Azure, which was apparently not done before.

PR #12, opened 2021-02-19
Setup CI and start populating from scipp/neutron

SimonHeybrock:


.azure-pipelines/code_quality_checks.yml
```diff
@@ -1,29 +0,0 @@
-stages:
-  - stage: 'code_quality'
-    displayName: 'Code Quality Checks'
-
-    jobs:
-      - job: 'python_checks'
```
OwenArnold, line 6: Code quality checks only include python parts. You want to take the clang-format etc parts too.

.azure-pipelines/main.yml
```diff
@@ -4,32 +4,128 @@ trigger:
...
+  - template: templates/code_quality_checks.yml
+
+  - stage: 'build_and_test'
+    displayName: 'Conda Package Build and Unit Tests'
 
     jobs:
-      - job: 'python_test'
-        displayName: 'Python tests'
-        timeoutInMinutes: 20
+      - job: 'linux'
```
OwenArnold, line 13: FYI, we should probably use the Matrix option and fold the linux and osx jobs together. FYI, see how I did this here: https://github.com/scipp/mantid_framework_conda_recipe/blob/master/.azure-pipelines/templates/build_and_package.yml#L15-L20

.azure-pipelines/main.yml
```diff
@@ -4,32 +4,128 @@ trigger:
...
+              echo "##vso[task.setvariable variable=conda_dir]$CONDA"
+            displayName: 'Configure Conda directories'
+          - bash: |
+              set -ex
+              conda --version
+              conda install --yes anaconda-client conda-build
+              conda config --set always_yes yes --set changeps1 no
+            displayName: 'Conda configuration'
+          - bash: |
+              conda build --user scipp --token "$ANACONDA_TOKEN" --channel conda-forge --label dev --python="$PYTHON_VERSION" ./conda
```
OwenArnold, line 38: To be clear you are providing CI to create a dev package. Presumably there would be a similar Release pipeline made at some later time?

OwenArnold, line 38: Does this line actually work? You don't specify the scipp or scipp/label/dev channel as part of the build.

.azure-pipelines/templates/test_can_install_and_import.yml
```diff
@@ -0,0 +1,141 @@
+jobs:
```
OwenArnold, line 1: Isn't this redundant. you do this already in the `meta.yaml` file? Note you can provide your own `run_test.sh` or `run_test.py` if you want to do more. For example https://github.com/scipp/mantid_framework_conda_recipe/blob/master/run_test.sh

.azure-pipelines/main.yml
SimonHeybrock: Not sure why Dan did not set it up in this way?

.azure-pipelines/main.yml
SimonHeybrock: See `release.yml`.

SimonHeybrock: In what sense? There is `--label dev`?

.azure-pipelines/templates/test_can_install_and_import.yml
SimonHeybrock: I don't know, I just copied what Dan set up.

.azure-pipelines/main.yml
OwenArnold: That's the label applied to the built/published package. But `conda-build` is not aware of the need to use the `scipp` channel to fetch the scipp dependency I think. I haven't seen the logs, so might be wrong about this.

.azure-pipelines/templates/test_can_install_and_import.yml
OwenArnold: I don't think you need it if you have the imports and requirements set in the `test` section of your `meta.yaml` that would be doing the exact equivalent of what you have here I think with far fewer lines of code.

.azure-pipelines/main.yml
SimonHeybrock: You are probably right, I had assumed the channel would be set somewhere in the `conda/` folder, I'll try adding it.

.azure-pipelines/templates/test_can_install_and_import.yml
OwenArnold: As a side note. We should probably utilise this in upstream `scipp` now that we have these changes https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#downstream-tests ?

SimonHeybrock: @DanNixon Do you remember why you added this? Did the `test` section not cover everything?

.azure-pipelines/main.yml
OwenArnold: Note that you may want different settings between your release pipeline and your master/pr pipeline in this regard.

.azure-pipelines/main.yml
```diff
@@ -4,32 +4,128 @@ trigger:
 pr: none
 
 stages:
-  - template: code_quality_checks.yml
-  - stage: 'test'
-    displayName: 'Python tests'
+  - template: templates/code_quality_checks.yml
```
OwenArnold, line 7: For both master and release pipelines we want to trigger that these downstream project pipelines of `scipp` are run when those pipelines succeed. Otherwise our scipp-neutron packages will be stale. See https://github.com/scipp/ess/blob/master/.azure-pipelines/master.yml#L6-L11 for the `scipp->ess master pipeline link`

SimonHeybrock: You may be right, but I am not so sure about this any more. Provided that the `scipp` packages are compatible, there should be no need to rebuild the `scippneutron` package? Only when API/ABI are broken we need to rebuild?

.azure-pipelines/main.yml
SimonHeybrock: Done (I think).

.azure-pipelines/main.yml
OwenArnold: Correct. Only if we break API/ABI, but if we accidentally do, then things will break in a really nasty way. Better to find out sooner by triggering the downstream project pipeline. We may even want to start including the status badge for the scippneutron pipelines in the scipp github `README` itself to flag this up.

.azure-pipelines/templates/test_can_install_and_import.yml
SimonHeybrock: > As a side note. We should probably utilise this in upstream `scipp` now that we have these changes https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#downstream-tests ?

I think this may be an answer to the other discussion (https://github.com/scipp/scippneutron/pull/12#discussion_r580927270) regarding whether we should trigger from upsream (scipp)? If we add downstream tests in scipp we would simply ensure that API/ABI required by `scippneutron` is not broken, rather than rebuilding `scippneutron` all the time.

.azure-pipelines/main.yml
SimonHeybrock: Wouldn't this be caught by the conda `downstream` tests that you linked above?

OwenArnold: > Wouldn't this be caught by the conda downstream tests that you linked above?
I haven't used that option, but yes if I understand correctly. we should also list `ess` in there as we don't want to break that event though there's only API considerations there. I think we should try and do that as soon as this pipeline is live (changes made in scipp repo of course).

.azure-pipelines/main.yml
```diff
@@ -4,32 +4,128 @@ trigger:
...
+            displayName: 'Package build'
+          - task: PublishBuildArtifacts@1
+            inputs:
+              PathtoPublish: "$(conda_dir)/conda-bld/linux-64"
+              ArtifactName: 'linux-64'
+            displayName: 'Publish Conda package artefacts'
+
+      - job: 'mac'
+        displayName: 'Mac OS'
+        timeoutInMinutes: 80
```
OwenArnold, line 50: You may find that timeouts can be revised for this smaller package as a follow up.

OwenArnold (approved):

PR #14, opened 2021-02-23
Prototype Scipp-Mantid comparison testing

OwenArnold:
Previously proposed here https://github.com/scipp/scipp/pull/1716

Moved across now that we have separate scipp-neturon

* Not including bespoke CI aspects yet. 
* Only single comparison so far, framework should be easy to extend

executing `pytest ... -s` will yield more detailed comparison info and timings.
```
<scipp.Dataset>
Dimensions: {{spectrum, 51200}, {tof, 100}}
Coordinates:
  diff                      float64    [counts]         (spectrum, tof)  [0.000000, 0.000000, ..., 0.000000, 0.000000]  [0.000000, 0.000000, ..., 0.000000, 0.000000]
  duration_mantid           float64    [s]              ()  [0.877711]
  duration_scipp            float64    [s]              ()  [0.137778]
  is_approx                 bool       [dimensionless]  (spectrum, tof)  [True, True, ..., True, True]
Data:
  histogram_events_test_with_CNCS_51936_event.nxs  bool       [dimensionless]  (spectrum, tof)  [True, True, ..., True, True]


PASSED
```

### DIFF
--- a/python/tests/mantid_scipp_comparison.py
+++ b/python/tests/mantid_scipp_comparison.py
@@ -1,0 +1,70 @@
+from mantid_data_helper import MantidDataHelper
+import scipp as sc
+import scipp.compat.mantid as mantid
+import time
+
+
+class MantidScippComparison:
+    def __init__(self, test_description=None):
+        self._test_description = test_description
+
+    def _execute_with_timing(self, op, input):
+        start = time.time()
+        result = op(input)
+        stop = time.time()
+        return result, (stop - start) * sc.Unit('s')
+
+    def _assert(self, a, b, allow_failure):
+        try:
+            assert sc.is_equal(a, b)
+        except AssertionError as ae:
+            if allow_failure:
+                print(ae)
+            else:
+                raise (ae)
+
+    def run(self, allow_failure=False):
+        import mantid.simpleapi as sapi
+        results = sc.Dataset()
+        for name, (hash, algorithm) in self._filenames.items():
+            file = MantidDataHelper.find_file(hash, algorithm)
+            print('Loading', name)
+            in_ws = sapi.Load(Filename=file, StoreInADS=False)
+            out_mantid_da, time_mantid = self._execute_with_timing(
+                self._run_mantid, input=in_ws)
+            in_da = mantid.from_mantid(in_ws).astype(
+                sc.dtype.float64)  # Converters set weights float32
+            out_scipp_da, time_scipp = self._execute_with_timing(
+                self._run_scipp, input=in_da)
+
+            self._assert(out_scipp_da, out_mantid_da, allow_failure)
+
+            result = sc.DataArray(sc.equal(out_mantid_da, out_scipp_da).data,
+                                  coords={
+                                      'diff':
+                                      out_mantid_da.data - out_scipp_da.data,
+                                      'is_approx':
+                                      sc.is_approx(
+                                          out_mantid_da.data,
+                                          out_scipp_da.data,
+                                          1e-9 * sc.Unit('counts') +
+                                          1e-9 * sc.abs(out_mantid_da.data)),
+                                      'duration_scipp':
+                                      time_scipp,
+                                      'duration_mantid':
+                                      time_mantid
+                                  })
+
+            results[f'with_{name}' if self._test_description is None else
+                    f'{self._test_description}_with_{name}'] = result
+        return results
+
+    @property
+    def _filenames(self):
+        pass
+
+    def _run_mantid(self, input):
+        raise RuntimeError("_run_mantid not implemented in base")
+
+    def _run_scipp(self, input):
+        raise RuntimeError("_run_scipp not implemented in base")
--- a/python/tests/mantid_scipp_comparison.py
+++ b/python/tests/mantid_scipp_comparison.py
@@ -1,6 +1,6 @@
 from mantid_data_helper import MantidDataHelper
 import scipp as sc
-import scipp.compat.mantid as mantid
+import scippneutron.mantid as mantid
 import time
 
 
--- a/python/tests/test_histogram_events.py
+++ b/python/tests/test_histogram_events.py
@@ -28,8 +28,7 @@ class HistogramEventsTest(MantidScippComparison):
     def _run_scipp(self, input):
         return sc.histogram(x=input,
                             bins=sc.Variable(dims=['tof'],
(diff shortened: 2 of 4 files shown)


python/tests/test_histogram_events.py
```diff
@@ -0,0 +1,40 @@
+from mantid_scipp_comparison import MantidScippComparison
+from mantid_data_helper import mantid_is_available
+import pytest
+import scipp.compat.mantid as mantid
```
OwenArnold, line 4: TODO needs to be `scippneutron`!

python/tests/test_histogram_events.py
```diff
@@ -0,0 +1,40 @@
...
+                                             values=np.arange(0, 1010, 10),
+                                             dtype=sc.dtype.float64,
+                                             unit=sc.units.us))
+
+
+@pytest.mark.skipif(not mantid_is_available(),
+                    reason='Mantid framework is unavailable')
+def test_histogram_events():
+    rebin = HistogramEventsTest()
+    print(rebin.run(allow_failure=True))
```
SimonHeybrock, line 40: Remove `print`?

python/tests/test_histogram_events.py
```diff
@@ -0,0 +1,40 @@
...
+                         Params=[0, 10, 1000],
+                         PreserveEvents=False,
+                         StoreInADS=False)
+        return mantid.from_mantid(out)
+
+    def _run_scipp(self, input):
+        return sc.histogram(x=input,
+                            bins=sc.Variable(dims=['tof'],
+                                             values=np.arange(0, 1010, 10),
+                                             dtype=sc.dtype.float64,
```
SimonHeybrock, line 32:
```suggestion
                                             values=np.linspace(0, 1000, num=11),
```

python/tests/test_histogram_events.py
```diff
@@ -0,0 +1,40 @@
...
+                                             unit=sc.units.us))
+
+
+@pytest.mark.skipif(not mantid_is_available(),
+                    reason='Mantid framework is unavailable')
```
SimonHeybrock, line 37: I wonder if this can be done a class (fixture) level or file level? I suppose once we add more tests like this adding the `skipif` to every single one becomes noise.

python/tests/test_histogram_events.py
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+    def _run_mantid(self, input):
+        import mantid.simpleapi as sapi
+        # Note Mantid rebin inclusive of last bin boundary
+        out = sapi.Rebin(InputWorkspace=input,
+                         Params=[0, 10, 1000],
+                         PreserveEvents=False,
+                         StoreInADS=False)
+        return mantid.from_mantid(out)
```
SimonHeybrock, line 26: Did you choose to have this conversion done by the child class (rather than the base) so specialized conversions can be supported?

python/tests/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def _filenames(self):
+        pass
+
+    def _run_mantid(self, input):
+        raise RuntimeError("_run_mantid not implemented in base")
+
+    def _run_scipp(self, input):
+        raise RuntimeError("_run_scipp not implemented in base")
```
SimonHeybrock, line 70: Are these needed? Isn't Python giving an appropriate error? Should we use `ABC` instead?

python/tests/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,70 @@
...
+                                      time_mantid
+                                  })
+
+            results[f'with_{name}' if self._test_description is None else
+                    f'{self._test_description}_with_{name}'] = result
+        return results
+
+    @property
+    def _filenames(self):
+        pass
```
SimonHeybrock, line 64: Dangerous? If forgotten to implement the user of this class may write many tests but it will not get run for any files?

python/tests/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,70 @@
...
+        import mantid.simpleapi as sapi
+        results = sc.Dataset()
+        for name, (hash, algorithm) in self._filenames.items():
+            file = MantidDataHelper.find_file(hash, algorithm)
+            print('Loading', name)
+            in_ws = sapi.Load(Filename=file, StoreInADS=False)
+            out_mantid_da, time_mantid = self._execute_with_timing(
+                self._run_mantid, input=in_ws)
+            in_da = mantid.from_mantid(in_ws).astype(
+                sc.dtype.float64)  # Converters set weights float32
```
SimonHeybrock, line 36: Can you elaborate? The converter sets *event weights* to `float32`, but should we really have this global conversion here?

OwenArnold: Where else would you propose to do this?

python/tests/mantid_scipp_comparison.py
OwenArnold: In follow up to this PR. I'm allowing users to implement `_workspaces` too, so it might be valid that no files are provided. However, it should not be possible to have no files **and** no workspaces as inputs to a given test. I'll change the using code to check for that.

python/tests/mantid_scipp_comparison.py
OwenArnold: Yes I think ABC would be better.

python/tests/test_histogram_events.py
OwenArnold: Yes

python/tests/test_histogram_events.py
OwenArnold: 👍

python/tests/test_histogram_events.py
OwenArnold: Agreed. I'll find a cleaner way to do this.

python/tests/mantid_scipp_comparison.py
SimonHeybrock: I suppose it should only be done for event data?

OwenArnold: 👍  yes, we do not want to generally do this

SimonHeybrock: I am not even sure `astype` supports (or should support) conversion of event weights?

OwenArnold:
Closing changes included on on #16

PR #16, opened 2021-02-26
More mantid scipp comparisons

OwenArnold:
Supersedes prototype PR  https://github.com/scipp/scippneutron/pull/14

* Unit conversions
* Beamline


python/tests/mantid_data_helper.py
```diff
@@ -7,6 +7,14 @@
 import subprocess as sp
```
SimonHeybrock, line 8: Above: Why was an empty `python/tests/__init__.py` odded?

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,92 @@
...
+                                                            input=in_ws)
+        in_da = mantid.from_mantid(in_ws).astype(
+            sc.dtype.float64)  # Converters set weights float32
+        out_scipp, time_scipp = self._execute_with_timing(self._run_scipp,
+                                                          input=in_da)
+
+        self._assert(out_scipp, out_mantid, allow_failure)
+
+        if isinstance(out_mantid, sc.DataArray):
+            result = sc.DataArray(sc.equal(out_mantid.data, out_scipp.data),
```
SimonHeybrock, line 38: I suppose this is only for output? For example, in `ConvertUnits` what we care about is mainly whether coords are equal, so you may have a failing comparison but this would be all `True`?

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _execute_with_timing(self, op, input):
+        start = time.time()
+        result = op(input)
+        stop = time.time()
+        return result, (stop - start) * sc.Unit('s')
+
+    def _assert(self, a, b, allow_failure):
+        try:
+            assert sc.is_equal(a, b)
```
SimonHeybrock, line 20: Why is `is_equal` used here? In general I assume this would fail as soon as any floating-point arithmetic is involved?

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,92 @@
...
+                                      'duration_mantid':
+                                      time_mantid
+                                  })
+        else:
+            result = out_mantid - out_scipp
+        return result
+
+    def _add_result_to_dataset(self, name, result, results):
+        results[f'with_{name}' if self._test_description is None else
+                f'{self._test_description}_with_{name}'] = result
```
SimonHeybrock, line 58: I don't understand, why is `Dataset` used here? In general the results all have different lengths I suppose, e.g., when files from different instruments are used?

python/tests/mantid_scipp_comparison/test_beamline_calculations.py
```diff
@@ -0,0 +1,58 @@
...
+                                        NumBanks=1,
+                                        XMax=200,
+                                        StoreInADS=False)
+        return {"sample_workspace": ws}
+
+
+@pytest.mark.skipif(not mantid_is_available(),
+                    reason='Mantid framework is unavailable')
+class TestBeamlineCalculations:
+    def test_bealine_calculations_l1(self):
```
SimonHeybrock, line 27:
```suggestion
    def test_beamline_calculations_l1(self):
```

python/tests/mantid_scipp_comparison/test_beamline_calculations.py
```diff
@@ -0,0 +1,58 @@
...
+        class L1Comparison(BeamlineComparision):
+            def _run_mantid(self, input):
+                return input.detectorInfo().l1() * sc.Unit('m')
+
+            def _run_scipp(self, input):
+                return sn.l1(input)
+
+        test = L1Comparison()
+        test.run(allow_failure=True)
```
SimonHeybrock, line 36: Why allowing failure here and below?

python/tests/mantid_scipp_comparison/test_beamline_calculations.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.mark.skipif(not mantid_is_available(),
+                    reason='Mantid framework is unavailable')
+class TestBeamlineCalculations:
+    def test_bealine_calculations_l1(self):
+        class L1Comparison(BeamlineComparision):
+            def _run_mantid(self, input):
+                return input.detectorInfo().l1() * sc.Unit('m')
+
+            def _run_scipp(self, input):
+                return sn.l1(input)
```
SimonHeybrock, line 33: Nice examples of how this can also be used for things that are not `Algorithm` 👍

python/tests/mantid_scipp_comparison/test_neutron_convert_units.py
```diff
@@ -0,0 +1,120 @@
...
+import scippneutron.mantid as converter
+import scippneutron as sn
+import scipp as sc
+import enum
+
+
+class Emode(enum.Enum):
+    Elastic = 0
+    Direct = 1
+    Indirect = 2
```
SimonHeybrock, line 13: Unused?

python/tests/mantid_data_helper.py
OwenArnold: Structured like this so that I can keep the new mantid-comparison tests in their own directory while sharing some of the higher level modules with the scippneutron unittests. Maybe there's a better way?

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
OwenArnold: This is only supposed to be for diagnostics. We do not use the results directly. I tried tackling this as inspired by the Mantid `CompareWorkspaces` type problem, where there is no good way to return the results of the comparison, which scipp actually give a lot more flexibility for and can be coupled with `sc.table`. This can be updated/improved later as results are not used directly.

SimonHeybrock: Can we just return the two result data arrays on failure? Simpler, and the client can decide what to do with it.

python/tests/mantid_data_helper.py
SimonHeybrock: So the init is needed such that you can import from the parent directory?

OwenArnold: Correct. I'm not aware of another way to do this.

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -17,7 +17,13 @@ def _execute_with_timing(self, op, input):
 
     def _assert(self, a, b, allow_failure):
         try:
-            assert sc.is_equal(a, b)
+            if isinstance(a, sc.DataArray):
+                assert sc.all(
+                    sc.is_approx(a.data, b.data, 1e-9 * a.data.unit +
+                                 1e-9 * sc.abs(a.data))).value
```
SimonHeybrock, line 23: So we are not testing, e.g., the result of unit conversions?

OwenArnold: Yes, good point. It's absolutely critical that we test that.` is_approx` is only implemented for variables. Unlike `is_equal` which is simply bound to `operator==`. I think we should therefore

1. Loop over all coords and compare them wish `is_approx` as well. For the purpose of this PR.
1. Fully provide this support for `is_approx` in scipp (i'm in favour of that), but it'll hold up these changes. I'm working on `is_approx` at the moment, so i can do this at the same time.
1. Refactor the test code to use `is_approx` on the `dataarray` directly.

SimonHeybrock: > Fully provide this support for is_approx in scipp (i'm in favour of that), but it'll hold up these changes. I'm working on is_approx at the moment, so i can do this at the same time.

I think what we currently call `is_approx` should be changed to `isclose` an mimick numpy behavior. Just like other scipp binary function is should do the normal coord and mask handling (compare coords, OR masks).

For unit tests a comparison that also does fuzzy compares of coords is certainly useful, but I think it is important to keep that separate.

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -0,0 +1,110 @@
...
+            y = b.meta[key]
+            if x.shape != y.shape:
+                raise RuntimeError(f'For meta {key} have different'
+                                   f' shapes {x.shape}, {y.shape}')
+            if val.dtype in [sc.dtype.float64, sc.dtype.float32]:
+                if sc.sum(~sc.isfinite(x)).value > 0 or sc.sum(
+                        ~sc.isfinite(y)).value > 0:
+                    raise RuntimeError(
+                        f'For meta {key} have non-finite entries')
+                coord_tol_factor = 1e-6
```
SimonHeybrock, line 36: Why so large?

OwenArnold: I'll investigated which unit conversion is problematic. Yes 1e-7 is too small leads to mismatch.

OwenArnold: Ok. I'm going to open up a separate issue for this.

SimonHeybrock (approved):

PR #17, opened 2021-03-01
Add direct nexus loading

matthew-d-jones:
Implements `scippneutron.load_nexus()`
Scope in file can be limited using `root` argument, for example to load from specific `NXentry`, or even to load a single `NXlog`. Behaviour in presence of multiple `NXentry` groups etc is documented by unit tests. Unit tests use an in-memory NeXus file which is constructed using a helper class (builder pattern).
If there are loadable event data in the file then a `DataArray` is returned, if there are only logs then a `Dataset` is returned, if nothing is loaded then returns `None`.
Loading is NeXus compliant, in particular it is not sensitive to group names or to the location of data groups in the hierarchy of the file.

I've not optimised performance, but it is currently roughly comparable to loading with `load()` via Mantid. The approach to concatenating data from different event data groups slows things a bit, for example our `PG3_4844_event.nxs` test file contains ~20 event data groups and `load_nexus` takes ~12 seconds, `load` takes ~8 seconds. NB, `load_nexus` does not load pulse times yet, loading pulse times and broadcasting them to events is a foreseeable performance hurdle.

Implemented loading:
- `NXevent_data` - and bin by detector-id
- `NXlog` (with or without `time` dataset)
- pixel positions
- instrument name
- experiment title

Not yet implemented (covered by tickets #19, #20, #21):
- parsing `NXtransformation` chains
- pulse times with event data
- source and sample position
- start of run time
- adjusting log times for `offset` attribute (make consistent, for example from start of run)

### DIFF
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -1,0 +1,28 @@
+from typing import Union, Any
+
+import numpy as np
+
+
+def ensure_str(str_or_bytes: Union[str, bytes]) -> str:
+    try:
+        str_or_bytes = str(str_or_bytes, encoding="utf8")  # type: ignore
+    except TypeError:
+        pass
+    return str_or_bytes
+
+
+class BadSource(Exception):
+    pass
+
+
+unsigned_to_signed = {
+    np.uint32: np.int32,
+    np.uint64: np.int64,
+}
+
+
+def ensure_not_unsigned(dataset_type: Any):
+    try:
+        return unsigned_to_signed[dataset_type]
+    except KeyError:
+        return dataset_type
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -2,11 +2,18 @@
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # @author Matthew Jones
 
-from typing import Union, Any, List, Optional
+from typing import Union, Any, List, Optional, Tuple, Dict
 
 import h5py
 import numpy as np
 import scipp as sc
+
+
+def _get_attr_as_str(h5_object, attribute_name: str):
+    try:
+        return h5_object.attrs[attribute_name].decode("utf8")
+    except AttributeError:
+        return h5_object.attrs[attribute_name]
 
 
 def ensure_str(str_or_bytes: Union[str, bytes]) -> str:
@@ -62,3 +69,32 @@ def get_units(dataset: h5py.Dataset) -> str:
     except (AttributeError, KeyError):
         return ""
     return ensure_str(units)
+
+
+def find_by_nx_class(
+        nx_class_names: Tuple[str, ...],
+        root: Union[h5py.File, h5py.Group]) -> Dict[str, List[h5py.Group]]:
+    """
+    Finds groups with requested NX_class in the subtree of root
+
+    Returns a dictionary with NX_class name as the key and list of matching
+    groups as the value
+    """
+    groups_with_requested_nx_class: Dict[str, List[h5py.Group]] = {
+        class_name: []
+        for class_name in nx_class_names
+    }
+
+    def _match_nx_class(_, h5_object):
+        if isinstance(h5_object, h5py.Group):
+            try:
+                nx_class = _get_attr_as_str(h5_object, "NX_class")
+                if nx_class in nx_class_names:
+                    groups_with_requested_nx_class[nx_class].append(h5_object)
+            except KeyError:
+                pass
+
+    root.visititems(_match_nx_class)
+    # Also check if root itself is an NX_class
+    _match_nx_class(None, root)
+    return groups_with_requested_nx_class
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -1,5 +1,8 @@
-from typing import Union, Any
+# SPDX-License-Identifier: GPL-3.0-or-later
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+# @author Matthew Jones
 
+from typing import Union, Any
 import numpy as np
 
 
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -37,6 +37,13 @@ def ensure_not_unsigned(dataset_type: Any):
 def load_dataset(dataset: h5py.Dataset,
                  dimensions: List[str],
                  dtype: Optional[Any] = None) -> sc.Variable:
+    """
+    Load an HDF5 dataset into a Scipp Variable
+    :param dataset: The dataset to load
+    :param dimensions: Dimensions for the output Variable
+    :param dtype: Cast to this dtype during load,
+      otherwise retain dataset dtype
+    """
     if dtype is None:
         dtype = ensure_not_unsigned(dataset.dtype.type)
     units = _get_units(dataset)
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -2,8 +2,11 @@
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # @author Matthew Jones
 
-from typing import Union, Any
+from typing import Union, Any, List, Optional
+
+import h5py
 import numpy as np
+import scipp as sc
 
 
 def ensure_str(str_or_bytes: Union[str, bytes]) -> str:
@@ -29,3 +32,25 @@ def ensure_not_unsigned(dataset_type: Any):
         return unsigned_to_signed[dataset_type]
     except KeyError:
         return dataset_type
+
+
+def load_dataset(dataset: h5py.Dataset,
+                 dimensions: List[str],
+                 dtype: Optional[Any] = None) -> sc.Variable:
+    if dtype is None:
+        dtype = ensure_not_unsigned(dataset.dtype.type)
+    units = _get_units(dataset)
+    variable = sc.empty(dims=dimensions,
+                        shape=dataset.shape,
+                        dtype=dtype,
+                        unit=units)
+    dataset.read_direct(variable.values)
+    return variable
+
+
+def _get_units(dataset: h5py.Dataset) -> str:
+    try:
+        units = dataset.attrs["units"]
+    except (AttributeError, KeyError):
+        return ""
+    return ensure_str(units)
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -22,6 +22,8 @@ class BadSource(Exception):
 
 
 unsigned_to_signed = {
+    np.uint8: np.int8,
+    np.uint16: np.int16,
     np.uint32: np.int32,
     np.uint64: np.int64,
 }
--- a/python/src/scippneutron/_loading_common.py
+++ b/python/src/scippneutron/_loading_common.py
@@ -46,16 +46,15 @@ def load_dataset(dataset: h5py.Dataset,
     """
     if dtype is None:
         dtype = ensure_not_unsigned(dataset.dtype.type)
-    units = _get_units(dataset)
     variable = sc.empty(dims=dimensions,
                         shape=dataset.shape,
                         dtype=dtype,
-                        unit=units)
+                        unit=get_units(dataset))
     dataset.read_direct(variable.values)
     return variable
 
 
-def _get_units(dataset: h5py.Dataset) -> str:
+def get_units(dataset: h5py.Dataset) -> str:
     try:
         units = dataset.attrs["units"]
     except (AttributeError, KeyError):
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # @author Matthew Jones
 
-from dataclasses import dataclass, astuple
+from dataclasses import dataclass
 import h5py
 from typing import Optional, List
 import numpy as np
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -3,8 +3,7 @@ import h5py
 from typing import Optional, List
 import numpy as np
 from ._loading_common import ensure_str, BadSource, ensure_not_unsigned
-from ..._scipp import core as sc
-from ..._bins import bin
+import scipp as sc
 from datetime import datetime
 from warnings import warn
 from itertools import groupby
@@ -165,12 +164,12 @@ def load_detector_data(
         return
     else:
 
-        def getDetectorId(events_and_max_det_id: DetectorData):
+        def get_detector_id(events_and_max_det_id: DetectorData):
             # Assume different detector banks do not have
             # intersecting ranges of detector ids
             return events_and_max_det_id.detector_ids[0]
 
-        event_data.sort(key=getDetectorId)
+        event_data.sort(key=get_detector_id)
 
         detector_ids = sc.Variable(
             dims=[_detector_dimension],
@@ -191,7 +190,7 @@ def load_detector_data(
         # Events in the NeXus file are effectively binned by pulse
         # (because they are recorded chronologically)
         # but for reduction it is more useful to bin by detector id
-        events = bin(events, groups=[detector_ids])
+        events = sc.bin(events, groups=[detector_ids])
         if pixel_positions_loaded:
             events.coords['position'] = pixel_positions
 
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -106,7 +106,7 @@ def _load_dataset(dataset: h5py.Dataset,
     return variable
 
 
-def _load_event_group(group: h5py.Group) -> DetectorData:
+def _load_event_group(group: h5py.Group, quiet: bool) -> DetectorData:
     error_msg = _check_for_missing_fields(group)
     if error_msg:
         raise BadSource(error_msg)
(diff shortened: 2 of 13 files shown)


python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+    if error_msg:
+        raise BadSource(error_msg)
+
+    # There is some variation in the last recorded event_index in files
+    # from different institutions. We try to make sure here that it is what
+    # would be the first index of the next pulse.
+    # In other words, ensure that event_index includes the bin edge for
+    # the last pulse.
+    event_id_ds = group["event_id"]
+    event_index = group["event_index"][...].astype(np.int64)
```
SimonHeybrock, line 106: Why do we need this to be `int64`? Is it unsigned in the file?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+    else:
+        event_index[-1] = event_id_ds.len()
+
+    number_of_events = event_index[-1]
+    event_time_offset_ds = group["event_time_offset"]
+    event_time_offset = sc.Variable([_event_dimension],
+                                    values=event_time_offset_ds[...],
+                                    dtype=ensure_not_unsigned(
+                                        event_time_offset_ds.dtype.type),
```
SimonHeybrock, line 120: Unless the unsigned->signed conversion is required, can we use `h5py.read_direct` to avoid the extra allocation and copy?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # Weights are not stored in NeXus, so use 1s
+    weights = sc.Variable([_event_dimension],
+                          values=np.ones(event_id.shape),
+                          dtype=np.float32)
+    data = sc.DataArray(data=weights,
+                        coords={
+                            'tof': event_time_offset,
+                            _detector_dimension: event_id
+                        })
```
SimonHeybrock, line 134: There are helpers in `sc.detail`, `move` and `move_to_data_array`, that can avoid copies here (and probably some other places).

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+            warn(f"Skipped loading {group.name} due to:\n{e}")
+
+    if not event_data:
+        return
+    else:
+
+        def get_detector_id(events_and_max_det_id: DetectorData):
+            # Assume different detector banks do not have
+            # intersecting ranges of detector ids
+            return events_and_max_det_id.detector_ids[0]
```
SimonHeybrock, line 174: Would it make sense to sort by the group name in the file?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+        pixel_positions_loaded = all(
+            [data.pixel_positions is not None for data in event_data])
+        events, _, pixel_positions = astuple(event_data.pop(0))
+        while event_data:
+            new_events, _, new_pixel_positions = astuple(event_data.pop(0))
+            events = sc.concatenate(events, new_events, dim=_event_dimension)
+            if pixel_positions_loaded:
+                pixel_positions = sc.concatenate(pixel_positions,
+                                                 new_pixel_positions,
+                                                 dim=_detector_dimension)
```
SimonHeybrock, line 192: So if I understand correctly this is concatenating all event lists before binning into pixels?

Another option would be to *first* do the binning and *then* concatenate. There may be a couple of advantages to this:
- Less memory consumption for temporaries/copies in `sc.bin`.
- Potentially faster `sc.bin` since it would act on pre-grouped data. Lookup of target pixel would only need to consider `Nbank_pixels` rather than `Ninstrument_pixels`.
- More future proof? For example, someone might want to preverse the bank-splitting, e.g., in instruments with detectors for different techniques such as imaging and diffraction, which would not get merged for the workflow?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # Events in the NeXus file are effectively binned by pulse
+        # (because they are recorded chronologically)
+        # but for reduction it is more useful to bin by detector id
```
SimonHeybrock, line 196: There was this issue I had open with ECDC that got recently implemented, to ensure that the file-writter would set a flag when things were *not* ordered by pulse (apprantely this is possible). Can we check for this flag?

SimonHeybrock (commented):
Only looked at the even-data loading so far:

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: Yes, it is typically uint32, but converting uint32 -> int32 or uint64 -> int64 would probably be better.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: I don't think so, there is no constraint on naming of detector groups, for DREAM we may have "endcap" and "mantle" for example.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: It's not useful until we load pulse times though right?
But I can go and check how that was implemented and add some code or comment here to avoid risk of someone else having to go delving in the Filewriter codebase.

SimonHeybrock: Might be useful. I still have in mind that we may want to do certain filtering operations before binning into pixels (e.g., to filter/split by polarization state, or filter out pulses based on bad chopper params). This might be easier if we know it is ordered.

python/src/scippneutron/_loading_log_data.py
```diff
@@ -0,0 +1,107 @@
...
+
+def _load_log_data_from_group(group: h5py.Group) -> Tuple[str, sc.Variable]:
+    property_name = group.name.split("/")[-1]
+    value_dataset_name = "value"
+    time_dataset_name = "time"
+
+    try:
+        values = group[value_dataset_name][...]
+    except KeyError:
+        raise BadSource(f"NXlog at {group.name} has no value dataset")
```
OwenArnold, line 54: I quite like the use of this for encoding compliance with the Nexus standard. Presumably it would be too verbose or too fragile to reference the standard itself?

python/tests/test_load_nexus.py
```diff
@@ -0,0 +1,416 @@
...
+        np.sort(event_time_offsets))
+
+    counts_on_detectors = loaded_data.bins.sum()
+    # No detector_number dataset in file so expect detector-id to be
+    # binned according to whatever detector-ids are present in event_id
+    # dataset: 2 on det 1, 1 on det 2, 2 on det 3
+    expected_counts = np.array([2, 1, 2])
+    assert np.allclose(counts_on_detectors.data.values, expected_counts)
+    expected_detector_ids = np.array([1, 2, 3])
+    assert np.allclose(loaded_data.coords['detector-id'].values,
```
OwenArnold, line 67: You could use `sc.is_approx` here.

python/tests/nexus_helpers.py
```diff
@@ -0,0 +1,244 @@
...
+    value_ds = log_group.create_dataset("value", data=log.value)
+    if log.value_units is not None:
+        value_ds.attrs.create("units", data=log.value_units)
+    if log.time is not None:
+        time_ds = log_group.create_dataset("time", data=log.time)
+        if log.time_units is not None:
+            time_ds.attrs.create("units", data=log.time_units)
+
+
+class InMemoryNexusFileBuilder:
```
OwenArnold, line 163: 👍

python/src/scippneutron/_loading_common.py
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
```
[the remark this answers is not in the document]
SimonHeybrock: This is not pointing to the repo but the organization. This is the zero throughout the projects at this point.

python/src/scippneutron/_loading_log_data.py
matthew-d-jones: The problem with that is that everything in the standard is optional. It is not against the standard for the value dataset to be missing, it is simply that we can't do anything with an NXlog that doesn't have a value dataset in our application.

python/tests/test_load_nexus.py
matthew-d-jones: Here I'm comparing against numpy arrays though.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: I've swapped binning and concatenation order as you suggest, and it did improve performance a little.
Another possible approach would be to avoid concatenation entirely by `read_direct` into slices of a single `Variable` for all event data groups. It would be less flexible, but something to consider if the time for the concatenation step is still significant.

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,201 @@
...
+                                    values=event_time_offset_ds[...],
+                                    dtype=ensure_not_unsigned(
+                                        event_time_offset_ds.dtype.type),
+                                    unit=_get_units(event_time_offset_ds))
+    event_id = sc.Variable(
+        [_event_dimension], values=event_id_ds[...],
+        dtype=np.int32)  # assume int32 is safe for detector ids
+
+    # Weights are not stored in NeXus, so use 1s
+    weights = sc.Variable([_event_dimension],
```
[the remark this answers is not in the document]
matthew-d-jones: @SimonHeybrock now implemented them and I've made use of them.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: The ticket for flagging unordered data is in selected, not yet implemented, ticket number is ECDC-641.
I'll try to move that along by making a proposal for how the flag could be recorded in file.

matthew-d-jones:
I think I have addressed all comments so far.
I have not used `read_direct` in a few places, including loading pixel positions, where a flatten operation is required.
The current build failure is the dependency version clash which needs to be resolved in the mantid framework package, I'm looking into that now.

python/src/scippneutron/_loading_detector_data.py
SimonHeybrock: I think the real way to improve here is to support concatenate with a vector of inputs, rather than just two inputs.

OwenArnold (commented):
Looks good.

This is one of the first notable features for scippneutron.
1. We should make a mention of this in some release notes. We havent yet created the pages for that. Would be similar to those in scipp.
2. We should maybe exercise the new functionality in some of our other docs, for example https://github.com/scipp/scippneutron/blob/main/docs/tutorials/neutron-diffraction.ipynb

Maybe both of these are best done as followups.

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # Weights are not stored in NeXus, so use 1s
+    weights = sc.ones(dims=[_event_dimension],
+                      shape=event_id.shape,
+                      dtype=np.float32)
+
+    detector_number_ds_name = "detector_number"
+    if detector_number_ds_name in group.parent:
+        # Hopefully the detector ids are recorded in the file
+        detector_ids = group.parent[detector_number_ds_name][...].flatten()
```
SimonHeybrock, line 121: Wondering why we are flattening this and the positions in a file above? Can't we keep the dimensionality?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,194 @@
...
+    else:
+        # Otherwise we'll just have to bin according to whatever
+        # ids we have a events for (pixels with no recorded events
+        # will not have a bin)
+        detector_ids = np.unique(event_id.values)
+
+    detector_ids = sc.Variable(dims=[_detector_dimension],
+                               values=detector_ids,
+                               dtype=np.int32)
```
SimonHeybrock, line 130: While `int32` is what Mantid uses, is there any reason against keeping what the file contains (with a potential switch to signed)?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,194 @@
...
+    for group in event_data_groups:
+        try:
+            new_event_data = _load_event_group(group, quiet)
+            event_data.append(new_event_data)
+        except BadSource as e:
+            warn(f"Skipped loading {group.name} due to:\n{e}")
+
+    if not event_data:
+        return
+    else:
```
SimonHeybrock, line 166: Maybe just drop the else, for less confusing spacing and indentation?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -0,0 +1,194 @@
...
+        events = sc.bin(detector_data.events,
+                        groups=[detector_data.detector_ids])
+        if pixel_positions_loaded:
+            events.coords['position'] = detector_data.pixel_positions
+        while event_data:
+            detector_data = event_data.pop(0)
+            new_events = sc.bin(detector_data.events,
+                                groups=[detector_data.detector_ids])
+            if pixel_positions_loaded:
+                new_events.coords['position'] = detector_data.pixel_positions
```
SimonHeybrock, line 190: Just for my understanding of Nexus: This implies that the files store a separate array of pixel positions for every `NXevent_data`, rather than a single array?

python/src/scippneutron/load_nexus.py
```diff
@@ -0,0 +1,138 @@
...
+        nx_log = "NXlog"
+        nx_entry = "NXentry"
+        nx_instrument = "NXinstrument"
+        groups = _find_by_nx_class(
+            (nx_event_data, nx_log, nx_entry, nx_instrument), nexus_file[root])
+
+        if len(groups[nx_entry]) > 1:
+            # We can't sensibly load from multiple NXentry, for example each
+            # could could contain a description of the same detector bank
+            # and lead to problems with clashing detector ids etc
```
SimonHeybrock, line 111: I suppose we could consider return a dict of data arrays or a dataset then? Is this something that is used in practice? IF so we can consider supporting it in the future.

python/tests/nexus_helpers.py
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    def _match_nx_class(_, h5_object):
+        if isinstance(h5_object, h5py.Group):
+            try:
+                nx_class = _get_attr_as_str(h5_object, "NX_class")
+                if nx_class in nx_class_names:
+                    groups_with_requested_nx_class[nx_class].append(h5_object)
+            except KeyError:
+                pass
```
SimonHeybrock, line 38: There is a very similar-looking method `_find_by_nx_class` above?

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: I thought about that, but wouldn't that involve a large memory overhead?

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: Yes, each `NXdetector` contains the pixel positions and `NXevent_data` for that detector.
Nesting `NXdetector`s to make this more flexible has been discussed, but there are no examples of that in wild yet.

python/src/scippneutron/load_nexus.py
matthew-d-jones: The only example I've seen of using multiple `NXentry` in a single file was from a light source facility.

python/tests/nexus_helpers.py
matthew-d-jones: Oops, something went wrong in the refactoring

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: If there is only one detector then sure, but otherwise it would surely lead to problems when concatenating with other detectors?

python/src/scippneutron/_loading_detector_data.py
SimonHeybrock: Why? The end result is the same size, we just avoid all the intermediate steps. In fact the memory used should be less. Say we have banks of sizes A, B, C, D:
- If we go stepwise, largest memory use will be in the last step `ABC + D + ABCD` plus what was previously used for A, B, C, and D individually.
- If we go in one step we just have ABCD.
-

SimonHeybrock:
> Looks good.
> 
> This is one of the first notable features for scippneutron.
> 
>     1. We should make a mention of this in some release notes. We havent yet created the pages for that. Would be similar to those in scipp.

We have, see https://scipp.github.io/scippneutron/about/release-notes.html. However, I somewhat disagree about adding this to the release note. Personally I'd prefer to try this ourselves for a while before publicizing it.

>     2. We should maybe exercise the new functionality in some of our other docs, for example https://github.com/scipp/scippneutron/blob/main/docs/tutorials/neutron-diffraction.ipynb
> 
> 
> Maybe both of these are best done as followups.

Yes, we should *not* do this in this release.

python/src/scippneutron/_loading_detector_data.py
SimonHeybrock: You are thinking along the lines of having, e.g., the DREAM mantle and endcap detectors, which have different logical 3D pixel volumes and could not be concatenated?

matthew-d-jones: The DREAM geometry is complicated enough that I just recorded positions and ids in 1D datasets anyway.
But it seems to be usual to record ids and positions for a simple rectangular imaging detector panel in 2D, to mirror the physical layout of pixels. If you have an instrument with one detector bank like that, and another bank with 1D id and position datasets, then there would be a problem.

SimonHeybrock: Maybe it should just fail then? Ignoring the dimensions that are contained in the file feels odd?

matthew-d-jones: The current implementation allows loading all data in all cases, it seems like the right place to start to me. 
Retaining the dimensionality would be possible in conjunction with the option of not concatenating the data from different panels. But that feels like adding complexity for guessed requirements at this stage.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: I've now changed it to retain the original type. This required checking that event ids and detector numbers are the same integer type in order to give a nice warning to the user and skip the current event group. I added units tests to cover this.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: done

python/tests/nexus_helpers.py
matthew-d-jones: removed duplicate

python/src/scippneutron/_loading_detector_data.py
SimonHeybrock: 👍

SimonHeybrock (approved):

PR #24, opened 2021-03-03
Coord naming conventions

SimonHeybrock:
See #scipp/1642.


SimonHeybrock:
The earlier segfault in CI came from a change of a map of builtin dimension labels that had not been merged into the `scipp` conda package yet. Since IDs are defined in a header, `scippneutron` used wrong ID in the map lookup.

OwenArnold:
> The earlier segfault in CI came from a change of a map of builtin dimension labels that had not been merged into the `scipp` conda package yet. Since IDs are defined in a header, `scippneutron` used wrong ID in the map lookup

Thanks for explaining.

docs/about/release-notes.rst
```diff
@@ -16,6 +16,16 @@ Breaking changes
...
+* Naming convention for (in particular) coords and attrs used by unit conversion has changed.
+  Generally what previously used hyphens `-` now uses underscore `_`.
+
+  * ``pulse-time`` is now ``pulse_time``
+  * ``sample-position`` is now ``sample_position``
+  * ``source-position`` is now ``source_position``
+  * ``energy-transfer`` is now ``energy_transfer``
+  * ``incident-energy`` is now ``incident_energy``
+  * ``final-energy`` is now ``final_energy``
+  * ``d-spacing`` is now ``dspacing` (no hypthen)
```
OwenArnold, line 28: We should specifically list these here https://github.com/scipp/scippneutron/blob/coord-naming-conventions/docs/user-guide/unit-conversions.ipynb rather than having users solve these problems one-by one if they are providing the data via some non-mantid route. This may well be the case for imaging input date for example in the short term.

OwenArnold (changes requested):
Apart from additional documentation suggestion. Everything else looks good.

docs/about/release-notes.rst
SimonHeybrock: I deliberately avoided this to avoid double work, since I will need to rewrite this anyway with the upcoming changes, documenting the logic behind how and which meta data is found, prioritized, or computed.

OwenArnold (approved):

PR #26, opened 2021-03-03
Support overriding beamline parameters

SimonHeybrock:
Fixes https://github.com/scipp/scipp/issues/1672 (and more).

How does it work?
- All operations try to refer only to what the actually need to know, e.g., `L2` instead of `position` and `sample_position`. The helper functions `l2` can then figure out how to obtain it.
- Most explicit matching coords take priority.

Todo (Future):
- Considering to support providing overriding via arguments to `convert` rather than via coords, but this is more complicated, and could become easier if the ownership model is changed. I suggest to avoid this for now. This would make it slightly ambiguous as to whether existing coords or coords from args are used, so maybe we want to avoid this entirely.

### DIFF
--- a/docs/user-guide/unit-conversions.ipynb
+++ b/docs/user-guide/unit-conversions.ipynb
@@ -115,7 +115,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The parameters given in the table above are retrived from the coordinates or attributes of the input data.\n",
+    "The parameters given in the table above are retrieved from the coordinates or attributes of the input data.\n",
     "The following graph illustrates which coords (or attrs) are known and used by `convert`.\n",
     "Read the graph as follows:\n",
     "- Ellipses represent geometry-related parameters of the beamline, e.g., `two_theta`.\n",
@@ -130,7 +130,7 @@
     "- If, e.g., `L2` is not found it is computed as the norm of the `scattered_beam` coord or attr.\n",
     "- If, `scattered_beam` is not found it us computed as `position - sample_position`.\n",
     "\n",
-    "There are two case, scattering conversion and non-scattering conversion (e.g., for beam monitors or for imaging beamlines).\n",
+    "There are two cases, scattering conversion and non-scattering conversion (e.g., for beam monitors or for imaging beamlines).\n",
     "\n",
     "Case 1: `convert(..., scatter=True)`"
    ]
--- a/docs/user-guide/unit-conversions.ipynb
+++ b/docs/user-guide/unit-conversions.ipynb
@@ -4,7 +4,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Unit conversions"
+    "# Unit conversions\n",
+    "\n",
+    "## Available conversions"
    ]
   },
   {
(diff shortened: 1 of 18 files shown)


SimonHeybrock:
These are the dot graph I added to the docs:
image

image

jl-wynen:
I take it, TOF is implicit in the graphs because there is no way to deduce like like, e.g. L1?

Would it make sense to have 'pixel_position' instead of just 'position' to avoid confusion? Or maybe support either when computing 'scattered_beam'?

SimonHeybrock:
> I take it, TOF is implicit in the graphs because there is no way to deduce like like, e.g. L1?

Well, partially. You can also convert from, e.g., `wavelength` to `energy`, then `tof` is not needed. So strictly speaking the graph should have all possible combinations of units, but that would be a bit messy.

> Would it make sense to have 'pixel_position' instead of just 'position' to avoid confusion? Or maybe support either when computing 'scattered_beam'?

The problem is that people use "detector", "pixel", and "spectrum" to describe the same or related concept. We do not want to support multiple ambiguous ways of specifying this. Which would take priority if someone has `pixel_position` and `spectrum_position`?

jl-wynen:
> So strictly speaking the graph should have all possible combinations of units, but that would be a bit messy.

Maybe, not sure how messy it would be. I was just confused how you want to get a wavelength from just 'Ltotal'.

> The problem is that people use "detector", "pixel", and "spectrum" to describe the same or related concept. We do not want to support multiple ambiguous ways of specifying this. Which would take priority if someone has pixel_position and spectrum_position?

Yes, I agree that we should not have a resolution order for different ways of saying the same thing. But is 'position' not ambiguous on its own?

SimonHeybrock:
> > The problem is that people use "detector", "pixel", and "spectrum" to describe the same or related concept. We do not want to support multiple ambiguous ways of specifying this. Which would take priority if someone has pixel_position and spectrum_position?
> 
> Yes, I agree that we should not have a resolution order for different ways of saying the same thing. But is 'position' not ambiguous on its own?

Not ambiguous, because `position` is actually a coordinate for a dimension of the data (often the dimension is called `spectrum`), whereas `sample_position` and `source_position` are scalar coordinates.

jl-wynen:
What if the detector can be moved and every event is recorded at a different position? Maybe someone understands 'position' as the position of the robot arm? Or the sample gets moved?

Don't know if this can really cause confusion in practice.

SimonHeybrock:
> What if the detector can be moved and every event is recorded at a different position? Maybe someone understands 'position' as the position of the robot arm? Or the sample gets moved?
> 
> Don't know if this can really cause confusion in practice.

In those cases there would be an *additional* dimension, e.g., `sample_position`, and `sample_position` would become a dimension-coord of that dim. `position` is still unambiguously a coord of `spectrum` dim.

docs/user-guide/unit-conversions.ipynb
```diff
@@ -45,6 +47,118 @@
...
+    "dot.edge('L1', 'Ltotal')\n",
+    "dot.edge('L2', 'Ltotal')\n",
+    "dot.edge('incident_beam', 'two_theta')\n",
+    "dot.edge('scattered_beam', 'two_theta')\n",
+    "dot.node('wavelength', shape='box', style='filled', color='lightgrey')\n",
+    "dot.node('energy', shape='box', style='filled', color='lightgrey')\n",
+    "dot.node('energy_transfer (direct)', shape='box', style='filled', color='lightgrey')\n",
+    "dot.node('energy_transfer (indirect)', shape='box', style='filled', color='lightgrey')\n",
+    "dot.node('dspacing', shape='box', style='filled', color='lightgrey')\n",
+    "dot.node('Q', shape='box', style='filled', color='lightgrey')\n",
```
OwenArnold, line 83: Q -> ||Q||

docs/user-guide/unit-conversions.ipynb
```diff
@@ -45,6 +47,118 @@
...
+    "- If a unit conversion requires a parameter the tree of paramameter ellipses is searched bottom-up.\n",
+    "\n",
+    "Example:\n",
+    "- Consider a unit conversion to `wavelength` (from, e.g., `tof`) *or* from `wavelength` (to, e.g., `tof`) will require the parameters connected to the rectangle by arrows, in this case `Ltotal`.\n",
+    "- If a coord or attr with name `Ltotal` is found it is used for the conversion.\n",
+    "- If `Ltotal` is not found it will be computed based on coords or attrs `L1` and `L2`.\n",
+    "- If, e.g., `L2` is not found it is computed as the norm of the `scattered_beam` coord or attr.\n",
+    "- If, `scattered_beam` is not found it us computed as `position - sample_position`.\n",
+    "\n",
+    "There are two case, scattering conversion and non-scattering conversion (e.g., for beam monitors or for imaging beamlines).\n",
```
OwenArnold, line 133:
```suggestion
    "There are two cases, scattering conversion and non-scattering conversion (e.g., for beam monitors or for imaging beamlines).\n",
```

docs/user-guide/unit-conversions.ipynb
```diff
@@ -45,6 +47,118 @@
...
+    "        display(scattering)\n",
+    "    else:\n",
+    "        display(nonscattering)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The parameters given in the table above are retrived from the coordinates or attributes of the input data.\n",
```
OwenArnold, line 118:
```suggestion
    "The parameters given in the table above are retrieved from the coordinates or attributes of the input data.\n",
```

docs/user-guide/unit-conversions.ipynb
```diff
@@ -45,6 +47,118 @@
...
+    "dot.edge('Ltotal', 'energy')\n",
+    "dot.edge('L1', 'energy_transfer (direct)')\n",
+    "dot.edge('L2', 'energy_transfer (direct)')\n",
+    "dot.edge('incident_energy', 'energy_transfer (direct)')\n",
+    "dot.edge('L1', 'energy_transfer (indirect)')\n",
+    "dot.edge('L2', 'energy_transfer (indirect)')\n",
+    "dot.edge('final_energy', 'energy_transfer (indirect)')\n",
+    "dot.edge('Ltotal', 'wavelength')\n",
+    "dot.edge('Ltotal', 'dspacing')\n",
+    "dot.edge('two_theta', 'dspacing')\n",
```
OwenArnold, line 93: Given you are creating dot diagrams for the docs. Maybe it's worth adding a simple schematic to show 2theta as a projection with source sample, detector?

SimonHeybrock:
> > So strictly speaking the graph should have all possible combinations of units, but that would be a bit messy.
> 
> Maybe, not sure how messy it would be. I was just confused how you want to get a wavelength from just 'Ltotal'.

Even with *only* conversion from tof, it get much less readable and useful, I think: 
image

If you really wanted to do this you would need a separate figure for every possible pair of input and output unit.

neutron/beamline.cpp
```diff
@@ -27,6 +27,9 @@ VariableConstView sample_position(const dataset::CoordsConstView &meta) {
 
 Variable flight_path_length(const dataset::CoordsConstView &meta,
                             const ConvertMode scatter) {
+  // TODO Avoid copies here and below if scipp buffer ownership model is changed
+  if (meta.contains(Dim("Ltotal")))
```
OwenArnold, line 31: maybe move these names to be class/global variables. To prevent future bugs.

neutron/beamline.cpp
```diff
@@ -36,10 +39,16 @@ Variable flight_path_length(const dataset::CoordsConstView &meta,
...
+  if (meta.contains(Dim("L1")))
+    return copy(meta[Dim("L1")]);
+  return norm(incident_beam(meta));
 }
 
 Variable l2(const dataset::CoordsConstView &meta) {
+  if (meta.contains(Dim("L2")))
+    return copy(meta[Dim("L2")]);
+  if (meta.contains(Dim("scattered_beam")))
+    return norm(meta[Dim("scattered_beam")]);
```
OwenArnold, line 51: Redefining the meaning of l2 here https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/DetectorInfo.cpp#L122-L132

neutron/beamline.cpp
```diff
@@ -27,6 +27,9 @@ VariableConstView sample_position(const dataset::CoordsConstView &meta) {
 
 Variable flight_path_length(const dataset::CoordsConstView &meta,
                             const ConvertMode scatter) {
+  // TODO Avoid copies here and below if scipp buffer ownership model is changed
+  if (meta.contains(Dim("Ltotal")))
+    return copy(meta[Dim("Ltotal")]);
   // If there is not scattering this returns the straight distance from the
   // source, as required, e.g., for monitors or imaging.
   if (scatter == ConvertMode::Scatter)
```
OwenArnold, line 35: Linked to above ltotal = l1 + l2 with magnitude fixed to 2.

OwenArnold (changes requested):

docs/user-guide/unit-conversions.ipynb
SimonHeybrock: The names here correspond to what you need to specify as an arg to convert, in this case `scn.convert(data, 'wavelength', 'Q')`, not ||Q||.

docs/user-guide/unit-conversions.ipynb
SimonHeybrock: I think the documentation linked at the top of the page does a better job at that and avoid duplicate efforts?

neutron/beamline.cpp
SimonHeybrock: I don't understand your comment?

neutron/beamline.cpp
SimonHeybrock: I don't understand your comment.

neutron/beamline.cpp
OwenArnold: What you have is fine. I misunderstood what's happening here. returning magnitude nor normalising to (basis) vector.

SimonHeybrock:
Note that there is one more thing I need to add/fix here: Update mechanism for changing coord <-> attrs in the actual `convert` implementation.

neutron/beamline.cpp
```diff
@@ -54,15 +63,36 @@ Variable scattering_angle(const dataset::CoordsConstView &meta) {
...
-  const auto l1 = norm(beam);
-  beam /= l1;
-  auto scattered = position(meta) - sample_position(meta);
-  const auto l2 = norm(scattered);
-  scattered /= l2;
-
-  return acos(dot(beam, scattered));
+  if (meta.contains(Dim("two_theta")))
+    return copy(meta[Dim("two_theta")]);
+  return acos(cos_two_theta(meta));
```
OwenArnold, line 95: This whole mechanism seems good to me and gives good flexibility providing people know how to use it. I wonder if `convert` ought to have some kind of verbose mode where it was easy to determine what was being used. Maybe having the l1/l2 flight_path functions from beamline is enough for diagnostics though.

SimonHeybrock: I think this would link into a larger discussion on logging by scipp. Not logging from `scipp` is maybe still what we want, but logging from `scippneutron` could be useful?

OwenArnold: I agree it's part of a larger discussion. I think users would be very grateful if they could more easily get extra logging information for some of the operations like `convert`

SimonHeybrock:
> Note that there is one more thing I need to add/fix here: Update mechanism for changing coord <-> attrs in the actual `convert` implementation.

Done, this should be ready now. As before I am quite unhappy about the coord<->attr swapping mechanism, but this is not really new and the only option we could come up with.

docs/user-guide/unit-conversions.ipynb
OwenArnold: It's certainly more thorough. I was hoping to see something very simple like this https://docs.mantidproject.org/nightly/_images/SpecularReflectionCorrection_figure1.png

SimonHeybrock: We can certainly add figures (multiple would be need for the different cases), but this is outside the scope of the PR, I think.

OwenArnold (approved):

SimonHeybrock:
> > > So strictly speaking the graph should have all possible combinations of units, but that would be a bit messy.
> > 
> > 
> > Maybe, not sure how messy it would be. I was just confused how you want to get a wavelength from just 'Ltotal'.
> 
> Even with _only_ conversion from tof, it get much less readable and useful, I think:

I think you are right that it would be less confusing if we also included `tof`. I will try this in a follow-up, where I will fix some handling of inelastic conversions. Those will complicate the graphs, so I will probably split the elastic/inelastic graphs. Then we can maybe include `tof`.

neutron/beamline.cpp
```diff
@@ -36,10 +39,16 @@ Variable flight_path_length(const dataset::CoordsConstView &meta,
 }
 
 Variable l1(const dataset::CoordsConstView &meta) {
```
jl-wynen, line 41: Uppercase L1 and L2 to match the Python names?

jl-wynen (commented):
Maybe add functions that store results of conversions into datasets, i.e. the equivalent of
```.py
dset.attrs['Ltotal'] = Ltotal(dset)
```
This would prevent potential mistakes from using the wrong name.

neutron/beamline.cpp
SimonHeybrock: Unless you are happy with putting this into `scipp` it cannot be `constexpr`, because `Dim(string)` is not.

SimonHeybrock: ... or did you mean a `constexpr char *`, so we'd use `Dim(labels::L2)`?

jl-wynen: Did not think about this issue. But just the string would work.

SimonHeybrock: I'll address this in a follow-up I am working on, maybe I can come up with a better mechanism...

PR #30, opened 2021-03-09
Convert dataset slice and other improvements

SimonHeybrock:
- Fixes scipp/scipp#1476
- Fix infinite recursion (kernel crash) when providing unknown dimension label.
- Print known labels in exceptions.


neutron/convert.cpp
```diff
@@ -175,14 +179,35 @@ void check_scattering(const Dim from, const Dim to, const ConvertMode scatter) {
         "` is only defined for a scattering process.");
 }
 
+void check_label(const Dim dim, const std::string &name) {
+  const std::array<Dim, 6> known{
+      NeutronDim::Tof,      NeutronDim::Energy, NeutronDim::Wavelength,
+      NeutronDim::DSpacing, NeutronDim::Q,      NeutronDim::EnergyTransfer};
+  if (std::find(known.begin(), known.end(), dim) == known.end()) {
+    std::ostringstream oss;
+    oss << "Unsupported " << name << " dimension `" << dim
```
OwenArnold, line 188: 👍  for better error handling.

In addition we need to make these https://github.com/scipp/scippneutron/blob/0d9c7de076fb00ed014cc9c2f1b44c8ea725b9a0/neutron/beamline.cpp#L17-L37 available in the documentation so that users have a fair chance of getting the naming right rather than the trial and error (or look at source code) approach that would currently be required.

Even better (or in addition) we could expose a function to list these `known` dimensions so they can be used programatically.

python/tests/test_mantid.py
```diff
@@ -170,11 +171,14 @@ def test_inelastic_unit_conversion(self):
                 1e-8 * sc.units.meV +
                 1e-8 * sc.abs(ref.coords['energy_transfer']))).value
         assert sc.all(
-            low_tof | sc.isnan(da.bins.data.coords['energy_transfer'])
+            low_tof
+            | sc.isnan(da.bins.constituents['data'].coords['energy_transfer'])
             | sc.is_approx(
```
OwenArnold, line 176: You'll need to update this to `isclose`. You can split `rtol` and `atol` as separate args.

OwenArnold (changes requested):

neutron/convert.cpp
SimonHeybrock: They are, see figure here: https://scipp.github.io/scippneutron/user-guide/unit-conversions.html#Beamline-geometry-parameters-used-in-unit-conversions.

OwenArnold (approved):
Some other `is_approx` usage in tests on the c++ side are failing. Apart from that, looks good.

SimonHeybrock:
> Some other `is_approx` usage in tests on the c++ side are failing. Apart from that, looks good.

Please have another look, also had to update the Mantid-scipp tests.

python/tests/mantid_scipp_comparison/mantid_scipp_comparison.py
```diff
@@ -33,24 +34,22 @@ def _fuzzy_compare(self, a, b, tol):
                         ~sc.isfinite(y)).value > 0:
                     raise RuntimeError(
                         f'For meta {key} have non-finite entries')
-                coord_tol_factor = 1e-6
-                vtol = sc.abs(
-                    a.meta[key]
-                ) * coord_tol_factor + coord_tol_factor * a.meta[key].unit
-                if not sc.all(sc.is_approx(a.meta[key], b.meta[key],
-                                           vtol)).value:
+                if not sc.all(
```
OwenArnold, line 37: Thanks. I'll update this later to use `scipp._util.isnear` as that supersedes this.

OwenArnold (approved):

SimonHeybrock:
This is probably going to fail, since I need to update scipp to work around the issue that we cannot have `counts**2` as a unit right now.

PR #31, opened 2021-03-11
Add crude logging on which beamline params are used in unit conversion

SimonHeybrock:
Would need a real logger to do this properly, which is a larger discussion. One option to consider for this is to wrap the `logging` Python module.

As it is now, there is no way of disabling the info messages, so it may be too noisy?


CMakeLists.txt
```diff
@@ -3,9 +3,13 @@
...
-project(scippneutron)
-set(version 0.0.1)
-add_definitions(-DSCIPPNEUTRON_VERSION="${version}")
+execute_process(
+  COMMAND git describe --tag --abbrev=0
+  WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+  OUTPUT_VARIABLE SCIPPNEUTRON_VERSION
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+project(scippneutron VERSION ${SCIPPNEUTRON_VERSION})
```
OwenArnold, line 12: This will be unset if `git describe --tag --abrev=0` fails. That would also result in no version information being written to output files. Should we explicitly error with a non-zero exit status?

Practically the only way I could see that happening would be if someone was working outside of git (download source for a version) or locally failed to fetch (or actively removed tags).

neutron/beamline.cpp
```diff
@@ -56,26 +57,42 @@ Variable Ltotal(const dataset::CoordsConstView &meta,
     return copy(meta[NeutronDim::Ltotal]);
   // If there is not scattering this returns the straight distance from the
   // source, as required, e.g., for monitors or imaging.
-  if (scatter == ConvertMode::Scatter)
+  if (scatter == ConvertMode::Scatter) {
+    logging::info() << NeutronDim::Ltotal
```
OwenArnold, line 61: I think you should also log where the geometry information is being taken from. Clearly the logging mechanism will need to be updated later to give better control.

neutron/beamline.cpp
```diff
@@ -90,12 +107,20 @@ Variable scattering_angle(const dataset::CoordsConstView &meta) {
 Variable incident_beam(const dataset::CoordsConstView &meta) {
   if (meta.contains(NeutronDim::IncidentBeam))
     return copy(meta[NeutronDim::IncidentBeam]);
+  logging::info() << NeutronDim::IncidentBeam
+                  << " not found, trying to compute from "
```
OwenArnold, line 111: " not found, trying to compute from " is repeated many times. Is think there is a helper function missing here.

OwenArnold (changes requested):
Looks good so far. The basic logging itself is a useful step forward.

CMakeLists.txt
SimonHeybrock: Indeed unlikely, so I'd prefer keeping the code short?

neutron/beamline.cpp
SimonHeybrock: So I should write something like `"No coord or attr named Ltotal found, ..."?

OwenArnold: For the case it's not there yes.

If it is found maybe something like "Ltotal coord or attr found. Using directly"

SimonHeybrock: Would about double the number of logged lines, is that desirable?

OwenArnold: Ultimately maybe not. But i'm worried that users will try to run this and not understand where it's taking it's information from, waste time figuring it out (for example trying to correct positions when they infact have separate l1 and l2 set) and then getting annoyed that it wasn't reported to them. What I'm saying is that I'd prefer cautioning on the side of being verbose with this operation.

OwenArnold (approved):

PR #33, opened 2021-03-12
Parse transformations in load nexus

### LINKED ISSUE #19: Parse transformation chains in load_nexus
`NXtransformations` chains need to be parsed and used when loading sample, source and detector positions.
Initial implementation could give up if a multivalued `NXlog` is encountered in the chain.

For ticket to be complete, `load_nexus` must be capable of loading:
- sample position,
- source position,
- and pixel positions
which depend on a chain of transformations.

matthew-d-jones:
Closes #19

Add support for parsing chains of transformations from NeXus. This allows loading source, sample and pixel positions.
Only ~200 added lines are the implementation, ~900 lines added to test code to cover the many edge cases.

This does not yet cover the case that a transformation has multiple values (scanned motion axis), #34.

I'm pretty sure `np.einsum` or similar could be used to apply the transformation to pixel positions more efficiently, let me know if you have an idea, but time taken appears to be small in comparison to loading event data anyway.

### DIFF
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -142,7 +142,7 @@ def _load_event_group(group: h5py.Group, file_root: h5py.File,
     detector_id_type = ensure_supported_int_type(detector_ids.dtype.type)
     event_id_type = ensure_supported_int_type(event_id_ds.dtype.type)
     _check_event_ids_and_det_number_types_valid(detector_id_type,
-                                                event_id_type, group.name)
+                                                event_id_type)
 
     detector_ids = sc.Variable(dims=[_detector_dimension],
                                values=detector_ids,
@@ -172,8 +172,7 @@ def _load_event_group(group: h5py.Group, file_root: h5py.File,
 
 
 def _check_event_ids_and_det_number_types_valid(detector_id_type: np.dtype,
-                                                event_id_type: np.dtype,
-                                                group_name: str):
+                                                event_id_type: np.dtype):
     """
     These must be integers and must be the same type or we'll have
     problems trying to bin events by detector id. Check here so that
@@ -182,16 +181,14 @@ def _check_event_ids_and_det_number_types_valid(detector_id_type: np.dtype,
     """
     if not np.issubdtype(detector_id_type, np.integer):
         raise BadSource(
-            f"detector_numbers dataset in NXdetector is not an integer "
-            f"type, skipping loading {group_name}")
+            "detector_numbers dataset in NXdetector is not an integer "
+            "type")
     if not np.issubdtype(event_id_type, np.integer):
-        raise BadSource(f"event_ids dataset is not an integer type, "
-                        f"skipping loading {group_name}")
+        raise BadSource("event_ids dataset is not an integer type")
     if detector_id_type != event_id_type:
         raise BadSource(
-            f"event_ids and detector_numbers datasets in corresponding "
-            f"NXdetector were not of the same type, skipping "
-            f"loading {group_name}")
+            "event_ids and detector_numbers datasets in corresponding "
+            "NXdetector were not of the same type")
 
 
 def load_detector_data(event_data_groups: List[h5py.Group],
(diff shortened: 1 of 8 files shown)


python/src/scippneutron/_loading_transformations.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    total_transform_matrix = get_full_transformation_matrix(group, root)
+    return np.matmul(total_transform_matrix, np.array([0, 0, 0, 1],
+                                                      dtype=float))[0:3]
+
+
+def get_full_transformation_matrix(group: h5py.Group,
+                                   root: h5py.File) -> np.ndarray:
+    """
+    Get the 4x4 transformation matrix for a component, resulting
```
SimonHeybrock, line 56: So this is an affine transformation, not just a rotation?

python/src/scippneutron/_loading_transformations.py
```diff
@@ -0,0 +1,230 @@
...
+        offset[0]
+    ],
+                       [
+                           rotation_matrix[1, 0], rotation_matrix[1, 1],
+                           rotation_matrix[1, 2], offset[1]
+                       ],
+                       [
+                           rotation_matrix[2, 0], rotation_matrix[2, 1],
+                           rotation_matrix[2, 2], offset[2]
+                       ], [0., 0., 0., 1.]])
```
SimonHeybrock, line 229: Does https://numpy.org/doc/stable/reference/generated/numpy.block.html work here?

python/src/scippneutron/_loading_transformations.py
```diff
@@ -0,0 +1,230 @@
...
+
+    if unit != sc.units.m:
+        magnitude_var = magnitude * unit
+        magnitude_var = sc.to_unit(magnitude_var, sc.units.m)
+        magnitude = magnitude_var.value
+    # -1 as describes passive transformation
+    vector = direction_unit_vector * -1. * magnitude
+    matrix = np.array([[1., 0., 0., vector[0] + offset[0]],
+                       [0., 1., 0., vector[1] + offset[1]],
+                       [0., 0., 1., vector[2] + offset[2]], [0., 0., 0., 1.]])
```
SimonHeybrock, line 155: `np.block`?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -65,13 +66,25 @@ def _load_positions(detector_group: h5py.Group,
...
 
+    array = np.array([x_positions, y_positions, z_positions]).T
+
     if "depends_on" in detector_group:
-        warn(f"Loaded pixel positions for "
-             f"{detector_group.name.split('/')[-1]} are relative to the "
-             f"detector, not sample position, as parsing transformations "
-             f"is not yet implemented")
+        # Add fourth element of 1 to each vertex, indicating these are
+        # positions not direction vectors
```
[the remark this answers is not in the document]
matthew-d-jones: I mean, if I put a 0 there then applied the transformation, it would be like acting on a vector; only rotations have an effect on it. With a 1 there, it acts as a position, both rotations and translations transform it.

python/src/scippneutron/_loading_transformations.py
matthew-d-jones: Yes, this can be the result of a whole string of transformations defined in the file.

python/tests/test_load_nexus.py
```diff
@@ -477,11 +553,392 @@ def test_load_nexus_does_not_load_pixel_positions_with_non_matching_shape():
...
+
+
+@pytest.mark.parametrize("component_class,component_name",
+                         ((Sample, "sample"), (Source, "source")))
+def test_skips_loading_component_if_more_than_one_in_file(
+        component_class: Union[Type[Source], Type[Sample]],
+        component_name: str):
+    # More than one source or sample is a serious error in the file, so
+    # load_nexus will display a warning and skip loading any sample rather
+    # than guessing which is the "correct" one.
```
SimonHeybrock, line 589: Mantid supports multiple samples in a workspace (or rather-child samples? I suppose this is used when it consists of multiple components). How is this represented in Nexus files?

matthew-d-jones: Thanks, I was not aware of this. I've posted a question in #tech-qa on Slack.

SimonHeybrock (approved):

PR #36, opened 2021-03-15
Fix comparisons of scipp vars with numpy arrays

matthew-d-jones:
No longer possible to directly compare scipp variable with a numpy array due to removal of `buffer_protocol` support from `VariableView`: https://github.com/scipp/scipp/pull/1766


OwenArnold:
Applied as part of #37

PR #38, opened 2021-03-15
Trigger main when scipp main completes

OwenArnold:
scipp **project**, Main **pipeline** is upstream trigger.

Help identify upstream builds that break things in scipp-neutron. This should be replaced/incorporated into the wider CI overhaul underway by @nvaytet 

**Tester**

It is possible to test these changes prior to merging, but it's convoluted, not worth the effort IMHO.

### DIFF
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -5,7 +5,7 @@ pr: none
 
 resources:
   pipelines:
-  - pipeline: upstream_scipp_release
+  - pipeline: upstream_scipp_main
     project: scipp
     source: Main
 
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -3,6 +3,12 @@ trigger:
 
 pr: none
 
+resources:
+  pipelines:
+  - pipeline: upstream_scipp_release
+    project: scipp
+    source: Main
+
 stages:
   - template: templates/code_quality_checks.yml
 


.azure-pipelines/main.yml
```diff
@@ -3,6 +3,12 @@ trigger:
 
 pr: none
 
+resources:
+  pipelines:
+  - pipeline: upstream_scipp_release
```
SimonHeybrock, line 8: Release?

SimonHeybrock (commented):
Is this triggering a new package build? Should it really? I suppose we are interested in two things:

- Does `scippneutron` still build and pass tests.
- Does the existing `scippneutron` conda package still work with the new uptream package?

It feels oddd to publish a new package for no reason.

OwenArnold:
> Is this triggering a new package build? Should it really? I suppose we are interested in two things:
> 
> * Does `scippneutron` still build and pass tests.
> * Does the existing `scippneutron` conda package still work with the new uptream package?
> 
> It feels oddd to publish a new package for no reason.

Maybe this is pointing at the need for a new pipleline type to handle this. I agree there is no need to build and publish. I especially like the idea of not having to build to check ABI compatibility.

.azure-pipelines/main.yml
SimonHeybrock:
```suggestion
  - pipeline: upstream_scipp_main
```

SimonHeybrock:
The new pipeline discussed is maybe the way to go long term, but for now this is simple and should do the job, so I merged this.

PR #39, opened 2021-03-16
Use sc.dtype.datetime64 for time series logs

OwenArnold:
Fixes https://github.com/scipp/scipp/issues/1698

### DIFF
--- a/python/src/scippneutron/mantid.py
+++ b/python/src/scippneutron/mantid.py
@@ -93,11 +93,8 @@ def make_variables_from_run_logs(ws):
             data_array = sc.DataArray(data=property_data,
                                       coords={
                                           dimension_label:
-                                          sc.Variable(
-                                              [dimension_label],
-                                              values=times,
-                                              dtype=sc.dtype.datetime64,
-                                              unit=sc.units.ns)
+                                          sc.Variable([dimension_label],
+                                                      values=times)
                                       })
             yield property_name, sc.Variable(value=data_array)
         elif not np.isscalar(values):
@@ -625,11 +622,8 @@ def convert_EventWorkspace_to_data_array(ws,
                                        dtype=sc.dtype.float32),
                            dims=['event'],
                            shape=[n_event])
-    pulse_times = sc.Variable(
-        dims=['event'],
-        shape=[n_event],
-        unit=sc.units.ns,
-        dtype=sc.dtype.datetime64) if load_pulse_times else None
+    pulse_times = sc.Variable(dims=['event'],
+                              shape=[n_event]) if load_pulse_times else None
 
     evtp = ws.getSpectrum(0).getEventType()
     contains_weighted_events = ((evtp == EventType.WEIGHTED)
--- a/python/src/scippneutron/mantid.py
+++ b/python/src/scippneutron/mantid.py
@@ -93,10 +93,11 @@ def make_variables_from_run_logs(ws):
             data_array = sc.DataArray(data=property_data,
                                       coords={
                                           dimension_label:
-                                          sc.Variable([dimension_label],
-                                                      values=times,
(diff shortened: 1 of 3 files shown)


python/src/scippneutron/mantid.py
```diff
@@ -93,10 +93,11 @@ def make_variables_from_run_logs(ws):
...
                                           dimension_label:
-                                          sc.Variable([dimension_label],
-                                                      values=times,
-                                                      dtype=sc.dtype.int64,
-                                                      unit=sc.units.ns)
+                                          sc.Variable(
+                                              [dimension_label],
+                                              values=times,
+                                              dtype=sc.dtype.datetime64,
+                                              unit=sc.units.ns)
```
SimonHeybrock, line 100: If `times` is a numpy array of `datetime`, are the explicit `dtype` and `unit` arguments required?

SimonHeybrock (changes requested):
Missing changes for `pulse_time`. Should be enough to just remove the current conversion to int we do there.

python/src/scippneutron/mantid.py
OwenArnold: dtype - no we don't need to have this. possibly makes things clearer but unnecessary otherwise. 
I'm not sure about unit - I'll see if I can remove this.

python/src/scippneutron/mantid.py
```diff
@@ -629,7 +629,7 @@ def convert_EventWorkspace_to_data_array(ws,
         dims=['event'],
```
SimonHeybrock, line 629: Unrelated, but we could use `sc.empty` instead of `sc.Variable` here and for `coord` to avoid the extra write to memory, for better performance.

SimonHeybrock: And we have `sc.ones` now, so we can avoid `sc.broadcast` above?

python/src/scippneutron/mantid.py
```diff
@@ -624,11 +622,8 @@ def convert_EventWorkspace_to_data_array(ws,
                                        dtype=sc.dtype.float32),
                            dims=['event'],
                            shape=[n_event])
-    pulse_times = sc.Variable(
-        dims=['event'],
-        shape=[n_event],
-        unit=sc.units.ns,
-        dtype=sc.dtype.int64) if load_pulse_times else None
+    pulse_times = sc.Variable(dims=['event'],
+                              shape=[n_event]) if load_pulse_times else None
```
SimonHeybrock, line 626: I don't think this is right, this will create with `dtype=double`.

python/src/scippneutron/mantid.py
```diff
@@ -616,14 +616,15 @@ def convert_EventWorkspace_to_data_array(ws,
...
-                                       variance=1.0,
-                                       unit=data_unit,
-                                       dtype=sc.dtype.float32),
-                           dims=['event'],
-                           shape=[n_event])
-    pulse_times = sc.Variable(dims=['event'],
-                              shape=[n_event]) if load_pulse_times else None
+    weights = sc.ones(dims=['event'],
+                      shape=[n_event],
+                      dtype=sc.dtype.float64,
```
SimonHeybrock, line 621: Should be float32 to avoid wasting memory?

python/src/scippneutron/mantid.py
```diff
@@ -618,7 +618,6 @@ def convert_EventWorkspace_to_data_array(ws,
                         dtype=sc.dtype.float64)
     weights = sc.ones(dims=['event'],
                       shape=[n_event],
-                      dtype=sc.dtype.float64,
```
SimonHeybrock: Why? This will be `double`, we want it to be `float32`, as before?

SimonHeybrock (approved):

PR #43, opened 2021-03-25
Set unit on pixel positions in load_nexus

### LINKED ISSUE #42: Unit missing on positions created by load_nexus
and leads to `scn.convert(data, 'tof', 'wavelength', scatter=True)` failing.

matthew-d-jones:
Closes #42 

Units for pixel positions were not being checked in file or set in the scipp datastructure during `load_nexus()`.

Changes:
- Loading pixel positions is skipped and a warning output if the "units" attribute is missing.
- Units are checked and offsets are converted to metres if they are not already in metres. Conversion to metres is done to ensure consistent units are used for transformation calculations.
- New behaviour is covered by unit tests.

### DIFF
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -6,7 +6,8 @@ from dataclasses import dataclass
 import h5py
 from typing import Optional, List
 import numpy as np
-from ._loading_common import BadSource, ensure_supported_int_type, load_dataset
+from ._loading_common import (BadSource, ensure_supported_int_type,
+                              load_dataset, get_units)
 import scipp as sc
 from datetime import datetime
 from warnings import warn
@@ -50,6 +51,18 @@ def _iso8601_to_datetime(iso8601: str) -> Optional[datetime]:
 
 def _load_pixel_positions(detector_group: h5py.Group, detector_ids_size: int,
                           file_root: h5py.File) -> Optional[sc.Variable]:
+    offsets_unit = get_units(detector_group["x_pixel_offset"])
+    if not offsets_unit:
+        warn(f"Skipped loading pixel positions as no units found on "
+             f"x_pixel_offset dataset in {detector_group.name}")
+        return None
+    if offsets_unit != sc.units.m:
+        warn(f"Skipped loading pixel positions as units found on "
+             f"x_pixel_offset dataset in {detector_group.name} are "
+             f"not supported. Only metres are currently supported "
+             f"for offsets datasets.")
+        return None
+
     try:
         x_positions = detector_group["x_pixel_offset"][...].flatten()
         y_positions = detector_group["y_pixel_offset"][...].flatten()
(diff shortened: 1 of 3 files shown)


python/src/scippneutron/_loading_detector_data.py
```diff
@@ -66,6 +83,10 @@ def _load_pixel_positions(detector_group: h5py.Group, detector_ids_size: int,
              f"dataset sizes do not match in {detector_group.name}")
         return None
 
+    x_positions = _convert_array_to_metres(x_positions, offsets_unit)
+    y_positions = _convert_array_to_metres(y_positions, offsets_unit)
+    z_positions = _convert_array_to_metres(z_positions, offsets_unit)
+
     array = np.array([x_positions, y_positions, z_positions]).T
```
SimonHeybrock, line 90: So I looked in `get_full_transformation_matrix` called below, and there is a similar mechanism for converting to `m` for the "translation" part of the transformation. Are both reading the same unit actually? If so, we could convert to `m` only at the very end, instead of doing the numpy->scipp->numpy conversion here?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -48,8 +49,24 @@ def _iso8601_to_datetime(iso8601: str) -> Optional[datetime]:
...
 
 
+def _convert_array_to_metres(array: np.ndarray, unit: str) -> np.ndarray:
+    if sc.Unit(unit) != sc.units.m:
+        return sc.to_unit(
+            sc.Variable(["temporary_variable"],
+                        values=array,
+                        unit=unit,
+                        dtype=np.float64), "m").values
+    return array
```
SimonHeybrock, line 59: I might suggest
```suggestion
        return sc.to_unit(
            sc.Variable(["temporary_variable"],
                        values=array,
                        unit=unit,
                        dtype=np.float64), "m").values
```
since with the rewrite for shallow copying we can refactor `to_unit` to not make a copy if the unit is unchanged.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: No, the units can be different, and there can be 100 different translations, with 100 different units, to apply to these pixels positions. Converting them all to metres first is the only sensible way I could think to handle that.

SimonHeybrock: Must be a joy working with files and fileformats!

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -50,13 +50,11 @@ def _iso8601_to_datetime(iso8601: str) -> Optional[datetime]:
 
 
 def _convert_array_to_metres(array: np.ndarray, unit: str) -> np.ndarray:
-    if sc.Unit(unit) != sc.units.m:
         return sc.to_unit(
             sc.Variable(["temporary_variable"],
                         values=array,
                         unit=unit,
                         dtype=np.float64), "m").values
```
SimonHeybrock, line 57: Bad indent now? Sorry, didn't include it in my change suggestion.

SimonHeybrock (approved):

PR #45, opened 2021-03-26
Test load_nexus handling of links

### LINKED ISSUE #44: Presence of links in file can cause load_nexus to fail
Torben reports that creating a link to an `NXevent_data` group in a file causes scipp to fail in `load_nexus`.

My guess is that `h5py.visit` now visits links as well as real datasets and groups. May need to check for, and ignore, links in `_loading_common.find_by_nx_class()`.

```
is_soft_link = isinstance(fh.get('soft_link', getlink=True), h5py.SoftLink)
```

matthew-d-jones:
Closes #44 

Turns out Scipp was doing the right thing (ignore links) already, but I've added a unit test to document it.

### DIFF
--- a/python/tests/nexus_helpers.py
+++ b/python/tests/nexus_helpers.py
@@ -91,6 +91,12 @@ class Source:
     distance_units: Optional[str] = None
 
 
+@dataclass
+class Link:
+    new_path: str
+    target_path: str
+
+
 def _add_event_data_group_to_file(data: EventData, parent_group: h5py.Group,
                                   group_name: str):
     event_group = _create_nx_class(group_name, "NXevent_data", parent_group)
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Do we know why/how it s doing the right thing? Is h5py ignoring them? Is it some missing attributes that links don't have?

matthew-d-jones:
> Do we know why/how it s doing the right thing? Is h5py ignoring them? Is it some missing attributes that links don't have?

I'm using `h5py.visit` to find groups with an `NX_class` that Scipp knows how to to load. `visit` ignores links, probably to avoid getting caught in loops.

SimonHeybrock (approved):

PR #47, opened 2021-04-07
Add data streaming interface to Scipp

matthew-d-jones:
Closes https://github.com/scipp/scipp/issues/1448

Intentionally not documented yet, but can be used like this (assuming there is a Kafka broker running on `localhost` with a topic called "event_data"):
```python
import asyncio

async def my_stream_func():
    async for data in scn.data_stream('localhost:9092', ['event_data'], interval_s=3.):
        print(data)
    
streaming_task = asyncio.create_task(my_stream_func())
```

In this first step of the implementation only event data are consumed from Kafka. Behaviour if any other data types are encountered is documented in the unit tests.

Three dependencies were added:
- `confluent-kafka` is the Kafka client library, there are other python implementations but this is the one used in ECDC software such as https://github.com/ess-dmsc/forwarder/
- `ess-streaming-data-types` implements serialisation and deserialisation of the data formats used in the ESS data streaming system
- `pytest-asyncio` is a plugin for pytest to facilitate testing async code

I could provide a `docker-compose` script that would launch a Kafka broker and populate it with event data if anyone wants to test on their machine.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -19,9 +19,6 @@ requirements:
     - python {{ python }}
     - scipp
     - h5py
-    - pip:
-        - confluent-kafka
-        - ess-streaming-data-types
 
 test:
   imports:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -19,8 +19,8 @@ requirements:
     - python {{ python }}
     - scipp
     - h5py
-    - python-confluent-kafka
     - pip:
+        - confluent-kafka
         - ess-streaming-data-types
 
 test:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -20,6 +20,8 @@ requirements:
     - scipp
     - h5py
     - python-confluent-kafka
+    - pip:
+        - ess-streaming-data-types
 
 test:
   imports:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -19,6 +19,7 @@ requirements:
     - python {{ python }}
     - scipp
     - h5py
+    - python-confluent-kafka
 
 test:
   imports:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -33,6 +33,7 @@ test:
     - matplotlib-base
     - psutil
     - pytest
+    - pytest-asyncio
     - pythreejs
     - Pillow
     - mantid-framework [not win]
--- a/python/src/scippneutron/_streaming_data_buffer.py
+++ b/python/src/scippneutron/_streaming_data_buffer.py
@@ -93,15 +93,15 @@ class StreamedDataBuffer:
                 if self._current_event + message_size > self._buffer_size:
                     await self._emit_data()
                 async with self._buffer_mutex:
-                    self._events_buffer.coords['detector_id'][
-                        'event', self._current_event:self._current_event +
-                        message_size] = deserialised_data.detector_id
-                    self._events_buffer.coords['tof'][
-                        'event', self._current_event:self._current_event +
-                        message_size] = deserialised_data.time_of_flight
-                    self._events_buffer.coords['pulse_time'][
-                        'event', self._current_event:self._current_event +
-                        message_size] = deserialised_data.pulse_time * \
+                    frame = self._events_buffer[
+                        'event',
+                        self._current_event:self._current_event + message_size]
+                    frame.coords[
+                        'detector_id'].values = deserialised_data.detector_id
+                    frame.coords[
+                        'tof'].values = deserialised_data.time_of_flight
+                    frame.coords['pulse_time'].values = \
+                        deserialised_data.pulse_time * \
                         np.ones_like(deserialised_data.time_of_flight)
                     self._current_event += message_size
                 return
--- a/python/src/scippneutron/_streaming_data_buffer.py
+++ b/python/src/scippneutron/_streaming_data_buffer.py
@@ -1,7 +1,6 @@
 import scipp as sc
 import numpy as np
 import asyncio
-from scipp.detail import move_to_data_array
 from streaming_data_types.eventdata_ev42 import deserialise_ev42
 from streaming_data_types.exceptions import WrongSchemaException
 from typing import Optional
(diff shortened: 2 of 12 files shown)


SimonHeybrock:
> I could provide a docker-compose script that would launch a Kafka broker and populate it with event data if anyone wants to test on their machine.

I think this would be extremely useful, especially since you will be moving to another project soon.

conda/meta.yaml
```diff
@@ -19,6 +19,9 @@ requirements:
     - python {{ python }}
     - scipp
     - h5py
+    - pip:
+        - confluent-kafka
+        - ess-streaming-data-types
```
SimonHeybrock, line 24: Hmm, aren't these dependencies optional and required only if the streaming is used?

python/src/scippneutron/data_stream.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Periodically yields accumulated data from stream.
+    If the buffer fills up more frequently than the set interval
+    then data is yielded more frequently.
+    1048576 event buffer is around 24 MB (with pulse_time, id, weights, etc)
+    :param kafka_broker: Address of the Kafka broker to stream data from
+    :param topics: Kafka topics to consume data from
+    :param buffer_size: Size of buffer to accumulate data in
+    :param interval_s: interval between yielding any new data
+      collected from stream
```
SimonHeybrock, line 58: Is the `_s` suffix indicating this is seconds? Would it be worth using a scalar variable here, and then something like `interval_s=sc.to_unit(interval, 's').value` below (edit: or better change `StreamedDataBuffer` to use a scalar variable, you get the point)? Would be safe and self-documenting.

python/src/scippneutron/_streaming_data_buffer.py
```diff
@@ -0,0 +1,118 @@
+import scipp as sc
+import numpy as np
+import asyncio
+from scipp.detail import move_to_data_array
```
SimonHeybrock, line 4: I deleted this today in the buffer-sharing rewrite, so maybe just don't use it here, since it will be redundant.

python/src/scippneutron/_streaming_data_buffer.py
```diff
@@ -0,0 +1,118 @@
...
+                                unit=sc.units.one,
+                                dtype=sc.dtype.int32)
+        pulse_times = sc.Variable(dims=['event'],
+                                  shape=[buffer_size],
+                                  unit=sc.units.ns,
+                                  dtype=sc.dtype.int64)
+        weights = sc.Variable(dims=['event'],
+                              unit=sc.units.one,
+                              values=np.ones(buffer_size, dtype=np.float32),
+                              variances=np.ones(buffer_size, dtype=np.float32))
```
SimonHeybrock, line 40: Consider using `sc.ones` and `sc.zeros`.

python/src/scippneutron/_streaming_data_buffer.py
```diff
@@ -0,0 +1,118 @@
...
+                         " messages with unrecognised FlatBuffer ids")
+                    self._unrecognised_fb_id_count = 0
+                if self._current_event == 0:
+                    return
+                new_data = self._events_buffer[
+                    'event', :self._current_event].copy()
+                self._current_event = 0
+            self._emit_queue.put_nowait(new_data)
+        except Exception as e:
+            print(e)
```
SimonHeybrock, line 80: Can you elaborate (or add a comment) on why exceptions or just ignored/printed?

python/src/scippneutron/_streaming_data_buffer.py
```diff
@@ -0,0 +1,118 @@
...
+                    self._events_buffer.coords['pulse_time'][
+                        'event', self._current_event:self._current_event +
+                        message_size] = deserialised_data.pulse_time * \
+                        np.ones_like(deserialised_data.time_of_flight)
+                    self._current_event += message_size
+                return
+            except WrongSchemaException:
+                self._unrecognised_fb_id_count += 1
+        except Exception as e:
+            print(e)
```
SimonHeybrock, line 118: Same question as above. Why are exceptions "ignored"?

python/src/scippneutron/_streaming_data_buffer.py
```diff
@@ -0,0 +1,118 @@
...
+                    self._events_buffer.coords['detector_id'][
+                        'event', self._current_event:self._current_event +
+                        message_size] = deserialised_data.detector_id
+                    self._events_buffer.coords['tof'][
+                        'event', self._current_event:self._current_event +
+                        message_size] = deserialised_data.time_of_flight
+                    self._events_buffer.coords['pulse_time'][
+                        'event', self._current_event:self._current_event +
+                        message_size] = deserialised_data.pulse_time * \
+                        np.ones_like(deserialised_data.time_of_flight)
```
SimonHeybrock, line 112: Might be more readable if written as
```python
frame = self._events_buffer['event', self._current_event:self._current_event + message_size]
frame.coords['detector_id'] = deserialised_data.detector_id
...
```

matthew-d-jones:
> > I could provide a docker-compose script that would launch a Kafka broker and populate it with event data if anyone wants to test on their machine.
> 
> I think this would be extremely useful, especially since you will be moving to another project soon.

I will make another PR to record in the developer docs how to test the live data using docker #49

conda/meta.yaml
matthew-d-jones: Yes, but these are not Mantid-size dependencies; `ess-streaming-data-types` is 80kb and `confluent-kafka` <2MB.
https://pypi.org/pypi/ess-streaming-data-types/json
https://pypi.org/pypi/confluent-kafka/json
Is it worth making them require separate manual installation?

SimonHeybrock: We have generally tried doing so (e.g., for all the plotting dependencies in scipp). Unless you have a strong opinion I would try to make them optional (not dependencies of the conda package), but include them in the `yml` files that are part of the docs.

SimonHeybrock: I am referring to there: https://github.com/scipp/scippneutron/tree/main/docs/environments

python/src/scippneutron/_streaming_data_buffer.py
matthew-d-jones: I had to do
```python
frame.coords['detector_id'].values = deserialised_data.detector_id
```

python/src/scippneutron/_streaming_data_buffer.py
matthew-d-jones: This `try-except` was mistakenly left there when I changed the implementation of this method.

python/src/scippneutron/_streaming_data_buffer.py
matthew-d-jones: Same as above.

SimonHeybrock (approved):

PR #52, opened 2021-04-09
Use correct conda channels in CI

SimonHeybrock:
- PR and `Main` must use `scipp/label/dev`, also for doc build
- `Release` must use `scipp`, also for doc build (configured via new param to template).


matthew-d-jones (approved):
LGTM. I will update #54 once this is merged.

scippneutron-developer.yml
```diff
@@ -6,7 +6,7 @@ name: scippneutron-developer
 
 channels:
   - conda-forge
-  - scipp
+  - scipp/label/dev
```
OwenArnold, line 9: Assumes that `mantid-framework` is in `scipp/label/dev` channel. Might be better to now explicitly change `- mantid-framework` to be `- scipp::mantid-framework` so we don't have to duplicate it in the dev channel just for the purposes of making this work.

SimonHeybrock:
As discussed on the call:
- Updated to use `-c scipp` for PR, Main, and Release pipelines.
- Still want to trigger Main from upstream (scipp) Main, using `-c scipp/label/dev`. I added a parameter to allow for configuring this, but did not take the final step, since it would require a lot of duplicate parameter checking. I put some TODOs in place and hope that we can do this in a nicer way once @nvaytet's CI cleanup is in place. This means that we temporarily have no upstream trigger.

SimonHeybrock:
I had temporarily added a `scippneutron-developer-latest.yml`, but in the last iteration this is not necessary any more since the `Main` pipeline does not use these environments (using `conda build`).

.azure-pipelines/templates/documentation_build.yml
```diff
@@ -1,3 +1,8 @@
+parameters:
+- name: conda_env
+  type: string
+  default: scippneutron-developer.yml
```
SimonHeybrock, line 4: This is not really needed in the latest iteration, but maybe we can leave it anyway, so this can be configured more easily in the future?

matthew-d-jones (approved):

PR #53, opened 2021-04-09
If duplicates, rename to avoid name clashes

### LINKED ISSUE #48: Monitors with duplicate component names disappear
When loading data from POLARIS, only a single monitor attr `monitor` appears in the laoded file. However, there should be 14 monitors.

The problem is `mantid.py` line 549: Apparently the name of the monitors in the component info is not unique.

Solution:
- Detect this and include the detector ID in the attr name in this case?
- Other?

OwenArnold:
Fixes #48 

Solution uses `uniqueName` added to Mantid's ComponentInfo, and therefore needs packages produced from [1.0.3](https://github.com/scipp/mantid_framework_conda_recipe/tree/1.0.3) of the mantid-framework recipe, which has been used to create a new `mantid-framework` version `6.0.20210412.0948` for all platforms. Following this, developers will need to regenerate their `scippneutron-developer` environments to collect the updated mantid-framework package.

Solution uses https://github.com/mantidproject/mantid/pull/31097

### DIFF
--- a/python/tests/test_mantid.py
+++ b/python/tests/test_mantid.py
@@ -929,5 +929,18 @@ def test_EventWorkspace_with_pulse_times():
                   [0].to_datetime64()))
 
 
+@pytest.mark.skipif(not mantid_is_available(),
+                    reason='Mantid framework is unavailable')
+def test_duplicate_monitor_names():
+    from mantid.simpleapi import LoadEmptyInstrument
+    ws = LoadEmptyInstrument(
+        InstrumentName='POLARIS',
+        StoreInADS=False)  # Has many monitors named 'monitor'
+    da = scn.mantid.from_mantid(ws)
+    assert da.attrs['monitor_0'].value.attrs['spectrum'].value == 1
+    assert da.attrs['monitor_12'].value.attrs['spectrum'].value == 13
+    assert da.attrs['monitor'].value.attrs['spectrum'].value == 14
+
+
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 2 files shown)


python/tests/test_mantid.py
```diff
@@ -929,5 +929,18 @@ def test_EventWorkspace_with_pulse_times():
...
+                    reason='Mantid framework is unavailable')
+def test_duplicate_monitor_names():
+    from mantid.simpleapi import LoadEmptyInstrument
+    ws = LoadEmptyInstrument(
+        InstrumentName='POLARIS',
+        StoreInADS=False)  # Has many monitors named 'monitor'
+    da = scn.mantid.from_mantid(ws)
+    assert da.attrs['monitor_0'].value.attrs['spectrum'].value == 1
+    assert da.attrs['monitor_12'].value.attrs['spectrum'].value == 13
+    assert da.attrs['monitor'].value.attrs['spectrum'].value == 14
```
OwenArnold, line 942: With https://github.com/mantidproject/mantid/pull/31097 All monitors are rename if any are duplicates:

```python
assert da.attrs['monitor_13'].value.attrs['spectrum'].value == 14
```

And the code to do the replacement would instead use
```python
if not comp_info.uniqueName(name):
    name = f'{name}_{det_index}'
```

SimonHeybrock: Wouldn't it be more natural of the montitor name matched the spectrum number?

SimonHeybrock (approved):

PR #54, opened 2021-04-10
Get live data dependencies from conda

matthew-d-jones:
Will need to be updated after #52 is merged (change required in  `scippneutron-developer-release.yml`)

The ESS serialisation lib is now on conda. The Kafka library is also on conda but not for Windows, so I've handled this dependency in the same way as we do for Mantid.

Perhaps it would be better to use the `ess-dmsc::ess-streaming-data-types` syntax rather than adding the extra channel everywhere for a single dependency?

### DIFF
--- a/.azure-pipelines/templates/test_can_install_and_import.yml
+++ b/.azure-pipelines/templates/test_can_install_and_import.yml
@@ -36,7 +36,7 @@ jobs:
           set -ex
           conda create -c conda-forge -n test python=3.7
           source activate test
-          conda install -c "$(packages_dir)" -c conda-forge -c scipp/label/dev scippneutron
+          conda install -c "$(packages_dir)" -c conda-forge -c scipp/label/dev -c ess-dmsc scippneutron
         displayName: 'Install Conda package'
       - bash: |
           set -ex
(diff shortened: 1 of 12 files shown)


SimonHeybrock:
> Perhaps it would be better to use the `ess-dmsc::ess-streaming-data-types` syntax rather than adding the extra channel everywhere for a single dependency?

Sounds like a good suggestion to me?

matthew-d-jones:
@SimonHeybrock I've made that change in the env files, but it is not possible to do it for `conda build` (https://github.com/conda/conda-build/issues/532)

.azure-pipelines/templates/test_can_install_and_import.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           set -ex
           conda create -c conda-forge -n test python=3.7
           source activate test
-          conda install -c "$(packages_dir)" -c conda-forge -c scipp/label/dev scippneutron
+          conda install -c "$(packages_dir)" -c conda-forge -c scipp/label/dev -c ess-dmsc scippneutron
```
SimonHeybrock, line 39: Shouldn't we rather test that it can be installed/imported *without* the optional dependency?

SimonHeybrock (approved):

PR #55, opened 2021-04-12
Document testing live data

### LINKED ISSUE #49: Document how to test live data interface
Record in the developer documentation how to test the live data interface by launching the NeXus-Streamer and Kafka using a docker-compose script.

matthew-d-jones:
Closes #49

Adds a docker-compose script and documentation on how to use it to test the live data functionality.

### DIFF
--- a/docs/developer/testing-live-data.rst
+++ b/docs/developer/testing-live-data.rst
@@ -109,4 +109,5 @@ After you are done testing you can clean up the containers and free up used disk
 
     .. code-block:: sh
 
-        docker system prune -a --volumes
+        docker rm -v live_data_producer_1
+        docker rm -v live_data_kafka_1
--- a/docs/developer/testing-live-data.rst
+++ b/docs/developer/testing-live-data.rst
@@ -2,7 +2,7 @@ Testing "live data"
 ===================
 
 The ESS data streaming system is based on the Apache Kafka data streaming platform.
-Testing scippneutron's interface to this system requires running a Kafka server and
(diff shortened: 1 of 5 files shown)


docs/developer/testing-live-data.rst
```diff
@@ -0,0 +1,112 @@
...
+          time_of_flight: [     12379936     14495801     14658190     ... truncated 756 elements ...     36832880   ]
+        }
+        ...
+
+Try using ``scippneutron.data_stream``, for example
+
+    .. code-block:: python
+
+        import asyncio
+
```
nvaytet, line 91: Add `import scippneutron as scn` ?

docs/developer/testing-live-data.rst
```diff
@@ -0,0 +1,112 @@
...
+will use more and more disk space until you terminate docker-compose.
+
+Clean Up
+~~~~~~~~
+
+After you are done testing you can clean up the containers and free up used disk space by running
+
+    .. code-block:: sh
+
+        docker system prune -a --volumes
```
nvaytet, line 112: Does this have the potential to delete some data in other (unrelated) docker containers that the user has created from other projects?

nvaytet (changes requested):
Thanks for writing this down very clearly.
Just a few comments.

docs/developer/testing-live-data.rst
matthew-d-jones: I've replaced it with commands to specifically remove the live data containers and their volumes

nvaytet (approved):

PR #57, opened 2021-04-13
Release notes for duplicate monitor fix

OwenArnold:
Follow on from #53 

Add info about fix to coming release notes.

### DIFF
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -9,13 +9,13 @@ Since v0.1.0
 Features
 ~~~~~~~~
 
-Breaking changes
-~~~~~~~~~~~~~~~~
-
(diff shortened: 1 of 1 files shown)


SimonHeybrock (commented):
I think I'd list it under a `Bugfixes` section, rather than as a breaking change?

OwenArnold:
> I think I'd list it under a `Bugfixes` section, rather than as a breaking change?

I don't have any strong opinon on this. Originally put as breaking change, because if you previously relied on extracting the single monitor as "monitor" that no longer exists. On the other hand, the fact that you were not properly guaranteed which monitor that was, it probably does represent a bugfix.

SimonHeybrock (approved):

Issue #56, opened 2021-04-12
Parse run start message in data_stream


matthew-d-jones:
The [run start message](https://github.com/ess-dmsc/streaming-data-types/blob/master/schemas/pl72_run_start.fbs) contains a field called `nexus_structure` which has template, in JSON format, for the NeXus file to be written by the Filewriter software. This template contains all the static metadata for the experiment and must be parsed by scippneutron to get information such as pixel positions.

Parsing the JSON structure is ticket #58. This ticket is for making use of the parsing with the messages from Kafka as part of the live data interface.

The JSON template also contains "stream" objects which describe where to find data which is published to Kafka by the data acquisition system during the experiment, for example detection event data and sample environment metadata. Scippneutron could use this information so that only the topic containing the run start messages is required to specified.

PR #60, opened 2021-04-22
Parse json nexus template

### LINKED ISSUE #58: Parse JSON template for NeXus
The run start message contains a field called nexus_structure which has template, in JSON format, for the NeXus file to be written by the Filewriter software. This template contains all the static metadata for the experiment and must be parsed by scippneutron to get information such as pixel positions.

We should avoid repeating parsing logic already implemented in load_nexus. A couple of ideas:

- use polymorphism for functions which navigate or extract information from the NeXus file or JSON string
- use the JSON string to create an in-memory NeXus and parse it using the existing loading code

I prefer the first idea unless it turns out to be impractical for some reason.

Implement parsing the JSON from file for this ticket and then make use of it to parse the Kafka message in #56

matthew-d-jones:
Closes #58

Adds `load_nexus_json()` to load a json representation of a NeXus file. This json is included in "run start" messages from the ESS streaming system and contains all the metadata known at the start of an experiment run, for example it may contain the shape of detectors, what the sample is, who the users are, etc. I have used a fixture to parameterise the existing `load_nexus()` unit tests so that they also build and test a json file with `load_nexus_json()`. There is a separate ticket for using the implemented json parsing in the data streaming interface: #56

The json representation differs from HDF5 NeXus files as it contains `stream` objects to describe where to find other data that are acquired during the experiment, for example detection events, sample environment measurements, chopper timestamps, etc. In the current implementation the `stream` objects are ignored, if one is used as a transformation then a warning is displayed and it is treated as a 0-distance translation (identity matrix) in position calculations. This is documented by a unit test in `python/tests/test_load_nexus_json.py`.


python/src/scippneutron/_loading_detector_data.py
```diff
@@ -57,28 +57,37 @@ def _convert_array_to_metres(array: np.ndarray, unit: str) -> np.ndarray:
...
         warn(f"Skipped loading pixel positions as no units found on "
-             f"x_pixel_offset dataset in {detector_group.name}")
+             f"x_pixel_offset dataset in {loading.get_name(detector_group)}")
         return None
 
     try:
-        x_positions = detector_group["x_pixel_offset"][...].flatten()
-        y_positions = detector_group["y_pixel_offset"][...].flatten()
-    except KeyError:
+        x_positions = loading.load_dataset_from_group_as_numpy_array(
```
SimonHeybrock, line 72: `loading.load_...` is a bit odd naming? Maybe `resource` or `nexus` instead of `loading`?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -57,28 +57,37 @@ def _convert_array_to_metres(array: np.ndarray, unit: str) -> np.ndarray:
...
+            detector_group, "y_pixel_offset").flatten()
+    except MissingDataset:
         return None
     try:
-        z_positions = detector_group["z_pixel_offset"][...].flatten()
-    except KeyError:
+        z_positions = loading.load_dataset_from_group_as_numpy_array(
+            detector_group, "z_pixel_offset").flatten()
+    except MissingDataset:
+        # z offsets are allowed to be missing, in which case use zeros
```
SimonHeybrock, line 82: This is standard Nexus? Is this for image/flat detectors?

python/src/scippneutron/_loading_json_nexus.py
```diff
@@ -0,0 +1,271 @@
...
+    try:
+        attributes = element["attributes"]
+        try:
+            return attributes[attribute_name]
+        except TypeError:
+            for attribute in attributes:
+                if attribute[_nexus_name] == attribute_name:
+                    return attribute[_nexus_values]
+    except KeyError:
+        pass
```
SimonHeybrock, line 52: Lots of exception handling here and below that just passes. Can you explain the reasoning behind this?

python/src/scippneutron/_loading_transformations.py
```diff
@@ -60,72 +67,110 @@ def get_full_transformation_matrix(group: h5py.Group,
...
+        matrix = np.eye(4, dtype=float)
+        transformations.append(matrix)
     else:
-        raise TransformationError(
-            f"Unknown transformation type "
-            f"'{attributes['transformation_type'].astype(str)}'"
-            f" at {transform.name}")
-    return attributes['depends_on']
+        try:
+            vector = loading.get_attribute_as_numpy_array(
```
SimonHeybrock, line 143: With all these function that reutrn numpy array, are we paying a potentially big extra cost of then coping into a scipp variable? Or are all the cases where this happens for small arrays (small compared to event data)?

python/src/scippneutron/load_nexus.py
```diff
@@ -36,54 +39,60 @@ def _open_if_path(file_in: Union[str, h5py.File]):
         yield file_in
 
 
-def _add_string_attr_to_loaded_data(group: h5py.Group, dataset_name: str,
-                                    attr_name: str, data: sc.Variable):
+def _add_string_attr_to_loaded_data(group: Union[h5py.Group,
+                                                 Dict], dataset_name: str,
+                                    attr_name: str, data: sc.Variable,
+                                    loading: Union[LoadFromHdf5,
+                                                   LoadFromJson]):
```
SimonHeybrock, line 46: If we had an (abstract) base class, would the type hinting be nicer?

python/src/scippneutron/load_nexus.py
```diff
@@ -36,54 +39,60 @@ def _open_if_path(file_in: Union[str, h5py.File]):
...
+def _load_title(entry_group: Group, data: sc.Variable,
+                loading: Union[LoadFromHdf5, LoadFromJson]):
+    _add_string_attr_to_loaded_data(entry_group.group, "title",
+                                    "experiment_title", data, loading)
 
 
-def load_nexus(data_file: Union[str, h5py.File], root: str = "/", quiet=True):
+def load_nexus(data_file: Union[str, h5py.File],
+               root: str = "/",
+               quiet=True) -> Optional[sc.Variable]:
```
SimonHeybrock, line 95: `sc.Variable` or `sc.DataArray`?

python/tests/nexus_helpers.py
```diff
@@ -97,109 +99,189 @@ class Link:
...
-    return added_transform.name
-
-
-class InMemoryNexusFileBuilder:
+@dataclass
+class Stream:
+    path: str
+    topic: str = "topic"
+    source: str = "source"
+    writer_module: str = "f142"
```
SimonHeybrock, line 114: What is this?

SimonHeybrock (commented):
Can't say too much about the details, but code looks clean and tests are added. Minor comments/questions:

python/src/scippneutron/load_nexus.py
matthew-d-jones: `sc.DataArray` or `sc.Dataset` depending on what data it finds. Maybe a union of those is preferable to specifying `sc.Variable`?

python/src/scippneutron/load_nexus.py
matthew-d-jones: Yes, although one can also just do something like `LoadFrom = Union[LoadFromHdf5, LoadFromJson]`, I will do that

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: yes and yes, mostly convenience for X-ray facilities who pretty much never need the z coord for pixels.

python/src/scippneutron/_loading_json_nexus.py
matthew-d-jones: Thanks for pointing this out, there is some inconsistency in whether the function returns `None` or raises a `MissingAttribute` exception. I will refactor to consistently do the latter.

python/src/scippneutron/_loading_transformations.py
matthew-d-jones: This is only used for small arrays involved in transformation calculations where I need to do some matrix multiplication and only the result ends up in a scipp variable. I will add doc comments to reduce the risk of them being misused.

python/tests/nexus_helpers.py
matthew-d-jones: The filewriter has a kind of plugin architecture of "writer modules" which are responsible for knowing how to deserialise a particular message type and how the data should be written to file. This field tells the filewriter which plugin to use for messages it finds on the specified topic with the specified source name. I'll add comments.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: I'll change it to `nexus`, my intention was to avoid it sounding like it owns the data but I agree it sounds odd.

python/src/scippneutron/load_nexus.py
SimonHeybrock: `sc.Variable` is not a base of either ofthe two, it is an independent class.

matthew-d-jones:
@SimonHeybrock Thanks, I believe I have addressed your comments so far.

python/src/scippneutron/_loading_json_nexus.py
matthew-d-jones: I've also comments where instances of `pass` remain

SimonHeybrock (approved):

PR #64, opened 2021-04-26
Load pixel positions without events

### LINKED ISSUE #61: Load detector positions even without event data
`load_nexus()` only attempts to load pixel positions for detectors that have `NXevent_data` groups. This should be changed for two main reasons:
- Detection data may be included as `NXdata` (histograms) rather than `NXevent_data` (loading `NXdata` is not yet implemented)
- In the case of loading from the JSON representation with the live-data functionality the event data are usually streamed separately to the geometry information.

matthew-d-jones:
Closes #61 

Added tests and changes so that detector ids and pixel positions are loaded even if there are no event data, or no valid event data, associated with them. This is important for data streaming where the "run start" message usually contains geometry information but no events.

### DIFF
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -333,7 +333,9 @@ def _create_empty_event_data(event_data: List[DetectorData]):
             detector_id_dtype = data.detector_ids.dtype
     if empty_events is None:
         if detector_id_dtype is None:
-            # Create empty data array with types matching streamed event data
+            # Create empty data array with types/unit matching streamed event
+            # data, this avoids need to convert to concatenate with event data
+            # arriving from stream
             empty_events = _create_events_data_array(np.int64, "ns", np.int32)
         else:
             # If detector_ids were loaded then match the type used for those
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -128,24 +128,24 @@ class DetectorData:
     pixel_positions: Optional[sc.Variable] = None
 
 
-def _create_events_data_array(
+def _create_empty_events_data_array(
         tof_dtype: Any = np.int64,
         tof_unit: Union[str, sc.Unit] = "ns",
         detector_id_dtype: Any = np.int32) -> sc.DataArray:
-    return sc.DataArray(data=sc.Variable(dims=[_event_dimension],
-                                         values=[],
-                                         variances=[],
-                                         dtype=np.float32),
+    return sc.DataArray(data=sc.empty(dims=[_event_dimension],
+                                      shape=[0],
+                                      variances=True,
+                                      dtype=np.float32),
                         coords={
                             _time_of_flight:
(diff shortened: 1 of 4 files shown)


python/src/scippneutron/_loading_detector_data.py
```diff
@@ -118,12 +123,59 @@ def _load_pixel_positions(detector_group: GroupObject, detector_ids_size: int,
...
+                            _time_of_flight:
+                            sc.Variable(dims=[_event_dimension],
+                                        values=[],
+                                        dtype=tof_dtype,
+                                        unit=tof_unit),
+                            _detector_dimension:
+                            sc.Variable(dims=[_event_dimension],
+                                        values=[],
+                                        dtype=detector_id_dtype)
+                        })
```
SimonHeybrock, line 148: Might be nicer to use, e.g., `sc.zeros` to give `shape` instead of empty values. Also (not having read how this us used below), does `data` miss variances?

python/src/scippneutron/_loading_detector_data.py
```diff
@@ -260,3 +309,73 @@ def get_detector_id(detector_data: DetectorData):
...
+            break
+        elif data.detector_ids is not None:
+            detector_id_dtype = data.detector_ids.dtype
+    if empty_events is None:
+        if detector_id_dtype is None:
+            # Create empty data array with types matching streamed event data
+            empty_events = _create_events_data_array(np.int64, "ns", np.int32)
+        else:
+            # If detector_ids were loaded then match the type used for those
+            empty_events = _create_events_data_array(np.int64, "ns",
```
SimonHeybrock, line 340: Is there a reason for the `ns` default? I don't think there is anything wrong with that, but Mantid uses `us` for TOF?

matthew-d-jones: nanoseconds are used in the Kafka messages, so this avoids needing to convert to concatenate with event data arriving from stream. I'll make that clearer in the comment.

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: > does data miss variances?

Yes, turns out that hadn't caused me a problem during concatenation because they are also missing from the loading implementation. I've now added them in both places.

> Might be nicer to use, e.g., sc.zeros to give shape instead of empty values.

I don't think that is what I want here. I use this function to generate an empty event data array so that I can simply concatenate data from all detectors, whether or not the detector contained loadable events.

SimonHeybrock: My point is, you could use `sc.zeros(..., shape=[0])`? But I don't mind this either.

SimonHeybrock (approved):

python/src/scippneutron/_loading_detector_data.py
matthew-d-jones: oh I see, feels slightly odd to me to use `zeros` to not create any zeros.

SimonHeybrock: `sc.empty` would also work.

PR #67, opened 2021-04-28
Parse run start message in live data

### LINKED ISSUE #56: Parse run start message in data_stream

matthew-d-jones:
Closes #56

#### Changes
If a run info topic is specified then:
- the last available run start message is retrieved by `data_stream`,
- can optionally to join stream at the start of the last run,
- other topics do not need to be specified, they are retrieved from the run start message.

### DIFF
--- a/python/src/scippneutron/_loading_detector_data.py
+++ b/python/src/scippneutron/_loading_detector_data.py
@@ -8,7 +8,6 @@ from typing import Optional, List, Any, Dict, Union
 import numpy as np
 from ._loading_common import (BadSource, MissingDataset, Group)
 import scipp as sc
-from datetime import datetime
 from warnings import warn
 from itertools import groupby
 from ._loading_transformations import get_full_transformation_matrix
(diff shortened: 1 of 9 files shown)


SimonHeybrock (approved):
I am a bit too far away from the details of Kafka so I cannot comment anything particular.

Code looks clear, and a nice number of explanations has been added, thanks! 👍

PR #71, opened 2021-04-29
Use released scipp in latest env

SimonHeybrock:


docs/environments/scippneutron-latest.yml
```diff
@@ -2,20 +2,20 @@ name: scippneutron
...
   - jupyter_nbextensions_configurator
-  - mantid-framework
   - matplotlib-base
   - nodejs
-  - pythreejs
-  - scippneutron
   - python-confluent-kafka
-  - ess-dmsc::ess-streaming-data-types
+  - pythreejs
+  - scipp/label/dev::mantid-framework
```
OwenArnold, line 19: FYI I we have the same version on main and dev label, but there is no harm in this.

OwenArnold (approved):

matthew-d-jones (approved):
Thanks, this also avoids the glibc version problem we see on CentOS 7

PR #72, opened 2021-05-04
Azure pipeline update

nvaytet:
- Using https://github.com/scipp/pipelines for the azure yml templates.
- To fetch the data files to build the docs, the convention is that if there is a file called `pre_docs.sh`, it is run before the docs are built.
- pin boost version to 1.75
- Use a secret file for the SSH key for deploying the docs to github pages. This file is called `gh_pages_deploy_key_public`
- Added a new parameter `gh_pages_subdirectory` which, if specified, will clone `scippneutron` and checkout the `gh_pages` branch for the docs.

See companion PR https://github.com/scipp/scipp/pull/1883

Here is the successful build with anaconda upload and github pages published: https://dev.azure.com/scipp/scippneutron/_build/results?buildId=5072&view=results


.azure-pipelines/main.yml
```diff
@@ -17,129 +26,26 @@ resources:
...
-            displayName: 'Publish Conda package artefacts'
-
-  - stage: 'post_build_checks'
-    displayName: 'Post Build Checks'
-
-    jobs:
-      - template: templates/documentation_build.yml
-      - template: templates/test_can_install_and_import.yml
+trigger:
+  - main
```
OwenArnold, line 30: Could we also have a pipeline `next` for scippneutron against `scipp/label/dev`? otherwise can do this as a separate PR.

nvaytet: I would say this would be a nice step to do after this PR so we can test how easy it is to extend this.

.azure-pipelines/release.yml
```diff
@@ -1,135 +1,44 @@
...
+  - repository: SharedPipelines
+    name: scipp/pipelines
+    type: github
+    endpoint: scipp
+    ref: main
+
 trigger:
   tags:
     include:
       - '*.*.*'
```
OwenArnold, line 22: Minor point, should we enable `auto_cancel` for this?

conda/bld.bat
```diff
@@ -1,19 +1,8 @@
...
-::  Build, install and move scipp Python library to site packages location
-cmake --build . --target install --config Release || echo ERROR && exit /b
-move %CONDA_PREFIX%\scippneutron %CONDA_PREFIX%\lib\
+if "%INSTALL_PREFIX%" == "" set INSTALL_PREFIX=%cd%\scipp_install & call tools\make_and_install.bat
+
+move %INSTALL_PREFIX%\scippneutron %CONDA_PREFIX%\lib\
+move %INSTALL_PREFIX%\bin\scippneutron*.dll %CONDA_PREFIX%\bin\
+move %INSTALL_PREFIX%\Lib\scippneutron*.lib %CONDA_PREFIX%\Lib\
+move %INSTALL_PREFIX%\Lib\cmake\scippneutron %CONDA_PREFIX%\Lib\cmake\
+move %INSTALL_PREFIX%\include\scippneutron* %CONDA_PREFIX%\include\
```
OwenArnold, line 7:
```suggestion
move %INSTALL_PREFIX%\include\scippneutron %CONDA_PREFIX%\include\
```

OwenArnold:
```suggestion
move %INSTALL_PREFIX%\include\scippneutron-* %CONDA_PREFIX%\include\
```

OwenArnold: See recent fix made for scipp https://github.com/scipp/scipp/blob/main/conda/bld.bat

conda/meta.yaml
```diff
@@ -32,27 +32,25 @@ test:
...
+    - INSTALL_PREFIX
 
 about:
   home: https://github.com/scipp/scippneutron
   license: BSD-3-Clause
   summary: Neutron scattering tools for Data Reduction
   description: Neutron scattering tools for Data Reduction
   dev_url: https://github.com/scipp/scippneutron
-  doc_url:
+  doc_url: https://scipp.github.io/scippneutron
```
OwenArnold, line 56: 👍

tools/make_and_install.bat
```diff
@@ -0,0 +1,18 @@
...
+
+:: Benchmarks
+cmake --build . --target all-benchmarks --config Release -- /m:%NUMBER_OF_PROCESSORS% || echo ERROR && exit /b
+
+:: C++ tests
+cmake --build . --target all-tests --config Release -- /m:%NUMBER_OF_PROCESSORS% || echo ERROR && exit /b
+bin\Release\scippneutron-test.exe || echo ERROR && exit /b
+
+:: Build and install Python library
+cmake --build . --target install --config Release -- /m:%NUMBER_OF_PROCESSORS% || echo ERROR && exit /b
```
OwenArnold, line 18: The new `make_and_install.bat` approach (though still wrapped in conda-build) is not documented for scippneutron to my knowledge.

OwenArnold (changes requested):

.azure-pipelines/release.yml
nvaytet: I'm not sure here. We don't have `autoCancel` for either `main` or `release` on both `scipp` and `scippneutron`.
I don't mind either way

OwenArnold: Its maybe too minor to worry about. I think in general you do want `main` to run to completion, but if you realise you made a mistake on release, and you push an updated tag, you probably want to cancel the running one. I think this is too niche a situation to worry about - leave it as is.

conda/build.sh
```diff
@@ -2,12 +2,25 @@
...
 
+# conda-build breaks all of the key mechanisms used by ccache,
+# preventing a cache hit from ever happening.
+# To allow the use of ccache in CI builds, we call an external build script
+# (called make_and_install.sh) first to build the C++ library,
+# run the C++ tests, and build the python library.
+# In this case, the INSTALL_PREFIX will be defined.
+# If it is not defined (e.g. for a local conda-build), we need to call that
+# same script here.
+# Using this separate script also minimizes code duplication.
```
OwenArnold, line 13: 👍

OwenArnold (approved):

nvaytet:
do not merge yet, I still need to make a test run with py3.8

nvaytet:
Ok, there are too many conda conflicts to create a py3.8 scippneutron, so for now, we will only release 3.7.
We can come back to this later if we want to add the 3.8.

PR #73, opened 2021-05-04
scippneutron with scipp>0.6.1 (shallow copy changes)

OwenArnold:
pipeline [next](https://dev.azure.com/scipp/scippneutron/_build?definitionId=56&_a=summary). PR support for this branch has been disabled in CI.


python/src/scippneutron/mantid.py
```diff
@@ -466,17 +466,23 @@ def _convert_MatrixWorkspace_info(ws,
...
     spec_dim, spec_coord = init_spec_axis(ws)
-
-    if common_bins:
-        coord = sc.Variable([dim], values=ws.readX(0), unit=unit)
-    else:
-        coord = sc.Variable([spec_dim, dim], values=ws.extractX(), unit=unit)
+    # if spectrum axis for data is not spectrum,
+    # we cannot attach corresponding geometry information
+    # such as positions directly
+    data_and_geom_match = spec_dim == 'spectrum'
```
SimonHeybrock, line 472: How does this occur? Aren't Mantid workspaces always 2D? With geometry info related to the second axis?

SimonHeybrock (commented):
I know this is a draft, but had a short (partial) look anyway:

neutron/convert.cpp
```diff
@@ -8,17 +8,17 @@
...
 
-#include <scipp/dataset/bins.h>
 #include <scipp/dataset/dataset.h>
 #include <scipp/dataset/dataset_util.h>
 
 #include "scipp/neutron/constants.h"
 #include "scipp/neutron/conversions.h"
 #include "scipp/neutron/convert.h"
+#include <scipp/dataset/bins_view.h>
+#include <sstream>
```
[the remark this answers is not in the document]
OwenArnold: No `sstream` must be included for `std::ostringstream`

docs/tutorials/powder-diffraction.ipynb
```diff
@@ -200,8 +200,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dspacing_vanadium = scn.convert(vanadium, 'tof', 'dspacing', scatter=True, out=vanadium)\n",
-    "dspacing_sample = scn.convert(sample, 'tof', 'dspacing', scatter=True)\n",
+    "dspacing_vanadium = scn.convert(vanadium.copy(), 'tof', 'dspacing', scatter=True)\n",
```
OwenArnold, line 203: Note the copy here is a work-around for an existing bug rather than a desirable fix. What's happening here is that `convert` will acted upon the coords of the data (bins) of the input, so `sample.data.coords` becomes `dspacing`. This causes a failure further down in the script where a convert for `sample` from `tof` to `wavelength` is attempted, this has no effect and the coord remains in `dspacing` - I believe this could be a fundamental problem with the copy constructor behaviour for `DataArray` but will investigate separately.

python/src/scippneutron/mantid.py
SimonHeybrock: We should not be adding something with `spectrum` dimension in that case. If the spectrum-axis has dimension Q then so should the geometry information. Mantid has a 1:1 correspondence between axis and these meta data, so that should be preserved. What is the problem with having say the `position` coord depend on the spectrum-axis?

docs/tutorials/powder-diffraction.ipynb
SimonHeybrock: Maybe you can explain in a call. (shallow) copying a data arrays makes copies of the meta-data dicts so there should not be a problem here.

neutron/convert.cpp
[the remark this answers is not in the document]
SimonHeybrock: Then that should replace the `ostream` include above?

python/tests/test_neutron.py
```diff
@@ -42,13 +42,6 @@ def test_neutron_convert():
     assert dspacing.coords['dspacing'].unit == sc.units.angstrom
 
 
-def test_neutron_convert_out_arg():
-    d = make_dataset_with_beamline()
-    dspacing = scn.convert(d, 'tof', 'dspacing', scatter=True, out=d)
-    assert dspacing.coords['dspacing'].unit == sc.units.angstrom
-    assert dspacing is d
-
-
```
SimonHeybrock, line 51: Given that we are considering to add this again later, have you considered not removing it (which would break existing notebooks), but rather accept the out arg and simply assign to it?

neutron/beamline.cpp
```diff
@@ -49,36 +50,35 @@ void log_not_found(const Dim dim, const Dim a, const Dim b) {
...
+Variable sample_position(const dataset::Coords &meta) {
   return get_param(meta, NeutronDim::SamplePosition);
 }
 
-Variable Ltotal(const dataset::CoordsConstView &meta,
-                const ConvertMode scatter) {
+Variable Ltotal(const dataset::Coords &meta, const ConvertMode scatter) {
   // TODO Avoid copies here and below if scipp buffer ownership model is changed
   if (find_param(meta, NeutronDim::Ltotal)) {
     return copy(meta[NeutronDim::Ltotal]);
```
SimonHeybrock, line 84: Remove `copy` and `TODO`

neutron/beamline.cpp
```diff
@@ -120,57 +120,55 @@ Variable L2(const dataset::CoordsConstView &meta) {
...
   return position(meta) - sample_position(meta);
 }
 
 namespace {
-auto normalize(Variable &&var) {
-  const auto length = norm(var);
-  var /= length;
-  return std::move(var);
+auto normalized(const Variable &var) { // TODO rename
+  return var / norm(var);
```
SimonHeybrock, line 145: I think you renamed already, remove TODO?

neutron/convert.cpp
```diff
@@ -270,14 +276,9 @@ T convert_impl(T d, const Dim from, const Dim to, const ConvertMode scatter) {
...
-                         to, scatter);
-}
-
-DataArray convert(const DataArrayConstView &d, const Dim from, const Dim to,
-                  const ConvertMode scatter) {
-  check_params(from, to, scatter);
-  return convert(DataArray(d), from, to, scatter);
+  Dataset out = coords_to_attrs(convert_impl(Dataset(d), from, to, scatter),
+                                from, to, scatter);
+  return out[d.name()];
```
SimonHeybrock, line 281: Why do we need to do this via dataset now? It also has the effect of marking the attributes readonly, which should not be required here?

python/src/scippneutron/mantid.py
```diff
@@ -463,20 +464,26 @@ def _convert_MatrixWorkspace_info(ws,
...
+    pos, rot, shp = get_detector_properties(
+        ws,
+        source_pos,
+        sample_pos,
+        spec_dim,
+        advanced_geometry=advanced_geometry)
+
+    coords = {spec_dim: spec_coord}
+    # possible x - coord
+    if not ws.id() == 'MaskWorkspace':
```
SimonHeybrock, line 477: Can you elaborate? Why this change? What is this doing?

SimonHeybrock (changes requested):
Had a look at the more recent changes, I think this is almost done? See comments below.

There are also a couple outstanding comments above I believe.

python/src/scippneutron/mantid.py
OwenArnold: This information is fictional for a mask workspace. Attempts to add an empty dimension variable (length 1) with value (1). The fact that the mask workspace carries this at all more speaks for the fixed nature of the MatrixWorkspace.

neutron/convert.cpp
```diff
@@ -62,7 +62,8 @@ T convert_generic(T &&d, const Dim from, const Dim to, Op op,
   }
 
   // 3. Rename dims
-  d.rename(from, to);
+  if (d.coords().contains(from))
+    d.rename(from, to);
```
SimonHeybrock, line 66: Hmmm, not quite? `rename` is also renaming the dimension, which we might not want to skip, even if there is no such coord?

neutron/convert.cpp
```diff
@@ -109,7 +111,9 @@ auto scatter_params(const Dim dim) {
                                  NeutronDim::FinalEnergy,
                                  NeutronDim::Ltotal,
                                  NeutronDim::L1,
-                                 NeutronDim::L2};
+                                 NeutronDim::L2,
+                                 NeutronDim::SamplePosition,
+                                 NeutronDim::SourcePosition};
```
SimonHeybrock, line 116: Should the positions (and actually also the `L*` coords) be included in the subspan below?

neutron/test/convert_test.cpp
```diff
@@ -429,9 +450,10 @@ TEST_P(ConvertTest, Tof_to_Energy_Elastic) {
   EXPECT_TRUE(equals(data.values<double>(), {1, 2, 3, 4, 5, 6}));
   EXPECT_EQ(data.unit(), units::counts);
 
-  for (const auto &dim : {NeutronDim::Position, NeutronDim::SourcePosition,
-                          NeutronDim::SamplePosition})
-    ASSERT_EQ(energy.coords()[dim], tof.coords()[dim]);
+  ASSERT_FALSE(energy.coords().contains(NeutronDim::SourcePosition));
+  ASSERT_FALSE(energy.coords().contains(NeutronDim::SamplePosition));
```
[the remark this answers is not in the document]
SimonHeybrock: I suppose this would be fixed by my comment above on the `pos_invariant` subspan.

neutron/convert.cpp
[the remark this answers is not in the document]
SimonHeybrock: Do we need `sstream` AND `ostream`?

.azure-pipelines/main.yml
```diff
@@ -11,6 +11,7 @@ parameters:
 # found on the Azure pipelines web interface under
 # scipp > scippneutron > Settings > Service connections
 
+# New releases of scipp will automatically cause this pipeline to run
```
SimonHeybrock, line 14: Is the idea here that this would publish an updated package, provided that there are no breaking changes?

What if it fails? Would conda install the wrong (new) upstream scipp but the old scippneutron, leading to failures when running? Would this not that imply that we should therefore require a specific version (maximum) version in scippneutron (maybe allowing different patch releases)?

.azure-pipelines/next.yml
```diff
@@ -0,0 +1,51 @@
...
+  default: false
+- name: scipp_channel
+  type: string
+  default: scipp/label/dev
+
+# Note: the 'endpoint' refers to the Azure service connection, which can be
+# found on the Azure pipelines web interface under
+# scipp > scippneutron > Settings > Service connections
+
+# New releases of scipp will automatically cause this pipeline to run
```
SimonHeybrock, line 14: Can you explain? I thought the idea of `next` was to use the `Main` builds of scipp?

.azure-pipelines/pull_request.yml
```diff
@@ -26,12 +26,14 @@ pr:
   branches:
     include:
       - '*'
+    exclude:
+      - next # Next has dedicated pipeline and is by nature incompatible with the stable scipp
```
SimonHeybrock, line 30: I don't think this true? We *do* want to have CI just as usual when merging `next` into `main`. Where we need the `next` pipelines is for PRs *into* `next`.

.azure-pipelines/main.yml
OwenArnold: At present this will only run when the `scipp` `Release pipleline` fully completes. That requires a tagged version of scipp.

A new `scippneutron` would only be made available if the pipeline fully completes. We probably need to do this manually or otherwise because new releases of `scipp/label/main` have the potential to break `scipp/label/dev::scippneutron`.

Another potentially better way of handling our requirements for a safe release procedure would be to somehow put scippneutron as a stage in the scipp release pipeline. The same would be true of scippneutron->ess.

.azure-pipelines/next.yml
OwenArnold: Well `scipp/label/dev::scippneutron` uses `scipp/label/main::scipp`. This is specifically for working against the next planned release of scipp, i.e. using `scipp/label/main::scipp`

.azure-pipelines/pull_request.yml
OwenArnold: Yes indeed. The changes here are just to prevent wasteful (almost guranteed breaking) runs of next scippneutron against `scipp/label/main::scipp`.

Basically we need another pipeline file to do as you say.

.azure-pipelines/next.yml
SimonHeybrock: So be "releases" you don't mean releases but dev releases?

neutron/convert.cpp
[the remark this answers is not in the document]
OwenArnold: No, I will change.

neutron/convert.cpp
OwenArnold: We should talk about this on a separate call. There are some things around the testing that you've highlighted that I need to discuss/resolve too.

neutron/convert.cpp
OwenArnold: Yes, if i recall the problem is this https://github.com/scipp/scippneutron/blob/next/neutron/convert.cpp#L37, so we need to find a better way to work around this.

.azure-pipelines/pull_request.yml
SimonHeybrock: Didn't you add that file above?

I really don't think this should be disabled. If we merge `next` into `main` we want PR builds to run. What should be done instead is to *not* work directly on `next`. Treat `next` as if it was `main`, i.e., create PRs into `next`. PRs into `next` should use the `Next` pipeline.

If we just don't open the PR `next -> main` too early we don't run into the problems you are seeing now, I think.

.azure-pipelines/next.yml
OwenArnold: Yes, will fix working.

neutron/convert.cpp
SimonHeybrock: Define `iter()`?

.azure-pipelines/pull_request.yml
OwenArnold: I think what we want to say is that for this particular pipeline, we only want to consider prs where `main` in the base/target branch. In this case, yes we should drop the `exclude` of `next`. I think we can do this in azure, I'll have to remind myself how.

We would then have a similar pipeline with some different channel settings for prs where `next` is the base/target branch

SimonHeybrock: Can you just conditionally select the conda channel based on the base branch, but use the same pipeline?

OwenArnold: > Can you just conditionally select the conda channel based on the base branch, but use the same pipeline?

Yes that should be doable.

nvaytet:
@OwenArnold you seem to have committed the outputs in the `exploring_data` notebook...

docs/tutorials/working-with-masks.ipynb
```diff
@@ -128,7 +128,7 @@
     "## Exercise 2: Masking spatially\n",
     "\n",
     "By masking an `x` range, mask the end of the tubes.\n",
-    "- Define `x = sc.geometry.x(data.coords['position'])` to extract only the x-component of the position coordinate.\n",
+    "- Define `x = data.coords['position'].x1` to extract only the x-component of the position coordinate.\n",
```
nvaytet, line 131: This will now change again with the `fields.x` properties

neutron/convert.cpp
```diff
@@ -89,26 +116,31 @@ static T convert_with_factor(T &&d, const Dim from, const Dim to,
...
+      NeutronDim::L2,
+      NeutronDim::Ltotal,
+      NeutronDim::IncidentEnergy,
+      NeutronDim::FinalEnergy,
+  };
   if (dim == NeutronDim::Tof)
     return scipp::span<Dim>{};
   return pos_invariant.count(dim) ? scipp::span(params)
-                                  : scipp::span(params).subspan(3);
+                                  : scipp::span(params).subspan(8);
```
nvaytet, line 143: Can you explain why this change was needed?

scippneutron-developer-no-mantid.yml
```diff
@@ -29,6 +29,7 @@ dependencies:
   - pythreejs
   - pyyaml
   - tbb
+  - traitlets=4.3.3 # see https://github.com/jupyter-widgets/pythreejs/issues/334
```
nvaytet, line 32: I believe this pinning and the one below on `sphinx` are no longer needed.

scippneutron-developer.yml
```diff
@@ -43,8 +43,9 @@ dependencies:
   - flake8
 
   # Docs
+  - ipython=7.2.0
```
nvaytet, line 46: i don't think this is needed anymore

neutron/convert.cpp
```diff
@@ -33,31 +51,40 @@ T convert_generic(T &&d, const Dim from, const Dim to, Op op,
...
-      continue;
-    const auto &[indices, dim, buffer] =
-        item.data().template constituents<bucket<DataArray>>();
-    if (!buffer.coords().contains(from))
-      continue;
-    auto buffer_coord = buffer.coords().extract(from);
-    auto coord = make_non_owning_bins(indices, dim, VariableView(buffer_coord));
-    transform_in_place(coord, args..., op_);
-    buffer.coords().set(to, std::move(buffer_coord));
+  auto convert_item = [&](auto &item) {
```
SimonHeybrock, line 62: Why is the loop body extracted as a lambda now?

neutron/convert.cpp
```diff
@@ -33,31 +51,40 @@ T convert_generic(T &&d, const Dim from, const Dim to, Op op,
...
+  };
+  for (auto &&item : iter(d)) {
+    convert_item(item);
   }
 
   // 3. Rename dims
-  d.rename(from, to);
+  const auto &keys = d.dims().labels();
+  auto it = std::find(keys.begin(), keys.end(), from);
+  if (it != keys.end())
```
SimonHeybrock, line 86: Why not `if(d.dims().contains(from))`?

neutron/convert.cpp
SimonHeybrock: This can best be understood looking at the graph in https://scipp.github.io/scippneutron/user-guide/unit-conversions.html#Beamline-geometry-parameters-used-in-unit-conversions. We have some "intermediate" conversions where some position information is still relevant (`wavelength`, for example), and other such as `dspacing` that are position-invariant.

neutron/convert.cpp
```diff
@@ -33,31 +51,40 @@ T convert_generic(T &&d, const Dim from, const Dim to, Op op,
...
+    item.setData(dataset::make_bins(indices, dim, buffer));
+    auto view = dataset::bins_view<DataArray>(item.data());
+    auto transformed = copy(view.coords()[from]);
+    transform_in_place(transformed, args..., op_, "scippneutron.convert");
+    view.coords().set(to, transformed);
+    view.coords().erase(from);
+  };
+  for (auto &&item : iter(d)) {
+    convert_item(item);
   }
```
SimonHeybrock, line 81: Probably shorter and more obvious than adding the `IteratorFacade`:
```cpp
if constexpr (std::is_same_v<T, DataArray>)
  convert_item(d);
else
  for (auto &&item : d)
    convert_item(item);
```

neutron/convert.cpp
OwenArnold: Yes, contains better. Forgot we had that.

neutron/convert.cpp
OwenArnold: There are a number of usages of `iter` within `convert.cpp`. I did originally have something similar to what you have above, but quickly ended up looking messy/duplicated.

SimonHeybrock:
@OwenArnold @nvaytet Have a look at my commits, I think it addresses Neil's as well as my comments. Also deduplicated the convert details between dense coords and event coords.

conda/meta.yaml
```diff
@@ -19,6 +19,7 @@ requirements:
     - python {{ python }}
     - scipp {{ scipp }}
     - h5py
+    - boost {{ boost }}
```
SimonHeybrock, line 22: @OwenArnold Why this?

docs/tutorials/exploring-data.ipynb
```diff
@@ -39,7 +39,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "data = scn.load(filename='LARMOR00049338')"
+    "data = scn.load(filename='/Users/spu92482/Downloads/LARMOR00049338.nxs')"
```
SimonHeybrock, line 42: @OwenArnold Was this committed by accident?

neutron/convert.cpp
```diff
@@ -145,7 +146,7 @@ T attrs_to_coords(T &&x, const Dim to, const ConvertMode scatter) {
       return;
     Variable attr(range.begin()->attrs()[field]);
     if constexpr (std::is_same_v<std::decay_t<T>, Dataset>) {
-      for (const auto &item : range) {
+      for (auto item : range) {
         core::expect::equals(item.attrs()[field], attr);
         item.attrs().erase(field);
       }
```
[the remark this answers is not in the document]
SimonHeybrock: @OwenArnold It seems this comment was not addressed, in fact it looks like in the current version we are not setting attrs as readonly any more?

conda/meta.yaml
OwenArnold: I recall that we had other cases where the `mantid-framework` boost runtime version was being stamped on. But I don't 100% recall it might have been from debugging. Please try to remove.

nvaytet (approved):
Apart from the current parsing issues in the CI, looks good to me

PR #77, opened 2021-05-17
Switch to ModuleNotFoundError for mantid related tests

OwenArnold:
We want CI to fail if there are mantid-framework problems resulting from environment/package clashes rather than silently skipping tests. A recent example would be the installation/upgrade of runtime boost 1.76 compatible with scipp/scipp-neutron, which caused an ImportError for mantid. Now we would see such errors directly as we only identify that the module is missing.

### DIFF
--- a/python/tests/test_mantid.py
+++ b/python/tests/test_mantid.py
@@ -10,14 +10,7 @@ import os
 import scipp as sc
 import scippneutron as scn
 from .mantid_data_helper import MantidDataHelper
-
-
-def mantid_is_available():
-    try:
-        import mantid  # noqa: F401
-        return True
-    except ImportError:
-        return False
+from .mantid_data_helper import mantid_is_available
 
 
 def memory_is_at_least_gb(required):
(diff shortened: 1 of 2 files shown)


python/tests/test_mantid.py
```diff
@@ -10,14 +10,7 @@
 import scipp as sc
 import scippneutron as scn
 from .mantid_data_helper import MantidDataHelper
```
nvaytet, line 12:
```suggestion
from .mantid_data_helper import MantidDataHelper, mantid_is_available
```

python/tests/test_mantid.py
```diff
@@ -10,14 +10,7 @@
...
-
-
-def mantid_is_available():
-    try:
-        import mantid  # noqa: F401
-        return True
-    except ImportError:
-        return False
+from .mantid_data_helper import mantid_is_available
```
nvaytet, line 13:
```suggestion
```

nvaytet (changes requested):
Would this really have prevented the issue with the boost version?

OwenArnold:
> Would this really have prevented the issue with the boost version?
No, but we would have seen and understood it earlier. The test suite was still passing. It was only the docbuilds that flagged up the real problem.

nvaytet (approved):

PR #79, opened 2021-05-18
Re-enable upstream trigger

OwenArnold:


SimonHeybrock:
Can you elaborate? This would publish a new scippneutron package if scipp is updated (and would likely fail to build?)

OwenArnold:
> Can you elaborate? This would publish a new scippneutron package if scipp is updated (and would likely fail to build?)
Indeed this is wrong. We have some better handling for this in `next`

PR #80, opened 2021-05-19
Test the dmsc web file server

nvaytet:
Testing getting the neutron data files from a web server at dmsc instead of the Mantid ftp server.


matthew-d-jones (commented):
There used to be something in the dev docs about using `fetch_neutron_data.py` to get the data which are required for building the docs. Should there be a note on how to get the data from the new server, or is pre_docs.sh also run when you build the docs target locally?

tools/pre_docs.sh
```diff
@@ -2,6 +2,31 @@
...
-echo -e "usagereports.enabled=0\ndatasearch.directories=$(pwd)/data" > $HOME/.mantid/Mantid.user.properties
+if test -z "${DOCS_PREFIX}"
+then
+  DOCS_PREFIX="$(pwd)"
+fi
+
+# Fetch data files from local dmsc web server.
+# Note that the final '/' is important, if not wget ends up downloading all
+# the other folders on the same level.
+wget -c -N -nv -r -np -nH --cut-dirs=2 -R index.html* https://public.esss.dk/groups/scipp/scippneutron/
```
OwenArnold, line 13: osx does not come with `wget` so will need to add via conda.

nvaytet: ah but it makes it annoying for someone who wants to build docs locally on osx...

nvaytet: maybe we should just try to use python scripts for everything?

docs/make_docs.py
```diff
@@ -0,0 +1,73 @@
...
+import argparse
+import urllib
+import requests
+import re
+from pathlib import Path
+import subprocess
+import sys
+
+parser = argparse.ArgumentParser(description='Build doc pages with sphinx')
+parser.add_argument('--prefix', default='build')
```
OwenArnold, line 15: defaults not specified so `None` I think?

docs/make_docs.py
```diff
@@ -0,0 +1,73 @@
...
+parser.add_argument('--work_dir', default='.doctrees')
+parser.add_argument('--data_dir', default='data')
+
+
+def download_file(source, target):
+    os.write(1, "Downloading: {}\n".format(source).encode())
```
OwenArnold, line 21: Minor: formatted string literals instead in a few places?

tools/build_conda.py
```diff
@@ -0,0 +1,68 @@
...
+# @author Neil Vaytet
+
+import os
+import shutil
+import glob
+import sys
+import build_cpp
+
+
+class FileMover():
```
OwenArnold, line 12: I'm guessing there's no good way of sharing this type with what you added to pipelines repo?

tools/build_cpp.py
```diff
@@ -0,0 +1,113 @@
...
+    platform = sys.platform
+
+    # Set up absolute directory paths
+    source_dir = os.path.abspath(source_dir)
+    prefix = os.path.abspath(prefix)
+    build_dir = os.path.abspath(build_dir)
+
+    # Default options
+    shell = False
+    parallel_flag = '-j{}'.format(multiprocessing.cpu_count())
```
OwenArnold, line 41: Unless you are using `make` under the hood, I think think you might not need this? At least while using ninja, parallel builds with available cores is the default. But maybe we also need this for MSBuild?

OwenArnold (approved):

tools/build_cpp.py
nvaytet: I had asked you about this and you said you still needed the `-j` flag with Ninja.
I'm happy to remove, less code.

tools/build_conda.py
nvaytet: yes, i can't think of a good way, see https://github.com/scipp/scipp/pull/1945#issuecomment-856431120
I think the conclusion is we leave the duplication for now and remove it in a later step

docs/make_docs.py
nvaytet: Are you saying the default build directory for the docs should be `None`?
I don't think sphinx will work if no target directory is specified?

tools/build_cpp.py
nvaytet: i think i will leave it for now

PR #84, opened 2021-05-27
Pin scipp version

SimonHeybrock:
The plan is to make a patch-release with this, to avoid installing new releases of the scipp package before scippneutron packages have been updated.

### DIFF
--- a/conda/conda_build_config.yaml
+++ b/conda/conda_build_config.yaml
@@ -3,3 +3,6 @@ python:
 
 boost:
   - 1.75
+
+scipp:
+  - 0.6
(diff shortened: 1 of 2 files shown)


conda/conda_build_config.yaml
```diff
@@ -3,3 +3,6 @@ python:
 
 boost:
   - 1.75
+
+scipp:
+  - 0.6
```
OwenArnold, line 8: I don't think this is enough. Also need to provide variable in meta.yaml. For example, see how boost version is used. https://github.com/scipp/scippneutron/blob/pin-scipp-version/conda/meta.yaml#L10

OwenArnold (changes requested):

OwenArnold (approved):

PR #89, opened 2021-06-01
Fix spurious warnings for data stream

### LINKED ISSUE #85: Do not warn of missing dataset in group if group has a stream as a child
When parsing JSON nexus_structure in run start messages of `data_stream`, warnings like the following are output in the case that a group has an NX_class attribute but is missing datasets because the data comes from stream.
```
scippneutron/_loading_log_data.py:21: UserWarning: Skipped loading raw_data_1/framelog/events_log due to:
NXlog 'events_log' has no value dataset
  warn(f"Skipped loading {group.path} due to:\n{e}")
```
If the group has a "stream" object as a child then this warning should not be output.

matthew-d-jones:
Closes #85

**Do not review or merge until #86 is merged.** (branches from feature branch)

Avoids warning the user of missing datasets if the group instead contains a stream object to provide the missing data.

### DIFF
--- a/python/src/scippneutron/file_loading/_common.py
+++ b/python/src/scippneutron/file_loading/_common.py
@@ -8,6 +8,18 @@ import h5py
 
 
 class BadSource(Exception):
+    """
+    Raise if something is wrong with data source which
+    prevents it being used. Warn the user.
+    """
+    pass
+
+
+class SkipSource(Exception):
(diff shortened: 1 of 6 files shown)


SimonHeybrock (approved):
Looks ok. Changed base into feature branch so I can merge.

PR #96, opened 2021-06-07
Fix advanced geometry issues

OwenArnold:
Problem reported by Celine on slack

Fix two problems.

1. x,y,z buffers set for deep copies, but the behaviour changed as part of 0.7 scipp resulting in shallow copies instead. As a result x = y = z, giving instrument detector positions as colinear points along a line defined by x values
2. Incorrect calculation using signed delta for phi averaging results in y inversion. Originally introduced here https://github.com/scipp/scipp/pull/1339


SimonHeybrock (approved):
Code looks good, test was added. Thanks!

PR #97, opened 2021-06-09
Load and store essential sample related information

### LINKED ISSUE #95: Store more sample information natively
Currently the converters from Mantid simply store https://docs.mantidproject.org/nightly/api/python/mantid/api/Sample.html using `dtype=PyObject`. For unit conversions as mentioned in #94 (involving single crystals) we require at least the UB matrix stored in a scipp-native manner.

Ensure this works for both conversion from Mantid as well as native Nexus loading and Kafka stream.

OwenArnold:
Oriented lattice information is a requirement for many unit conversion operations, particularly those that are specific to single crystal diffraction. We need to be able to store that information in `scippneutron`.

We need features that.

1. can load this information from file
2. can convert the information from existing mantid workspaces

In either of the above the resulting data structures should be uniform.

Nexus defines a number of ways to define an oriented lattice. The most common is likely the Busing and Levi convention. 

* Loads potentially unique ub for every sample
* This does not address the existing problem of the `NXSample_components` discussed with @matthew-d-jones for which the loader is not compliant. That would be dealt with as separate refactor.

The changes here allow us to load UB (combined) and separate U matrix via nexus OR converted via Mantid. In either case the interface/naming is uniform for scipp.

Fixes #95.


matthew-d-jones (commented):
Looks good. My only query is if we really want to create the UB matrix as identity matrix when it is missing from file. Perhaps it would be less likely to mislead the user if the attribute is not created at all when it is missing from file, such that algorithms trying to use it would throw?

python/src/scippneutron/mantid.py
```diff
@@ -508,6 +520,12 @@ def _convert_MatrixWorkspace_info(ws,
         info["coords"]["incident_energy"] = _extract_einitial(ws)
     elif ws.getEMode() == DeltaEModeType.Indirect:
         info["coords"]["final_energy"] = _extract_efinal(ws, spec_dim)
+
+    if ws.sample().hasOrientedLattice():
+        info["attrs"].update({
+            "sample_ub": make_sample_ub(ws),
+            "sample_u": make_sample_u(ws)
```
SimonHeybrock, line 527: Since it is `UB` and not `ub`, i.e., a mathematical symbol, should this be uppercase in the attr names as well?

python/src/scippneutron/mantid.py
```diff
@@ -108,10 +108,22 @@ def make_variables_from_run_logs(ws):
...
+def make_sample_ub(ws):
+    # B matrix transforms the h,k,l triplet into a Cartesian system
+    # https://docs.mantidproject.org/nightly/concepts/Lattice.html
+    return sc.matrix(value=(ws.sample().getOrientedLattice().getUB()))
+
+
+def make_sample_u(ws):
+    # U matrix rotation for sample alignment
+    # https://docs.mantidproject.org/nightly/concepts/Lattice.html
+    return sc.matrix(value=(ws.sample().getOrientedLattice().getU()))
```
SimonHeybrock, line 124: Looking at https://github.com/mantidproject/documents/blob/master/Design/UBMatriximplementationnotes.pdf, `B` should have a unit, since it converts between an `hkl` vector and `Q`? And by extension, so should `UB`?

python/tests/test_load_nexus.py
```diff
@@ -1268,3 +1263,59 @@ def test_linked_datasets_are_found(load_function: Callable):
...
+
+
+def test_loads_sample_u_matrix(load_function: Callable):
+    builder = NexusBuilder()
+    builder.add_component(
+        Sample("sample",
+               orientation_matrix=np.ones(shape=[3, 3]),
+               orientation_matrix_units=""))
+    loaded_data = load_function(builder)
+    assert "sample_u_matrix" in loaded_data
```
SimonHeybrock, line 1289: Are these standard names in Nexus? `mantid.py` uses different names?

python/src/scippneutron/file_loading/_sample.py
```diff
@@ -0,0 +1,61 @@
...
+        if units == sc.units.dimensionless:
+            warn(f"'{name}' dataset in {nx_class} is missing units")
+        return matrix, units
+    else:
+        return None, None
+
+
+def _get_ub_of_component(group: GroupObject, nx_class: str,
+                         nexus: LoadFromNexus) -> Tuple[np.ndarray, sc.Unit]:
+    return _get_matrix_of_component(group, nx_class, nexus, "ub_matrix")
```
SimonHeybrock, line 27: If this is the name in Nexus, should we adopt the same, just like for other entries? If not, what is the rationale?

python/src/scippneutron/mantid.py
OwenArnold: Yes, sounds reasonable.

python/src/scippneutron/mantid.py
OwenArnold: Indeed. should be 1/A.

python/tests/test_load_nexus.py
OwenArnold: It is prefixed with the name of the sample. But there is a wider issue about the way that the loaders are reading the NXsample information - I believe the current way is wrong and that may affect the output naming. `u_matrix` is `orientation_matrix` in nexus, but that's different from the Busing Levi convention which uses `U`, `UB` and `B`, I would think that the latter makes more sense. so we have `U_matrix` and `UB_matrix`

python/src/scippneutron/file_loading/_sample.py
OwenArnold: see above

python/src/scippneutron/file_loading/_sample.py
```diff
@@ -0,0 +1,61 @@
...
+def _get_matrix_of_component(group: GroupObject, nx_class: str,
+                             nexus: LoadFromNexus,
+                             name: str) -> Tuple[np.ndarray, sc.Unit]:
+    matrix_found, _ = nexus.dataset_in_group(group, name)
+    matrix_unit_found, _ = nexus.dataset_in_group(group, f"{name}_units")
+    if matrix_found:
+        matrix = nexus.load_dataset_from_group_as_numpy_array(group, name)
+        units = nexus.get_unit(nexus.get_dataset_from_group(group, name))
+        if units == sc.units.dimensionless:
+            warn(f"'{name}' dataset in {nx_class} is missing units")
```
SimonHeybrock, line 19: Is this a good warning in this place? If the loaded matrix is a rotation matrix then the unit should indeed be dimensionless?

python/src/scippneutron/file_loading/_sample.py
```diff
@@ -0,0 +1,61 @@
...
+        return matrix, units
+    else:
+        return None, None
+
+
+def _get_ub_of_component(group: GroupObject, nx_class: str,
+                         nexus: LoadFromNexus) -> Tuple[np.ndarray, sc.Unit]:
```
SimonHeybrock, line 26: Do you know if a typical file actually stores the correct unit of UB?

OwenArnold: I've had a look through some of the ORNL ones and cant find any with the UB. I've done the same with some of the ISIS files, but for many techniques the UB is entered by the user at data reduction time, it does not come written into the file.

SimonHeybrock: Would it make sense to look for a unit, if not found warn and use `1/Angstrom`?

SimonHeybrock:
Builds failing?

SimonHeybrock (approved):

SimonHeybrock (approved):

SimonHeybrock:
Still a build issue here.

PR #98, opened 2021-06-10
Doctest

jl-wynen:


docs/build_docs.py
```diff
@@ -47,15 +48,26 @@ def download_multiple(remote_url, target_dir, extensions):
             download_file(os.path.join(remote_url, f), target)
 
 
+def get_abs_path(path, root):
```
nvaytet, line 51: Once again, it would be nice if we could find a way to no duplicate all these functions between `pipelines`, `scipp` and `scippneutron`

nvaytet (approved):

.azure-pipelines/pull_request.yml
```diff
@@ -14,7 +14,7 @@ resources:
     name: scipp/pipelines
     type: github
     endpoint: scipp
-    ref: main
+    ref: doctest
```
nvaytet, line 17: Remember to change this back to `main`

docs/build_docs.py
jl-wynen: The problem is that we need this file locally and not just in the CI.

nvaytet: I know, i was just mentioning this again, but I don't have a good idea on how we can solve it.

PR #103, opened 2021-06-17
Fix conversion of slices of binned data

### LINKED ISSUE #102: `convert` broken for slices of binned data
```python
scn.convert(new_data.copy(), 'tof', 'dspacing', scatter=True)
```
raises
```python
---------------------------------------------------------------------------
DimensionError                            Traceback (most recent call last)
<ipython-input-12-902261b1e9d2> in <module>
----> 1 scn.convert(sample['spectrum',0], 'tof', 'dspacing', scatter=True)

DimensionError: Cannot add coord exceeding DataArray dims
```

SimonHeybrock:
Fixes #102 by making a copy if input is a slice.

### DIFF
--- a/neutron/convert.cpp
+++ b/neutron/convert.cpp
@@ -66,7 +66,7 @@ T convert_generic(T &&d, const Dim from, const Dim to, Op op,
   for (auto &&item : iter(d)) {
     if (item.dtype() != dtype<core::bin<DataArray>>)
       continue;
-    if (dataset::bins_view<DataArray>(item.data()).coords().contains(from))
+    if (!dataset::bins_view<DataArray>(item.data()).coords().contains(from))
       continue;
     const auto data = item.data().is_slice() ? copy(item.data()) : item.data();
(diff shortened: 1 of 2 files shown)


neutron/convert.cpp
```diff
@@ -3,22 +3,22 @@
 /// @file
 /// @author Simon Heybrock
 #include <set>
+#include <sstream>
```
OwenArnold, line 6: What is this needed for?

OwenArnold (approved):

neutron/convert.cpp
SimonHeybrock: You added it. I just moved it from below.

SimonHeybrock: https://github.com/scipp/scippneutron/pull/73/files#r627401502

OwenArnold: Ah sorry I didn't see the deletion and it didn't correspond as an include I could see was needed for any of the new changes.

PR #105, opened 2021-06-23
Remove boost-cpp pin

OwenArnold:
Explore work around applied here #84 technically we should not need to pin our build-time only dependencies. If we do, need to establish what exactly is brining in a conflicting runtime boost version. Partial fix to https://github.com/scipp/scipp/issues/1888

Consider alongside https://github.com/scipp/pipelines/pull/6

Will need to do similar things for scipp itself where pinning can also be removed.


SimonHeybrock:
Closing, since we will get boost using conan.

PR #106, opened 2021-06-23
Avoid error when loading log in NeXus file with unrecognised units

matthew-d-jones:
Warn the user and use `dimensionless` if units in `NXlog` are not recognised.
Problem was discovered when trying to load ISIS files with units such as "is_vetoing", "frames" and "events". I notice that "mBar" is also not recognised, should we do something about that one?

### DIFF
--- a/python/src/scippneutron/file_loading/_log_data.py
+++ b/python/src/scippneutron/file_loading/_log_data.py
@@ -70,7 +70,7 @@ def _load_log_data_from_group(group: Group,
         unit = sc.Unit(unit)
     except sc.UnitError:
         warn(f"Unrecognised unit '{unit}' for value dataset "
-             f"in NXlog '{group.path}'")
+             f"in NXlog '{group.path}'; setting unit as 'dimensionless'")
         unit = sc.units.dimensionless
 
     try:
--- a/python/src/scippneutron/file_loading/_log_data.py
+++ b/python/src/scippneutron/file_loading/_log_data.py
@@ -69,7 +69,7 @@ def _load_log_data_from_group(group: Group,
     try:
         unit = sc.Unit(unit)
     except sc.UnitError:
-        warn(f"Unrecognised unit '{unit}' for value dataset "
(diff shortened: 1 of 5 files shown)


python/tests/test_load_nexus.py
```diff
@@ -1268,3 +1268,24 @@ def test_linked_datasets_are_found(load_function: Callable):
...
                           expected_detector_ids)
+
+
+def test_warning_but_no_error_for_unrecognised_log_unit(
+        load_function: Callable):
+    values = np.array([1.1, 2.2, 3.3])
+    times = np.array([4.4, 5.5, 6.6])
+    name = "test_log"
+    builder = NexusBuilder()
+    unknown_unit = "elephants"
```
OwenArnold, line 1279: :)

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -66,6 +66,12 @@ def _load_log_data_from_group(group: Group,
 
     unit = nexus.get_unit(
         nexus.get_dataset_from_group(group.group, value_dataset_name))
+    try:
+        unit = sc.Unit(unit)
+    except sc.UnitError:
+        warn(f"Unrecognised unit '{unit}' for value dataset "
```
OwenArnold, line 72:
```suggestion
        warn(f"Unrecognized unit '{unit}' for value dataset "
```

OwenArnold (changes requested):
Aside from one minor change, looks good.

OwenArnold (approved):

OwenArnold:
> "mBar" is also not recognised, should we do something about that one?
units library is place to fix this specific issue, thought the fix here is correct as there could be many unit types that are not understood. Is `mbar` accepted by the units library?

matthew-d-jones:
> > "mBar" is also not recognised, should we do something about that one?
> 
> units library is place to fix this specific issue, thought the fix here is correct as there could be many unit types that are not understood. Is `mbar` accepted by the units library?

`mbar` works fine, it is the capitalisation that causes the problem.

PR #110, opened 2021-07-01
Fix missing code block in data stream dev docs

matthew-d-jones:


nvaytet (approved):

nvaytet:
Could you at the same time add the msissing `import scipp as sc` in the code example at the end.
As well as maybe `create_task` call for people who don't know about asyncio?

nvaytet (changes requested):
Could you at the same time add the msissing `import scipp as sc` in the code example at the end.
As well as maybe create_task call for people who don't know about `asyncio`?

nvaytet (approved):

PR #115, opened 2021-07-09
Make load_nexus compatible with h5py-lookalikes

SimonHeybrock:
- Avoid checking for instances of `h5py` classes
- Explicitly return attribute as numpy array, since it may return a `list` otherwise.

### DIFF
--- a/python/src/scippneutron/file_loading/_hdf5_nexus.py
+++ b/python/src/scippneutron/file_loading/_hdf5_nexus.py
@@ -60,7 +60,7 @@ class LoadFromHdf5:
         }
 
         def _match_nx_class(_, h5_object):
-            if isinstance(h5_object, h5py.Group):
+            if LoadFromHdf5.is_group(h5_object):
                 try:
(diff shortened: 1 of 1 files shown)


python/src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -191,7 +191,7 @@ def get_object_by_path(group: Union[h5py.Group, h5py.File],
     def get_attribute_as_numpy_array(node: Union[h5py.Group, h5py.Dataset],
                                      attribute_name: str) -> np.ndarray:
         try:
-            return node.attrs[attribute_name]
+            return np.array(node.attrs[attribute_name])
```
danielballan, line 194: If this _may_ be a numpy array already, you may prefer `np.asarray` to avoid an unnecessary copy.

nvaytet (approved):

danielballan:
It is good to see this one go on in!

PR #118, opened 2021-07-20
Use conan for C++ deps

SimonHeybrock:
Changes nearly identical to what we did in scipp.


python/CMakeLists.txt
```diff
@@ -16,5 +16,8 @@ set_target_properties(
                            "${CMAKE_INSTALL_RPATH}/../${CMAKE_INSTALL_LIBDIR}"
 )
 
+pybind11_strip(_scippneutron)
```
OwenArnold, line 19: What is the purpose of these calls?

OwenArnold: I see what cmake_strip is doing and think I understand this now.

SimonHeybrock: See https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode. It is needed since we don't use `pybind11_add_module` anymore.

OwenArnold (changes requested):
I manage to get this to work only with cmake 3.12 locally. For reasons similar to those explained here https://github.com/scipp/scipp/pull/2063

I wanted to make a similar modification to the environments but there are dependency conflicts between cmake 3.12 and mantid-framework we need to understand first.

Other than that, I think these changes are fine and I can build scippneutron locally

CMakeLists.txt
```diff
@@ -54,25 +54,48 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
...
+conan_cmake_autodetect(conan_settings)
+conan_cmake_install(
+  PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR} SETTINGS ${conan_settings} BUILD
+  outdated
+)
+
+find_package(Benchmark)
+find_package(Boost 1.67 REQUIRED)
+find_package(Eigen3 REQUIRED)
+find_package(GTest REQUIRED)
```
OwenArnold, line 86:
```suggestion
find_package(GTest CONFIG REQUIRED)
```

OwenArnold (changes requested):

CMakeLists.txt
OwenArnold: See https://github.com/scipp/scipp/pull/2063 and https://gitlab.kitware.com/cmake/cmake/-/issues/17365

Issue #20, opened 2021-03-01
Adjust log times to consistent epoch in load_nexus


matthew-d-jones:
`NXlog`s can have an `offset` attribute, values in the `time` dataset are relative to this `offset`, and this is currently not handled in `load_nexus`.
Loaded logs must be adjusted to a consistent epoch, I suggest start of run (if available) or otherwise unix epoch. Alternatively times could be converted to datetime?

This is not immediately causing a problem because ESS files are written with all times relative to unix epoch.

PR #120, opened 2021-07-26
Load nexus time offsets

Tom-Willemsen:
Add corrections of `NXLog` timestamps:
- Convert all times to be relative to start of run
  * In the case where start of run = start of log timestamp this is a no-op, but this is not necessarily the case
- Implement corrections for logged times with a scaling factor, which needs to be applied to times before arriving at the units of time specified in `NXLog @units`
- Add test cases for `units != s` in the NXLog entry - they will get converted to s as part of loading.

Closes https://github.com/scipp/scippneutron/issues/20

### DIFF
--- a/python/src/scippneutron/file_loading/_log_data.py
+++ b/python/src/scippneutron/file_loading/_log_data.py
@@ -3,11 +3,12 @@
 # @author Matthew Jones
 
 import numpy as np
-from typing import Tuple, List
+from typing import Tuple, List, Union
 import scipp as sc
-from ._common import (BadSource, SkipSource, MissingDataset, Group)
+from ._common import (BadSource, SkipSource, MissingDataset, MissingAttribute, Group)
 from ._nexus import LoadFromNexus, ScippData
 from warnings import warn
+from dateutil.parser import parse as parse_date
 
 
 def load_logs(loaded_data: ScippData, log_groups: List[Group],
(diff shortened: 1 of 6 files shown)


python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -145,15 +153,30 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
...
+
+    if groups[nx_entry]:
+        _load_title(groups[nx_entry][0], loaded_data, nexus)
+        _load_start_and_end_time(groups[nx_entry][0], loaded_data, nexus)
+
+        try:
+            run_start_time = nexus.load_scalar_string(
+                groups[nx_entry][0].group, "start_time")
+        except (AttributeError, TypeError, MissingDataset):
+            run_start_time = None
```
OwenArnold, line 165: You have this `try-except-default` type pattern a lot. I wonder if having some generic helper such as:

```
def apply_or_default(function, default, *exceptions):
    try:
        return function()
    except exceptions:
        return default
```
You probably need to work in keyword args to pass function arguments in the above.

python/tests/nexus_helpers.py
```diff
@@ -354,6 +368,12 @@ def add_instrument(self, name: str):
     def add_title(self, title: str):
         self._title = title
 
+    def add_start_time(self, start_time: str):
```
OwenArnold, line 371: minor question: is `start_time` and `end_time` enough to determine via the API that you are referring to the data (NXentry)?

python/tests/test_load_nexus.py
```diff
@@ -1289,3 +1294,159 @@ def test_warning_but_no_error_for_unrecognised_log_unit(
...
+    builder.add_log(
+        Log(name="test_log",
+            value=np.zeros(shape=(len(times), )),
+            time=np.array(times),
+            start_time=log_start))
+
+    loaded_data = load_function(builder)
+
+    assert np.allclose(loaded_data["test_log"].values.coords['time'].values,
+                       (np.array(times) + start_time_delta))
```
OwenArnold, line 1336: I'm not clear on why you parameterise with `start_time_delta` other than to sanity check as you write these function? Or is there another reason i'm missing?

OwenArnold: using `np.allclose` but not checking the units on the `loaded_data`

python/tests/test_load_nexus.py
```diff
@@ -1289,3 +1294,159 @@ def test_warning_but_no_error_for_unrecognised_log_unit(
...
+
+    loaded_data = load_function(builder)
+
+    assert np.allclose(loaded_data["test_log"].values.coords['time'].values,
+                       ((np.array(times) * scaling_factor) + start_time_delta))
+
+
+@pytest.mark.parametrize(
+    "units", ("ps", "ns", "us", "ms", "s", "minute", "hour", "day", "year"))
+def test_adjust_log_times_with_different_time_units(units,
```
OwenArnold, line 1368: 👍

python/tests/test_load_nexus.py
```diff
@@ -1289,3 +1294,159 @@ def test_warning_but_no_error_for_unrecognised_log_unit(
...
+    builder.add_log(
+        Log(name="test_log_1",
+            value=np.zeros(shape=(1, )),
+            time=np.array([1]),
+            start_time="1970-01-01T00:00:00Z"))
+
+    with pytest.warns(UserWarning, match="Invalid run start date "):
+        loaded_data = load_function(builder)
+
+        assert "test_log_1" not in loaded_data
```
OwenArnold, line 1452: Good testing.

OwenArnold (changes requested):
Basically looks to me like this works properly. Some minor changes needed.

python/src/scippneutron/file_loading/load_nexus.py
Tom-Willemsen: I did a test implementation of this applying it to one instance of the pattern here https://github.com/scipp/scippneutron/commit/13c1c6d95b633762ec27bf3d13e7a4bd452004de , but I don't think this improves readability. If you disagree, happy to go ahead with this though.

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -3,25 +3,91 @@
...
+    except (ParserError, OverflowError):
+        raise BadSource(
+            f"Invalid run start date '{run_start}' when loading NXLog.")
+
+    _scale = sc.scalar(
+        value=scaling_factor if scaling_factor is not None else 1.,
+        unit=sc.units.dimensionless,
+        dtype=sc.dtype.float64)
+
+    return (raw_times_s * _scale) + (_log_start_ts - _run_start_ts)
```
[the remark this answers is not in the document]
Tom-Willemsen: If `run_start` is before `log_start`:
- `_run_start_ts < _log_start_ts`
- `_log_start_ts - _run_start_ts` is positive
- This will increase the raw times
- This seems to make sense; the raw times were relative to the log start time, and the converted times which are larger have been converted to be relative to the (earlier) run start time.

Or in other words, assuming no unit conversions/scaling factors, I think the following should hold:
```
times relative to start of run == times relative to start of log
```
which is
```
run_start_ts + (converted times) == log_start_ts + (raw times)
```
which rearranges to the maths above.

The test case `test_adjust_log_times_without_scaling_factor` does test both `log_start_ts < run_start_ts` and `log_start_ts > run_start_ts`

python/src/scippneutron/file_loading/load_nexus.py
OwenArnold: Fair enough, and +1 for trying it out. I think I agree once you start specifying tuples of possible exceptions it gets cumbersome.

OwenArnold (approved):
Looks good. Happy to merge, though would be good to know this works with a FileWriter produced nexus file. Maybe we can find one.

PR #121, opened 2021-07-29
Add clcache for windows

OwenArnold:
WIP, see upstream pipelines https://github.com/scipp/pipelines/pull/8


SimonHeybrock:
@OwenArnold Conflicts need resolving before we can merge.

tools/build_cpp.py
```diff
@@ -100,14 +106,20 @@ def main(prefix='install', build_dir='build', source_dir='.'):
     run_command(['cmake', '-B', '.', '-S', source_dir, '-LA'], shell=shell)
 
     # Compile benchmarks, C++ tests, and python library
+    start = time.time()
```
nvaytet, line 109: Do we want to keep the timings or was this for debugging?

OwenArnold: It was originally for debugging, but we should keep I think to see how we are getting on with ccache and compilation times.

scippneutron-developer-no-mantid.yml
```diff
@@ -17,6 +17,8 @@ dependencies:
   - ninja
   - tbb-devel
   - scipp
+  - cmake
+  - pip
```
nvaytet, line 21: Why do you need `pip` here?

OwenArnold: This is where we get ccache from.

nvaytet: And the cmake is because the system version is too old?

OwenArnold: Yes, we've also had some pain from this before I think (out of date system cmake), better we control in our environments and have cross-platform consistency.

nvaytet (approved):

PR #125, opened 2021-08-03
Speed up windows builds

Tom-Willemsen:
For https://github.com/scipp/scipp/issues/2078

Same fix as in https://github.com/scipp/scipp/pull/2080

Not expecting a large performance increase in this case as scippneutron doesn't build much C++ anyway - but useful for consistency with scipp.


nvaytet (approved):

nvaytet:
40 mins for the conda-build on windows... we could really do with `mamba`

PR #126, opened 2021-08-05
Fix iter compilation error against Scipp main

nvaytet:
The `iter` helper class was removed from Scipp, we copy it back in `convert.cpp` that uses it.

This is temporary since `convert.cpp` will change to using `transform_coords`, but it allows to compile scippneutron until these changes have been finalised.

### DIFF
--- a/neutron/convert.cpp
+++ b/neutron/convert.cpp
@@ -26,6 +26,14 @@ using namespace scipp::dataset;
 namespace scipp::neutron {
 
 namespace {
+
+template <class T> static decltype(auto) iter(T &d) {
+  if constexpr (std::is_same_v<T, Dataset>)
+    return d;
+  else
+    return d.iterable_view();
+}
+
 // Iterable facade around a single object
 template <class T> class IterableFacade {
 public:


jl-wynen (approved):

nvaytet:
Note that this won't pass CI until a new version of Scipp is released

neutron/convert.cpp
```diff
@@ -26,6 +26,14 @@ using namespace scipp::dataset;
 namespace scipp::neutron {
 
 namespace {
+
+template <class T> static decltype(auto) iter(T &d) {
+  if constexpr (std::is_same_v<T, Dataset>)
+    return d;
+  else
+    return d.iterable_view();
+}
```
OwenArnold, line 35: Wasn't the intention also to remove this https://github.com/scipp/scippneutron/blob/a213687a0e125dc139b334232c5212bf6594d946/neutron/convert.cpp#L37-L51 ?

OwenArnold (changes requested):

neutron/convert.cpp
jl-wynen: Yes. I am working on rewriting `convert` in terms of `transform_coords`. But until then, this PR makes scippneutron compilable against the latest scipp.

OwenArnold: But the IterableFacade is no longer used as far as I can tell

jl-wynen: Yes, so it can be removed.

nvaytet: it doesn't compile if i remove it:
```
/home/nvaytet/code/neutron/source/neutron/convert.cpp: In instantiation of ‘decltype(auto) scipp::neutron::{anonymous}::iter(T&) [with T = scipp::dataset::DataArray]’:
/home/nvaytet/code/neutron/source/neutron/convert.cpp:240:31:   required from ‘T scipp::neutron::convert_impl(T, scipp::units::Dim, scipp::units::Dim, scipp::neutron::ConvertMode) [with T = scipp::dataset::DataArray]’
/home/nvaytet/code/neutron/source/neutron/convert.cpp:306:59:   required from here
/home/nvaytet/code/neutron/source/neutron/convert.cpp:34:14: error: ‘class scipp::dataset::DataArray’ has no member named ‘iterable_view’
   34 |     return d.iterable_view();
      |            ~~^~~~~~~~~~~~~
/home/nvaytet/code/neutron/source/neutron/convert.cpp: In instantiation of ‘T scipp::neutron::convert_impl(T, scipp::units::Dim, scipp::units::Dim, scipp::neutron::ConvertMode) [with T = scipp::dataset::DataArray]’:
/home/nvaytet/code/neutron/source/neutron/convert.cpp:306:59:   required from here
/home/nvaytet/code/neutron/source/neutron/convert.cpp:240:3: error: forming reference to void
  240 |   for (const auto &item : iter(d))
      |   ^~~
```

nvaytet: But anyway, since we have just agreed not to release Scipp before all the stuff in convert is done, then this PR is obsolete

OwenArnold: SFINAE - but yes lets just do as you suggest Neil.

Issue #112, opened 2021-07-07
`load` adds a single TOF bin around event data, while `load_nexus` does not


nvaytet:
`load_nexus` also needs to add this single TOF bin.

SimonHeybrock:
Just as a note why this is useful and why `load` does this: This allows for creating 2d visualizations by default (intensity vs `[pixel, tof]`), whereas otherwise figures would just be 1d (intensity vs pixel)

PR #127, opened 2021-08-05
Add single ToF bin around event data

Tom-Willemsen:
Closes https://github.com/scipp/scippneutron/issues/112

### DIFF
--- a/python/src/scippneutron/file_loading/_detector_data.py
+++ b/python/src/scippneutron/file_loading/_detector_data.py
@@ -268,22 +268,50 @@ def load_detector_data(event_data_groups: List[Group], detector_groups: List[Gro
 
     pixel_positions_loaded = all(
         [data.pixel_positions is not None for data in event_data])
+
+    _min_tof = min(
+        sc.min(data.events.coords[_time_of_flight]).values for data in event_data)
+    _max_tof = max(
+        sc.max(data.events.coords[_time_of_flight]).values for data in event_data)
+
+    # This can happen if there were no events in the file at all as sc.min will return
+    # double_max and sc.max will return double_min
+    if _min_tof >= _max_tof:
+        _min_tof = float("-inf")
+        _max_tof = float("inf")
+
     detector_data = event_data.pop(0)
+
+    _tof_edges = sc.Variable(
+        values=[_min_tof, _max_tof + 1],  # +1 so that bins include all data
+        dims=[_time_of_flight],
+        unit=detector_data.events.coords[_time_of_flight].unit,
+        dtype=detector_data.events.coords[_time_of_flight].dtype,
+    )
 
     # Events in the NeXus file are effectively binned by pulse
     # (because they are recorded chronologically)
     # but for reduction it is more useful to bin by detector id
-    events = sc.bin(detector_data.events, groups=[detector_data.detector_ids])
+    events = sc.bin(detector_data.events,
+                    groups=[detector_data.detector_ids],
+                    edges=[_tof_edges])
+
     if pixel_positions_loaded:
         # TODO: the name 'position' should probably not be hard-coded but moved
         # to a variable that cah be changed in a single place.
         events.coords['position'] = detector_data.pixel_positions
     while event_data:
         detector_data = event_data.pop(0)
-        new_events = sc.bin(detector_data.events, groups=[detector_data.detector_ids])
+
+        new_events = sc.bin(detector_data.events,
+                            groups=[detector_data.detector_ids],
+                            edges=[_tof_edges])
+
         if pixel_positions_loaded:
             new_events.coords['position'] = detector_data.pixel_positions
+
         events = sc.concatenate(events, new_events, dim=_detector_dimension)
+
     return events
 
 
--- a/python/src/scippneutron/file_loading/_detector_data.py
+++ b/python/src/scippneutron/file_loading/_detector_data.py
@@ -16,8 +16,6 @@ _detector_dimension = "detector_id"
 _event_dimension = "event"
 _time_of_flight = "tof"
 
-TOF_EPSILON = 1
-
 
 class DetectorIdError(Exception):
     pass
@@ -283,8 +281,18 @@ def load_detector_data(event_data_groups: List[Group], detector_groups: List[Gro
 
     detector_data = event_data.pop(0)
 
+    tof_dtype = detector_data.events.coords[_time_of_flight].dtype
+
+    if str(tof_dtype).startswith("int"):
+        _max_tof += 1
+    else:
+        _max_tof = np.nextafter(_max_tof, float("inf"))
+
     _tof_edges = sc.array(
-        values=[_min_tof, _max_tof + TOF_EPSILON],  # so that bins include all data
+        values=[
+            _min_tof,
+            _max_tof,
+        ],
         dims=[_time_of_flight],
         unit=detector_data.events.coords[_time_of_flight].unit,
         dtype=detector_data.events.coords[_time_of_flight].dtype,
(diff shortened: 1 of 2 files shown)


python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -268,22 +268,50 @@ def get_detector_id(data: DetectorData):
...
+
+    # This can happen if there were no events in the file at all as sc.min will return
+    # double_max and sc.max will return double_min
+    if _min_tof >= _max_tof:
+        _min_tof = float("-inf")
+        _max_tof = float("inf")
+
     detector_data = event_data.pop(0)
 
+    _tof_edges = sc.Variable(
```
OwenArnold, line 285: minor: use `sc.array`

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -268,22 +268,50 @@ def get_detector_id(data: DetectorData):
 
     pixel_positions_loaded = all(
         [data.pixel_positions is not None for data in event_data])
+
+    _min_tof = min(
```
OwenArnold, line 272: I guess you do it like this because we could have multiple nxevent_data groups and concatenating before taking the sc.min would be wasteful?

python/tests/test_load_nexus.py
```diff
@@ -83,14 +83,14 @@ def test_loads_data_from_single_event_data_group(load_function: Callable):
     # May be reordered due to binning (hence np.sort)
     assert np.array_equal(
         np.sort(
```
OwenArnold, line 85: Is `sc.identical(sc.sort(...), sc.sort(...))` too slow?

python/src/scippneutron/file_loading/_detector_data.py
OwenArnold, line 285: minor: use `sc.array`

python/tests/test_load_nexus.py
```diff
@@ -1385,3 +1385,32 @@ def test_nexus_file_with_invalid_run_start_date_warns_and_skips_logs(
...
+        event_index=np.array([0, 3, 3, 5]),
+    )
+
+    builder = NexusBuilder()
+    builder.add_event_data(event_data)
+
+    loaded_data = load_function(builder)
+
+    # Size 2 for each of the two bin edges around a single bin
+    assert len(loaded_data.coords["tof"]) == 2
```
OwenArnold, line 1408: FYI, can use `shape` rather than `len`

OwenArnold (changes requested):

Tom-Willemsen:
> Did you benchmark the performance of this? I am wondering if we lose much time in case the user binds in tof right after loading anyway.

I just ran some benchmarks. There is a bit of fluctuation, but on this branch the average on my machine for loading `SANS2D_ESS_Example1.nxs` is approx 1.8 seconds. On `main`, it takes approx 1.7 seconds.

> Or is the time dominated by mantid anyway?

This is the "pure scipp" loader; it is not using mantid.

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: It hadn't occured to me to try concatenating before binning. Do you think there would be an advantage to that approach? Happy to try it if so.

SimonHeybrock: It is deliberate that the loader concatenates NXevent_data groups *after* binning. If we did it the other way round it would be significantly less efficient (Matt has tried and tested this during the original implementation, and confirmed the expected performance difference).

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -268,22 +268,50 @@ def get_detector_id(data: DetectorData):
...
+        sc.min(data.events.coords[_time_of_flight]).values for data in event_data)
+    _max_tof = max(
+        sc.max(data.events.coords[_time_of_flight]).values for data in event_data)
+
+    # This can happen if there were no events in the file at all as sc.min will return
+    # double_max and sc.max will return double_min
+    if _min_tof >= _max_tof:
+        _min_tof = float("-inf")
+        _max_tof = float("inf")
```
SimonHeybrock, line 281: Why do you need INF here instead of what you get from the function?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -268,22 +268,50 @@ def get_detector_id(data: DetectorData):
...
+
     detector_data = event_data.pop(0)
 
+    _tof_edges = sc.array(
+        values=[_min_tof, _max_tof + 1],  # +1 so that bins include all data
```
SimonHeybrock, line 286: Can you just add a float epsilon? 1 can be quite large, depending on the context.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -268,22 +268,50 @@ def get_detector_id(data: DetectorData):
...
+        dtype=detector_data.events.coords[_time_of_flight].dtype,
+    )
+
     # Events in the NeXus file are effectively binned by pulse
     # (because they are recorded chronologically)
     # but for reduction it is more useful to bin by detector id
-    events = sc.bin(detector_data.events, groups=[detector_data.detector_ids])
+    events = sc.bin(detector_data.events,
+                    groups=[detector_data.detector_ids],
+                    edges=[_tof_edges])
```
SimonHeybrock, line 297: Can you please verify if this has performance implications? It might, since we are now "binning" in 2d instead of 1d. If it does, we can instead add the extra dim and coord afterwards.

Tom-Willemsen: I looked at the performance in a comment above

> I just ran some benchmarks. There is a bit of fluctuation, but on this branch the average on my machine for loading SANS2D_ESS_Example1.nxs is approx 1.8 seconds. On main, it takes approx 1.7 seconds.

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: Because otherwise min will be `double_max` and max will be `double_min`, which would be the wrong way round. I could just invert min and max in this case if you think that's cleaner.

SimonHeybrock: Hmm, have you tried if `plot()` works with INF boundaries?

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: I agree that this isn't ideal. However, I'm also not sure that we can just use floating-point `epsilon` directly as `value + epsilon` will only be larger than `value` if `value <= 1`

```
>>> 0.5 + np.finfo("float64").eps == 0.5
False
>>> 5 + np.finfo("float64").eps == 5
True
```

So in this example if we use `eps` we may run into the last values in the ToF bin being excluded.

Furthermore, the ToF data may not even be floating-point, it is allowed to just be integer, in which case the `+1` is the smallest we could use.

I've made it configurable as a constant but left it at 1 for now, happy to discuss further if you like

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: On further reflection I decided to just swap the limits, `Inf` has a lot of potential edge-cases and pitfalls when using the data later on (e.g. the plotting one you mention)

python/src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Use https://numpy.org/doc/stable/reference/generated/numpy.nextafter.html

SimonHeybrock: Apparently Python 3.9 will also have it: https://stackoverflow.com/a/61119174 (but we still require older versions).

SimonHeybrock (approved):
Looks good, but there is a conflict to resolve.

OwenArnold (approved):

PR #128, opened 2021-08-09
Docs build tooling

jl-wynen:
Use the same CMake / Python scripts as in scipp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,10 +150,16 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 # Custom target for building tests. all excludes tests by default
 add_custom_target(all-tests)
 
-# Custom target for building documentation
+# Custom install target for docs to depend on.
 add_custom_target(
-  docs COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/docs/build_docs.py
+  install-scippneutron COMMAND cmake --build ${CMAKE_CURRENT_BINARY_DIR}
+                               --target install
 )
+
+include(docs)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -150,11 +150,17 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
...
-# Custom target for building documentation
+# Custom install target for docs to depend on.
 add_custom_target(
-  docs COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/docs/build_docs.py
+  install-scippneutron COMMAND cmake --build ${CMAKE_CURRENT_BINARY_DIR}
+                               --target install
 )
 
+include(docs)
+add_docs_target(docs BUILDER html DEPENDS install-scippneutron)
```
OwenArnold, line 160: Why is it not possible to depend on the `install` target directly?

jl-wynen: You can only depend on targets that you defined yourself, not targets generated by CMake: https://cmake.org/cmake/help/latest/command/add_dependencies.html

OwenArnold: Ah yes. I think I've encountered this before. Thanks for reminding me.

CMakeLists.txt
```diff
@@ -150,11 +150,17 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
...
+add_docs_target(doctest BUILDER doctest DEPENDS docs)
+add_docs_target(linkcheck BUILDER linkcheck DEPENDS docs)
```
OwenArnold, line 162: Missing `docs.cmake` (like scipp has), unknown function

jl-wynen: Oh, right

PR #129, opened 2021-08-13
Add source/sample components to the instrument view

### LINKED ISSUE #23: Include beamline components in instrument view
How can we display other beamline components in the instrument view, not just pixels?
- monitors
- source
- sample

OwenArnold:
Fixes #23

Basically `instrument_view` will take a dictionary called `components`, which maps the name of something considered "meta" (such that we can establish its position - sc.vector) to one of a bunch of known shapes. shape names are mapped to drawing functions which take care of scaling and orienting the shapes depending upon the beam direction and overall component distribution (distances)

Code in `ess` will allow us to forward these dictionaries on an instrument by instrument basis. One could imaging having a `odin.instrument_view`, which would internally provide the full `components` dictionary.

### DIFF
--- a/python/src/scippneutron/instrument_view.py
+++ b/python/src/scippneutron/instrument_view.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
-# @author Neil Vaytet
+# @author Neil Vaytet and Owen Arnold
 
 import numpy as np
 import pythreejs as p3
@@ -50,9 +50,9 @@ def _find_beam(det_com, pos):
 
 
 def _alignment_matrix(to_align, target):
-    rotaxis = np.cross(to_align, target)
+    rot_axis = np.cross(to_align, target)
     magnitude = np.linalg.norm(to_align) * np.linalg.norm(target)
-    axis_angle = np.arcsin(rotaxis / magnitude)
+    axis_angle = np.arcsin(rot_axis / magnitude)
     return Rot.from_rotvec(axis_angle).as_matrix()
 
 
@@ -151,7 +151,8 @@ def _plot_components(scipp_obj, additional, positions_var, scene):
                       det_center=det_center)
     # Reset camera
     camera = _get_camera(scene)
-    camera.far = furthest_distance * 5.0
+    if camera:
+        camera.far = furthest_distance * 5.0
 
 
 def _get_camera(scene):
--- a/python/src/scippneutron/instrument_view.py
+++ b/python/src/scippneutron/instrument_view.py
@@ -4,6 +4,7 @@
 
 import numpy as np
 import pythreejs as p3
+import scipp as sc
 
 
 def _cube(position, message, sz, color):
@@ -73,16 +74,43 @@ def _add_sample(position, scene, sz=(1, 1, 1)):
     _unpack_to_scene(scene, _cube(position, message="sample", sz=sz, color="#808080"))
 
 
-def _add_beamline(scipp_obj, scene):
-    # TODO calculate sizes
+def _calculate_component_width(det_center, component, fixed_width=None):
+    if fixed_width:
+        return fixed_width
+    else:
+        scaling_factor = 1 / 5.0
+        return sc.norm(det_center - component).value * scaling_factor
 
-    if "source_position" in scipp_obj.meta:
-        _add_source(tuple(scipp_obj.meta["source_position"].value), scene, sz=(1, 1, 1))
+
+def _add_beamline(scipp_obj, positions_var, scene):
+    sample_width = None
+    det_center = sc.mean(positions_var)
+    beamline_comp_positions = []
+
     if "sample_position" in scipp_obj.meta:
-        _add_sample(tuple(scipp_obj.meta["sample_position"].value),
-                    scene,
-                    sz=(0.5, 0.5, 0.5))
-    # TODO reset camera far field
+        sample_width = _calculate_component_width(det_center,
+                                                  scipp_obj.meta["sample_position"])
+        sample = scipp_obj.meta["sample_position"]
+        beamline_comp_positions.append(sc.norm(sample - det_center).value)
+        _add_sample(tuple(sample.value), scene, sz=([sample_width] * 3))
+    if "source_position" in scipp_obj.meta:
+        source_width = _calculate_component_width(det_center,
+                                                  scipp_obj.meta["source_position"],
+                                                  sample_width)
+        source = scipp_obj.meta["source_position"]
+        beamline_comp_positions.append(sc.norm(source - det_center).value)
+        _add_source(tuple(source.value), scene, sz=([source_width] * 3))
+
+    if not beamline_comp_positions:
+        return None
+    return np.max(np.array(beamline_comp_positions)) * 5.0
+
+
+def _get_camera(plt):
+    for child in plt.view.figure.scene.children:
+        if isinstance(child, p3.PerspectiveCamera):
+            return child
+    return None
 
 
 def instrument_view(scipp_obj=None,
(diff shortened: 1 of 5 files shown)


nvaytet:
I think I agree with @SimonHeybrock 's suggestion of moving this into the `ess` repo.

I initially thought that it would be ok to do some of the common things here, because the `source_position` and `sample_position` are already special keys we rely on for unit conversions, but actually this is also going away with the new way to handle conversions via the `transform_coords` method.

We would leave the instrument view as it is in scippneutron, so that one can still display non-ESS instruments, but if you want to load all the components for loki, the special keys should be defined on a per-instrument basis.

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
 # @author Neil Vaytet
 
 import numpy as np
+import pythreejs as p3
+import scipp as sc
 
 
-def instrument_view(scipp_obj=None, positions="position", pixel_size=None, **kwargs):
+def _cube(position, message, sz, color):
```
nvaytet, line 10: I find the variable naming a little obscure here.
- I think it is not necessarily a `cube`? the box can have different xyz extents?
- `sz` is not very explicit
- `message` sounds like it has some communication aspect, but in fact it's just the text to be displayed above the component?

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
...
+    geometry = p3.BoxGeometry(width=sz[0],
```
nvaytet, line 11: I think `sz[0]` is hard to follow. I would either make a dict `{height: 1, ...}` or just use 3 separate variables?

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
...
+    if fixed_width:
+        return fixed_width
+    else:
+        scaling_factor = 1 / 5.0
+        return sc.norm(det_center - component).value * scaling_factor
+
+
+def _add_beamline(scipp_obj, positions_var, scene):
+    sample_width = None
+    det_center = sc.mean(positions_var)
```
nvaytet, line 87: I think you can get this information from the camera, and then you wouldn't need to compute the mean anymore.

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
...
+        source_width = _calculate_component_width(det_center,
+                                                  scipp_obj.meta["source_position"],
+                                                  sample_width)
+        source = scipp_obj.meta["source_position"]
+        beamline_comp_positions.append(sc.norm(source - det_center).value)
+        _add_source(tuple(source.value), scene, sz=([source_width] * 3))
+
+    if not beamline_comp_positions:
+        return None
+    return np.max(np.array(beamline_comp_positions)) * 5.0
```
nvaytet, line 106: It's a little strange that `_add_beamline` returns a distance between beamline components.
Maybe you can reset the camera in place here? You may need to pass the camera in anyway, if you want to avoid using `positions_var` as suggested above?

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
...
+
+def _add_beamline(scipp_obj, positions_var, scene):
+    sample_width = None
+    det_center = sc.mean(positions_var)
+    beamline_comp_positions = []
+
+    if "sample_position" in scipp_obj.meta:
+        sample_width = _calculate_component_width(det_center,
+                                                  scipp_obj.meta["sample_position"])
+        sample = scipp_obj.meta["sample_position"]
```
nvaytet, line 93:
```suggestion
        sample_position = scipp_obj.meta["sample_position"]
```

python/src/scippneutron/instrument_view.py
```diff
@@ -3,10 +3,129 @@
...
+        beamline_comp_positions.append(sc.norm(sample - det_center).value)
+        _add_sample(tuple(sample.value), scene, sz=([sample_width] * 3))
+    if "source_position" in scipp_obj.meta:
+        source_width = _calculate_component_width(det_center,
+                                                  scipp_obj.meta["source_position"],
+                                                  sample_width)
+        source = scipp_obj.meta["source_position"]
```
nvaytet, line 100:
```suggestion
        source_position = scipp_obj.meta["source_position"]
```

nvaytet (changes requested):

python/src/scippneutron/instrument_view.py
```diff
@@ -116,13 +118,14 @@ def _get_camera(plt):
 def instrument_view(scipp_obj=None,
                     positions="position",
                     pixel_size=None,
-                    plot_non_pixels=True,
+                    also_plot=None,
```
SimonHeybrock, line 121: How about `components` or something like that?

OwenArnold: image
Not done yet, but this gives a flavour of how things would be done

SimonHeybrock: 👍

python/src/scippneutron/instrument_view.py
OwenArnold: I did think of this, but was struggling to figure out how to read that information out of the camera. Camera has lookAt, but there doesn't seem to be a way that I could see to read this back without some complex logic based on the frustrum coordinates.

python/src/scippneutron/instrument_view.py
```diff
@@ -1,12 +1,181 @@
...
+
+
+def _plot_components(scipp_obj, additional, positions_var, scene):
+    det_center = sc.mean(positions_var)
+    furthest_key, furthest_distance = _furthest_component(det_center, scipp_obj,
+                                                          additional)
+    # Some scaling to set width according to distance from detector center
+    scaling_factor = 1 / 10.0
+    width = furthest_distance * scaling_factor
+    shapes = {"cube": _box, "cylinder": _cylinder, "disk": _disk_chopper}
```
nvaytet, line 139: Could we just name these `monitor`, `source`, `sample` `chopper` instead of `cube` and `box`, `cylinder` etc...
i.e. standardize the way each instrument represents a monitor and a sample?
Don't know if we also want to standardize the colors for each component?

nvaytet (changes requested):

python/src/scippneutron/instrument_view.py
OwenArnold: Ok will do. My reasoning was that some instruments might have a preference on how they represented components of each type, i.e use a cube for the sample and a cylinder for the source interchangeably - I think we get more convenience with your approach so lets take that. I think plotting all these non-detector components as grey is sufficient. Maybe we can consider upgrading the colours if we get specific requests about this.

SimonHeybrock: Hmm, might want to decouple the shape from the category (monitor, source, ...). Maybe just standardize the color<->category mapping?

SimonHeybrock: Consider someone with an instrument with a unique component. We would constantly get new request ("I need analyzer crystals", "I need slits", "I need XYZ", ...). I feel *not* having hard-coded names (or exclusively hard-coded) would be the better option.

nvaytet: So do you want "not hard-coded" or "exclusively hard-coded"?
I didn't really get what was the conclusion.
I think you would slowly build a library of component representations, slits, etc.. and over time I don't think we will get more and more requests, just a bit to start with.

python/src/scippneutron/instrument_view.py
```diff
@@ -21,14 +190,23 @@ def instrument_view(scipp_obj=None, positions="position", pixel_size=None, **kwa
...
-                pixel_size=pixel_size,
-                **kwargs)
+    plt = plot(scipp_obj,
+               projection="3d",
+               positions=positions,
+               pixel_size=pixel_size,
+               **kwargs)
+
+    # Add additional components from the beamline
+    if components and not isinstance(plot, PlotDict):
```
SimonHeybrock, line 208: Is this meant to read `plt`? `plot` is just the function?

OwenArnold: Yes, that's a bug.

SimonHeybrock:
Given that this is a new feature, I would assume that we will have to iterate on this. Potential further comments may thus be addressed at a later point, I'll merge so we can try this out in practice.

PR #143, opened 2021-08-20
Read nxdisk chopper

### LINKED ISSUE #137: load-nexus needs to load NXdisk_choppers
Need this information for wfm, visualisation and many other purposes.

OwenArnold:
Read `NXdisk_chopper` information via `scippneutron`. Fixes #137

### DIFF
--- a/python/src/scippneutron/file_loading/_json_nexus.py
+++ b/python/src/scippneutron/file_loading/_json_nexus.py
@@ -217,13 +217,10 @@ class LoadFromJson:
         except MissingAttribute:
             units = sc.units.dimensionless
 
-        common_creation_args = {'dtype': dtype, 'unit': units}
-        if isinstance(dataset[_nexus_values], list):
-            return sc.array(dims=dimensions,
-                            values=dataset[_nexus_values],
-                            **common_creation_args)
-        else:
-            return sc.scalar(value=dataset[_nexus_values], **common_creation_args)
+        return sc.array(dims=dimensions,
+                        values=np.asarray(dataset[_nexus_values]),
+                        dtype=dtype,
+                        unit=units)
 
     def load_dataset_from_group_as_numpy_array(self, group: Dict, dataset_name: str):
         """
--- a/python/src/scippneutron/file_loading/_json_nexus.py
+++ b/python/src/scippneutron/file_loading/_json_nexus.py
@@ -196,7 +196,7 @@ class LoadFromJson:
         Load a dataset into a Scipp Variable (array or scalar)
         :param group: Group containing dataset to load
         :param dataset_name: Name of the dataset to load
-        :param dimensions: Dimensions for the output Variable
+        :param dimensions: Dimensions for the output Variable. If empty, yields scalar.
         :param dtype: Cast to this dtype during load,
           otherwise retain dataset dtype
         """
@@ -217,13 +217,13 @@ class LoadFromJson:
         except MissingAttribute:
             units = sc.units.dimensionless
 
+        common_creation_args = {'dtype': dtype, 'unit': units}
         if isinstance(dataset[_nexus_values], list):
             return sc.array(dims=dimensions,
                             values=dataset[_nexus_values],
-                            dtype=dtype,
-                            unit=units)
+                            **common_creation_args)
         else:
-            return sc.scalar(value=dataset[_nexus_values], dtype=dtype, unit=units)
+            return sc.scalar(value=dataset[_nexus_values], **common_creation_args)
 
     def load_dataset_from_group_as_numpy_array(self, group: Dict, dataset_name: str):
         """
--- a/python/src/scippneutron/file_loading/_json_nexus.py
+++ b/python/src/scippneutron/file_loading/_json_nexus.py
@@ -190,7 +190,7 @@ class LoadFromJson:
     def load_dataset(self,
                      group: Dict,
                      dataset_name: str,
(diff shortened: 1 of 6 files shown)


python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -65,6 +66,21 @@ def _load_instrument_name(instrument_groups: List[Group], data: ScippData,
                                     "instrument_name", data, nexus)
 
 
+def _load_chopper(chopper_groups: List[Group], data: ScippData, nexus: LoadFromNexus):
+    for chopper_group in chopper_groups:
+        chopper_name = chopper_group.path.split("/")[-1]
+        print("chopper_name", chopper_name)
```
Tom-Willemsen, line 72: Probably remove this debug print?

python/tests/nexus_helpers.py
```diff
@@ -241,7 +250,9 @@ def add_dataset(parent: Dict, name: str, data: Union[str, np.ndarray]) -> Dict:
         if isinstance(data, str):
             dataset_info = {"string_size": len(data), "type": "string"}
         elif isinstance(data, float):
-            dataset_info = {"size": 1, "type": "float32"}
+            dataset_info = {"size": 1, "type": "float64"}  # This is worth considering!
```
Tom-Willemsen, line 253: > This is worth considering

Do we have an issue to consider this? I think we may forget about it if it is just a comment in the code...

python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -26,6 +26,7 @@
 nx_sample = "NXsample"
 nx_source = "NXsource"
 nx_detector = "NXdetector"
+nx_chopper = "NXdisk_chopper"
```
Tom-Willemsen, line 29: Consider calling this `nx_disk_chopper` or similar to disambiguate from fermi choppers etc which we may have to support later? This would also keep it consistent with nexus class names

OwenArnold: The name is set by the Nexus standard i.e. https://manual.nexusformat.org/classes/base_classes/NXdisk_chopper.html?highlight=nxdisk_chopper but maybe i misunderstand your comment. You are correct though, the standard does support https://manual.nexusformat.org/classes/base_classes/NXfermi_chopper.html#nxfermi-chopper

python/src/scippneutron/file_loading/load_nexus.py
OwenArnold: Yes absolutely thanks Tom

python/tests/nexus_helpers.py
OwenArnold: Prior to this, floats are being written as `float32`. I don't think this is right, but maybe I can call @matthew-d-jones  in to comment on why this was done in the first place.

python/src/scippneutron/file_loading/load_nexus.py
Tom-Willemsen: Sorry, perhaps my comment wasn't entirely clear - I was suggesting renaming the python variable to match the nexus class name

```suggestion
nx_disk_chopper = "NXdisk_chopper"
```

python/tests/nexus_helpers.py
SimonHeybrock: I suppose it should not matter much, since this is just for the tests?

python/src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -218,12 +218,12 @@ def load_dataset(self,
...
-                               dtype=dtype,
-                               unit=units)
-
-        raise NotImplementedError("Loading scalar datasets not implemented")
+            return sc.array(dims=dimensions,
+                            values=dataset[_nexus_values],
+                            dtype=dtype,
+                            unit=units)
+        else:
+            return sc.scalar(value=dataset[_nexus_values], dtype=dtype, unit=units)
```
SimonHeybrock, line 226: Is the branching actually needed? If `dims=[]` the `if` branch should do the same as the `else` does?

OwenArnold: The overloads for variable creation with scipp 0.7 don't seem to support this. Empty dims does not seem to be enough to determine scalar creation. I'm made a small refactor anyway.

SimonHeybrock: Works for me:
```python
import scipp as sc
import numpy as np
a = np.array(3)
print(sc.array(dims=[], values=a))
print(sc.__version__)
```
```
<scipp.Variable> ()      int64  [dimensionless]  [3]
0.7.1
```

OwenArnold: Yes, this works, I was trying with a bare `float` rather than a single value in a `np.array`. I'll change.

python/tests/nexus_helpers.py
```diff
@@ -325,6 +336,7 @@ def __init__(self):
         self._detectors: List[Detector] = []
         self._logs: List[Log] = []
         self._instrument_name: Optional[str] = None
+        self._choppers: Optional[Chopper] = []
```
Tom-Willemsen, line 339: Should this be 
```suggestion
        self._choppers: Optional[List[Chopper]] = []
```
? otherwise not sure the default makes sense.

Issue #140, opened 2021-08-19
`load_logs` fails to load non-utf8 string attributes


Tom-Willemsen:
For example, using [this file](https://testdata.mantidproject.org/ftp/external-data/MD5/9f9ac3443b96e9ee8d62674d31a152dc), `load_nexus` fails because the log for `Attenuator_angle` has units of `°` (ascii `0xb0`) which is invalid in utf-8 as it is in the extended ascii range.

The nexus format specifies the string encoding of attributes, we should load using the specified encoding rather than always assuming utf-8.

Stack trace:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\load_nexus.py", line 116, in load_nexus
    loaded_data = _load_data(nexus_file, root, LoadFromHdf5(), quiet)
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\load_nexus.py", line 169, in _load_data
    load_logs(loaded_data, groups[nx_log], nexus, run_start_time=run_start_time)
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\_log_data.py", line 18, in load_logs
    log_data_name, log_data = _load_log_data_from_group(
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\_log_data.py", line 142, in _load_log_data_from_group
    unit = nexus.get_unit(nexus.get_dataset_from_group(group.group, value_dataset_name))
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\_hdf5_nexus.py", line 155, in get_unit
    return _ensure_str(units)
  File "C:\Users\wilt\Documents\11015_ESS_SCIPP\scippneutron\install\scippneutron\file_loading\_hdf5_nexus.py", line 22, in _ensure_str
    str_or_bytes = str(str_or_bytes, encoding="utf8")  # type: ignore
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb0 in position 0: invalid start byte
```

PR #144, opened 2021-08-20
Fix units encoding error

Tom-Willemsen:
Closes https://github.com/scipp/scippneutron/issues/140

This PR ensures that we always load string datasets and attributes with the encodings specified in the nexus file.

For ASCII-encoded data in the nexus file, it allows the common extended ascii (latin-1) format, but warns as this is technically unspecified behaviour. Data files from ISIS do sometimes use these extended ASCII code points (see example LET data file linked to ticket). Data marked as UTF-8 will still be loaded correctly using that encoding.

See also https://docs.h5py.org/en/stable/strings.html

### DIFF
--- a/python/src/scippneutron/file_loading/_hdf5_nexus.py
+++ b/python/src/scippneutron/file_loading/_hdf5_nexus.py
@@ -1,28 +1,54 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # @author Matthew Jones
+import warnings
 
 from typing import Union, Any, List, Optional, Tuple, Dict
 
 import h5py
+import h5py.h5a
+import h5py.h5t
 import numpy as np
 import scipp as sc
 from ._common import Group, MissingDataset, MissingAttribute
 
 
 def _get_attr_as_str(h5_object, attribute_name: str):
-    try:
-        return h5_object.attrs[attribute_name].decode("utf8")
-    except AttributeError:
-        return h5_object.attrs[attribute_name]
+    return _ensure_str(h5_object.attrs[attribute_name],
+                       LoadFromHdf5.get_string_encoding(h5_object, attribute_name))
 
 
-def _ensure_str(str_or_bytes: Union[str, bytes]) -> str:
-    try:
-        str_or_bytes = str(str_or_bytes, encoding="utf8")  # type: ignore
-    except TypeError:
-        pass
-    return str_or_bytes
+def _ensure_str(str_or_bytes: Union[str, bytes], encoding: str) -> str:
+    """
+    See https://docs.h5py.org/en/stable/strings.html for justification about some of
+    the operations performed in this method. In particular, variable-length strings
+    are returned as `str` from h5py, but need to be encoded using the surrogateescape
+    error handler and then decoded using the encoding specified in the nexus file in
+    order to get a correctly encoded string in all cases.
+
+    Note also that the nexus standard leaves unspecified the behaviour of H5T_CSET_ASCII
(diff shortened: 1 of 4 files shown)


python/src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -10,19 +11,55 @@
...
+
+    if encoding == "ascii":
+        try:
+            return str(str_or_bytes, encoding="ascii")
+        except UnicodeDecodeError as e:
+            decoded = str(str_or_bytes, encoding="latin-1")
+            warnings.warn(f"Encoding for bytes '{str_or_bytes}' declared as ascii, "
+                          f"but contains characters in extended ascii range. Assuming "
+                          f"extended ASCII (latin-1), but this behaviour is not "
+                          f"specified by the HDF5 or nexus standards and may therefore "
```
OwenArnold, line 57: Minor point, we are using American English spelling on the project `behaviour` -> `behavior`

python/src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -10,19 +11,55 @@
 from ._common import Group, MissingDataset, MissingAttribute
 
 
+def cset_to_encoding(cset: int):
```
OwenArnold, line 14: minor rename `_cset_to_encoding` also you are inconsistently applying type hinting for the return types in these functions, though you do for most arguments.

python/tests/nexus_helpers.py
```diff
@@ -36,25 +36,25 @@ class EventData:
     event_time_offset: Optional[np.ndarray]
     event_time_zero: Optional[np.ndarray]
     event_index: Optional[np.ndarray]
-    event_time_zero_unit: Optional[str] = "ns"
-    event_time_zero_offset: Optional[str] = "1970-01-01T00:00:00Z"
-    event_time_offset_unit: Optional[str] = "ns"
+    event_time_zero_unit: Optional[Union[str, bytes]] = "ns"
```
OwenArnold, line 39: Minor: Might be worth defining this Union somewhere common as is heavily used - though maybe doesn't make it much shorter or clearer?

OwenArnold (changes requested):
Overall I think what you have done makes sense.

python/tests/nexus_helpers.py
SimonHeybrock: I'd rather have this explicit than having to look up another name.

Tom-Willemsen: I think on balance I agree with @SimonHeybrock that it's easier to understand as is? I can see the arguments both ways though and happy to change if you feel strongly about it.

SimonHeybrock:
@OwenArnold Did you have more comments on this?

python/src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -10,19 +11,66 @@
...
+    """
+    if cset == h5py.h5t.CSET_ASCII:
+        return "ascii"
+    elif cset == h5py.h5t.CSET_UTF8:
+        return "utf-8"
+    else:
+        warnings.warn(f"Unknown character set in HDF5 data file. Expected data "
+                      f"types are {h5py.h5t.CSET_ASCII} (H5T_CSET_ASCII) or "
+                      f"{h5py.h5t.CSET_UTF8} (H5T_CSET_UTF8) but got '{cset}'. "
+                      f"Assuming data is UTF-8 encoded but this may be incorrect.")
```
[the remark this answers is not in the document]
Tom-Willemsen: No, not aware of any facilities that do this. It's not actually allowed by the nexus standard. I've changed it to an error.

PR #145, opened 2021-08-23
Add no-setup option to build_docs.py

nvaytet:
Depends on https://github.com/scipp/pipelines/pull/13

We are also now downloading the data files for the docs only once (the setup is skipped when running doctest)

### DIFF
--- a/docs/build_docs.py
+++ b/docs/build_docs.py
@@ -4,55 +4,62 @@
 
 import os
 import argparse
-import urllib
 import requests
-import re
 from pathlib import Path
 import subprocess
 import sys
+import tarfile
 
 parser = argparse.ArgumentParser(description='Build doc pages with sphinx')
 parser.add_argument('--prefix', default='build')
 parser.add_argument('--work_dir', default='.doctrees')
 parser.add_argument('--data_dir', default='data')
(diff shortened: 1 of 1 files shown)


docs/build_docs.py
```diff
@@ -4,82 +4,76 @@
...
+
+    args = parser.parse_known_args()[0]
+
+    docs_dir = Path(__file__).parent.absolute()
+    work_dir = _get_abs_path(path=args.work_dir, root=docs_dir)
+    prefix = _get_abs_path(path=args.prefix, root=docs_dir)
+    data_dir = _get_abs_path(path=args.data_dir, root=docs_dir)
+
+    if not args.no_setup:
+        _setup(data_dir=data_dir)
```
OwenArnold, line 75: If the `data_dir` is different from what was used already, don't we want to ensure that `_setup` is still run?

OwenArnold (approved):
Approve anyway, as this prevents broken builds

docs/build_docs.py
nvaytet: I the person running the script is explicitly requesting no setup via a command line argument, I think the setup should be skipped?

PR #147, opened 2021-08-25
Use scippbuildtools for building

nvaytet:
This is a first illustration of what the build setup could look like if we use a small conda package containing the helper functions that reduce the duplication between the repositories.

The files that use the common repo `scippbuildtools` are:
- `tools/build_cpp.py`
- `tools/build_conda.py`
- `docs/build_docs.py`
- `docs/conf.py`
- `docs/_static/theme_overrides.css`

If we are happy with this, we can then apply it to all the other repos.

Depends on https://github.com/scipp/pipelines/pull/15


jl-wynen:
The new plan is to make a template repo to generate build files for scipp[neutron] / ess.

PR #152, opened 2021-09-01
Add recipes page to docs

SimonHeybrock:
Nothing fancy for now, and lacking testing, this is just to make a start. I suggest we go with this for a while to figure out if this is a desirable solution, or whether we want fully functional and self-contained examples (which might be much harder to parse and read through).

### DIFF
--- a/docs/user-guide/recipes.ipynb
+++ b/docs/user-guide/recipes.ipynb
@@ -83,11 +83,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Here we assume that the input has only a single bin along the 'tof' dimension\n",
-    "tof_interval = sc.concatenate(tof_mask_min, tof_mask_max, 'tof')\n",
-    "tof = sc.sort(sc.concatenate(da.coords['tof'], tof_interval, 'tof'), key='tof')\n",
-    "da = sc.bin(da, edges=[tof])  # update existing binning with new bounds\n",
-    "da.masks['prompt_pulse'] = (tof >= tof_mask_min) & (tof < tof_mask_max)"
+    "tof = sc.array(dims=['tof'], unit='ms', values=[tof_min, mask_start, mask_end, tof_max])\n",
+    "da = sc.bin(da, edges=[tof])  # bin in 'tof', updating prior 'tof' binning if present\n",
+    "da.masks['prompt_pulse'] = (tof >= tof['tof', 1]) & (tof < tof['tof', 2])"
    ]
   },
   {
@@ -113,7 +111,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "da['spectrum', index].plot()  # plot spectrum with index 0\n",
+    "da['spectrum', index].plot()  # plot spectrum with given index\n",
     "da['spectrum', sc.scalar(spectrum_number)]  # plot spectrum with given spectrum-number, provided da.coords['spectrum'] exists"
    ]
   },
@@ -135,7 +133,7 @@
     "spectra = {}\n",
     "spectra['name1'] = da1['spectrum', index1]\n",
     "spectra['name2'] = da1['spectrum', index2]\n",
-    "spectra['name3'] = da2['spectrum', index3]  # Will use different axes unless units and dimensions\n",
+    "spectra['name3'] = da2['spectrum', index3]\n",
     "sc.plot(spectra)"
    ]
   }
--- a/docs/user-guide/recipes.ipynb
+++ b/docs/user-guide/recipes.ipynb
@@ -1,0 +1,167 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,167 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "104b0d14-6ee6-4559-97b0-6c98dc15a26f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Here we assume that the input has only a single bin along the 'tof' dimension\n",
```
jl-wynen, line 86: Is the example still useful enough like this?

We should have a high level function that does it.

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,167 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "52e1e178-e729-49eb-8bec-fa9fbc60ece8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "da['spectrum', index].plot()  # plot spectrum with index 0\n",
```
jl-wynen, line 116: Not actually showing for index 0.

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,167 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c4aadf8d-1fc7-406d-87b8-0d7dbb7a3117",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "spectra = {}\n",
+    "spectra['name1'] = da1['spectrum', index1]\n",
+    "spectra['name2'] = da1['spectrum', index2]\n",
+    "spectra['name3'] = da2['spectrum', index3]  # Will use different axes unless units and dimensions\n",
```
jl-wynen, line 138: The comment seems to have lost its ending.
Also, it might be more confusing than helpful. People might wonder what even are 'axes'.

jl-wynen (commented):
Didn't you want to make this testable?

docs/user-guide/recipes.ipynb
SimonHeybrock: removed

docs/user-guide/recipes.ipynb
SimonHeybrock: I think the example (masking event data) is important, since this will be a common application (in fact it was required for both reflectometry and powder diffraction independently).

I agree that the code to make the edges is probably not too helpful. I will change to just explicitly create some edges. We can address creating new edges based on existing ones in a another recipe, once we have a better idea how to best implement it.

jl-wynen:
>But how do we hide output? Just not execute the notebook as part of CI, but just run manually?

Can we set up pytest to execute the notebook using https://testbook.readthedocs.io/en/latest/usage/index.html#using-execute-to-control-which-cells-are-executed-before-test and https://docs.next.tech/creator/how-tos/testing-techniques/testing-jupyter-notebook-code-with-pytest
But we might have to run every cell on its own and also name the results of function calls so that we can refer to them in the test.

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,165 @@
+{
```
nvaytet, line 1: Not sure why you want to hide the output, couldn't we just run this as a normal notebook? or do you think once you start adding plot output it will dilute the focus of this page?
It would be the easiest to just run this notebook, which would make sure recipes are never broken.

SimonHeybrock: This would be intended to become a longer list, a bit like http://xarray.pydata.org/en/stable/howdoi.html (but maybe slightly more complex/concrete), so yes, I think it would dilute the focus. Maybe full examples could be on sub-pages?

nvaytet: Would be useful if we could have buttons to hide or show the output. Jupyter book has this but I don't think Sphinx does: https://jupyterbook.org/interactive/hiding.html

SimonHeybrock:
> Would be useful if we could have buttons to hide or show the output.

Yes, may be useful in some cases.

Should we close this until someone comes up with a more sophisticated solution?

nvaytet:
> > Would be useful if we could have buttons to hide or show the output.
> 
> Yes, may be useful in some cases.
> 
> Should we close this until someone comes up with a more sophisticated solution?

Is using Jupyter-book to build our docs an option? https://jupyterbook.org/start/create.html
It's using sphinx under the hood. I don't know how it handles the generation of an API reference, etc...
But it handles notebooks and markdown files out of the box.

SimonHeybrock:
Should we close this, or merge?

jl-wynen (approved):
I think it is fine to merge this. There is no harm done in replacing the page later.

PR #153, opened 2021-09-01
Better load error messages

### LINKED ISSUE #151: Validate filenames before calling Mantid
Calling `scn.load(filename='bogus.nxs')`with a non-existent file currently produces the following error
```
Traceback (most recent call last):
  File "/home/jl/Software/Miniconda3/envs/nscipp-2/lib/python3.7/site-packages/mantid/simpleapi.py", line 139, in Load
    algm.setProperty('Filename', filename)  # Must be set first
ValueError: Invalid value for property Filename (list of str lists) from string "bogus.nxs": When setting value of property "Slave": File "bogus.nxs" not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "asd.py", line 4, in <module>
    da = scn.load(filename='bogus.nxs')
  File "/home/jl/Work/nscipp-2/install/scippneutron/mantid.py", line 839, in load
    with run_mantid_alg(mantid_alg, filename, **mantid_args) as loaded:
  File "/home/jl/Software/Miniconda3/envs/nscipp-2/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/home/jl/Work/nscipp-2/install/scippneutron/mantid.py", line 35, in run_mantid_alg
    ws = getattr(mantid, alg)(*args, **kwargs)
  File "/home/jl/Software/Miniconda3/envs/nscipp-2/lib/python3.7/site-packages/mantid/simpleapi.py", line 141, in Load
    msg = 'Problem setting "Filename" in {}-v{}: {}'.format(name, algm.name(), algm.version(),
NameError: name 'name' is not defined
```
which, apart from showing a bug in Mantid, is not particularly easy to understand. We can do better.

OwenArnold:
Fixes #151

It's not as simple as just checking the file path i.e. `os.path.exists` or similar because Mantid will also happily load based on run numbers or incomplete file paths that are taken to be joined with known `Config` data search directories. We therefore have to ask mantid if it can load the file. There are two routes to doing this.

1. `mantid.api.FileFinder`
1. `mantid.api.FindLoaderRegistry`

Preferred option was to use the first as easier to interpret results via return directly. However this function appears to hang on an Instrument Download step in the framework manager initialisation, I am therefore using 2 even though it relies on catching specific exception types and allowing others through.

### DIFF
--- a/python/src/scippneutron/mantid.py
+++ b/python/src/scippneutron/mantid.py
@@ -862,8 +862,21 @@ def load(filename="",
                            advanced_geometry=advanced_geometry)
 
 
+def _is_mantid_loadable(filename):
+    from mantid.api import FileFinder
+    if FileFinder.getFullPath(filename):
+        return True
+    else:
+        try:
+            # findRuns throws rather than return empty so need try-catch
+            FileFinder.findRuns(filename)
+            return True
+        except Exception:
+            return False
+
+
 def _check_file_path(filename, mantid_alg):
-    from mantid.api import FileFinder, AlgorithmManager, FrameworkManager, FileProperty
+    from mantid.api import AlgorithmManager, FrameworkManager, FileProperty
     FrameworkManager.Instance()
     alg = AlgorithmManager.createUnmanaged(mantid_alg)
     filename_property = [
@@ -875,9 +888,9 @@ def _check_file_path(filename, mantid_alg):
     # is called directly we attempt to find FileProperty. Otherwise paths should be
     # absolute
     if filename_property or mantid_alg == 'Load':
-        if not FileFinder.getFullPath(filename):
+        if not _is_mantid_loadable(filename):
             raise ValueError(
-                f"Mantid cannot find file {filename} and therefore will not load it."
+                f"Mantid cannot find {filename} and therefore will not load it."
             ) from None
     else:
         if not os.path.isfile(filename):
--- a/python/src/scippneutron/mantid.py
+++ b/python/src/scippneutron/mantid.py
@@ -9,6 +9,7 @@ import uuid
 import warnings
 import numpy as np
 import scipp as sc
+import os
 
 
 @contextmanager
@@ -834,16 +835,11 @@ def load(filename="",
              instrument geometry.
     :rtype: Dataset
     """
-    from mantid.api import FileLoaderRegistry
 
     if mantid_args is None:
         mantid_args = {}
 
-    try:
-        FileLoaderRegistry.chooseLoader(filename)
-    except ValueError:
-        raise ValueError(
-            f"Mantid cannot find file {filename} and therefore will not load it.")
+    _check_file_path(filename, mantid_alg)
 
     with run_mantid_alg(mantid_alg, filename, **mantid_args) as loaded:
         # Determine what Load has provided us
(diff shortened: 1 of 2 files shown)


python/src/scippneutron/mantid.py
```diff
@@ -834,18 +834,25 @@ def load(filename="",
...
+    from mantid.api import FileLoaderRegistry
 
     if mantid_args is None:
         mantid_args = {}
 
+    try:
+        FileLoaderRegistry.chooseLoader(filename)
+    except ValueError:
+        raise ValueError(
+            f"Mantid cannot find file {filename} and therefore will not load it.")
```
SimonHeybrock, line 846: Looks good, and should be simple enough to add a unit test that we get the expected `ValueError` from `scn.load(filename='does-not-exist.nxs')`?

python/src/scippneutron/mantid.py
```diff
@@ -834,18 +834,25 @@ def load(filename="",
              instrument geometry.
     :rtype: Dataset
     """
+    from mantid.api import FileLoaderRegistry
 
     if mantid_args is None:
         mantid_args = {}
 
+    try:
+        FileLoaderRegistry.chooseLoader(filename)
```
SimonHeybrock, line 843: Have you tried of this works if `mantid_alg` is given, i.e., for cases where `Load` would not work?

OwenArnold: I haven't tried that but it's a good point. If the `mantid_alg` is not declared via `DECLARE_FILELOADER_ALGORITHM` I think it will not automatically go through this load path construction mechanism.

python/src/scippneutron/mantid.py
jl-wynen: Better raise `from None` to hide the Mantid error.

jl-wynen (commented):
>  Mantid will also happily load based on run numbers

Sounds useful but might also be risky if it relies on naming conventions that scipp (-based tools) don't necessarily follow.

> or incomplete file paths that are taken to be joined with known Config data search directories

Do we want to use another program's config file? ` I guess, we cannot stop Mantid from using that file but we could prevent incomplete file paths by checking the file name ourselves.

OwenArnold:
> > Mantid will also happily load based on run numbers
> 
> Sounds useful but might also be risky if it relies on naming conventions that scipp (-based tools) don't necessarily follow.
> 
> > or incomplete file paths that are taken to be joined with known Config data search directories
> 
> Do we want to use another program's config file? ` I guess, we cannot stop Mantid from using that file but we could prevent incomplete file paths by checking the file name ourselves.

Good points, but I'm not sure what makes the most sense. On the one hand removing it would make setting up and cross comparing mantid and scippneutron quite  painful for some of our users. On the other hand you are right, we have behaviour that depends on the config of a third-party application.

SimonHeybrock (approved):
Nice solution 👍

jl-wynen (commented):
Looks good as far as I can tell. But one small thing, it would be nice to convert the filename to `str` before passing it to Mantid because Mantid cannot handle `pathlib.Path` which is a bit annoying.

python/src/scippneutron/mantid.py
```diff
@@ -838,14 +839,16 @@ def load(filename="",
     if mantid_args is None:
         mantid_args = {}
 
+    _check_file_path(str(filename), mantid_alg)
```
jl-wynen, line 842: `str(filename)` also needed in l 844

nvaytet:
@jl-wynen can we merge this?

PR #156, opened 2021-09-07
Set DLL search paths

Tom-Willemsen:


tools/build_cpp.py
```diff
@@ -112,6 +112,15 @@ def main(prefix='install', build_dir='build', source_dir='.', caching=False):
...
     print('Compilation took ', end - start, ' seconds')
+    
+    if platform == "win32":
+        import importlib.resources
+        with importlib.resources.path("scipp", "__init__.py") as path:
+            dll_directory = (path.parent.parent / "bin").resolve()
+            # For scipp dlls
+            os.environ["PATH"] += os.pathsep + str(path.parent.resolve())
+            # For tbb dlls
+            os.environ["PATH"] += os.pathsep + str((path.parent.parent.parent / "Library" / "Bin").resolve())
```
SimonHeybrock, line 123: So changing PATH will obviously fix what happens during `conda --build`. But will it affect things on the user side, if someone installs `scippneutron` via `conda`?

PR #161, opened 2021-09-09
Instrument view settings per item for additional components

### LINKED ISSUE #154: Custom components in instrument view, part 2
This is a list of things I think we may need to change/add in the custom component support:

- Current mechanism relies on meta data with given name that contains the position. This generally not the case, e.g., a chopper or monitor is typically a scalar variable with `dtype=DataArray`. The position would be an attribute of this nested array.
- Automatic component size and text size setup is bound to cause trouble.
- User may want display name different from coord name, e.g., to display "sample" instead of "sample_position".

Suggestion:
- `components` should still be a `dict`, I think this is good.
- Keys of `components` should be *display names*
- Values should be dicts, defining position, shape, color, ...

Example:
```python
scn.instrument_view(
    da,
    components={
        'source': {
            'position': 'source_position',
            'color': 'ababab',
            'size': 0.1 * sc.Unit('m'),
            'text_size': 1.0 * sc.Unit('m'),
            'type': 'cube'
        },
        'monitor3': {
            'position': da.attrs['monitor3'].value.coords['position'],
            'color': 'fafafa',
            'type': 'cube'
        },
    })
```

@OwenArnold @nvaytet

OwenArnold:
Fixes #154 

Controlling text size is difficult and has not been implemented:
* Text size in pythreejs only controls the resolution within the Mesh.
* In order to determine proper scaling factors I need to know the coordinates of the entire scene. This is currently known to the model of the Plot3D, but is not exposed.

Note on problem 2 defined above. Could probably handle this as follows if required.
1. Compute the area of the text mesh
1. Compute the area of the scene in the plane of the text mesh (see below)
1. Refer to the default ratio of the two above as `1`
1. User can provide a number which is used to scale the text box by and this in turn will increase/decrease font size.

```
def _scene_dimensions(position):
    x=sc.max(position.fields.x)-sc.min(position.fields.x)
    y=sc.max(position.fields.y)-sc.min(position.fields.y)
    z=sc.max(position.fields.z)-sc.min(position.fields.z)
    return sc.geometry.position(x,y,z)
```


python/src/scippneutron/instrument_view.py
```diff
@@ -32,11 +32,11 @@ def _box(position, display_text, bounding_box, color, **kwargs):
     material = p3.MeshBasicMaterial(color=color)
     mesh = p3.Mesh(geometry=geometry, material=material)
     mesh.position = tuple(position.value)
-    text_position = tuple(position.value + np.array([0, bounding_box[1], 0]))
+    text_position = tuple(position.value + np.array([0, 0.7 * bounding_box[1], 0]))
```
nvaytet, line 35: (unrelated to this line)
One issue I have with the way the text is represented is that is shows up on just one face of the plane geometry, and when you rotate the camera, at some point the text disappears.
I think we should use a `Sprite`, like i did for the text along the outline box axes:
https://github.com/scipp/scipp/blob/main/python/src/scipp/plotting/figure3d.py#L266
Sprites are always facing towards the camera.

python/src/scippneutron/instrument_view.py
```diff
@@ -112,47 +112,55 @@ def _unpack_to_scene(scene, items):
         scene.add(items)
 
 
-def _add_to_scene(position, scene, shape, display_text, bounding_box, **kwargs):
+def _add_to_scene(position, scene, shape, display_text, bounding_box, color, **kwargs):
```
nvaytet, line 115: Do we want to have an option for opacity also?

nvaytet (changes requested):

python/src/scippneutron/instrument_view.py
nvaytet:
```Py
def _text_mesh(position, display_text):
    text = p3.TextTexture(string=display_text, color='black', size=300)
    text_material = p3.SpriteMaterial(map=text, transparent=True)
    size = 1.0
    return p3.Sprite(material=text_material,
                     position=position,
                     scale=[size, size, size])
```

docs/user-guide/instrument-view.ipynb
```diff
@@ -110,7 +110,50 @@
...
+    "Lets add those."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gem_da.coords['sample_position']=sc.vector(value=[0,1,0], unit=sc.units.m)\n",
```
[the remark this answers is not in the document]
OwenArnold: Thanks @nvaytet that's a bug left over from some testing i was doing in the notebook I'll remove.

python/src/scippneutron/instrument_view.py
OwenArnold: Thanks @nvaytet good suggestion

python/src/scippneutron/instrument_view.py
OwenArnold: Hmm, that might have uses should one want to show the vacuum tank of a SANS instrument of a component maybe? Are their other concrete examples you are thinking of?

nvaytet: Yeah i was thinking if you have a component inside a component.
Once thing we might also consider doing it providing a 'wireframe' option?
It's quite easy to do with the `EdgesGeometry`:
```Py
def _cylinder(position, display_text, bounding_box, color, **kwargs):
    geometry = p3.CylinderGeometry(radiusTop=bounding_box[0] / 2,
                                   radiusBottom=bounding_box[0] / 2,
                                   height=bounding_box[1],
                                   radialSegments=12,
                                   heightSegments=12,
                                   openEnded=False,
                                   thetaStart=0,
                                   thetaLength=2.0 * np.pi)
    edges = p3.EdgesGeometry(geometry)
    mesh = p3.LineSegments(geometry=edges,
                           material=p3.LineBasicMaterial(color='#000000'))
    mesh.position = tuple(position.value)
```
Screenshot at 2021-09-09 16-04-54

python/src/scippneutron/instrument_view.py
```diff
@@ -74,11 +74,11 @@ def _disk_chopper(position, display_text, bounding_box, color, **kwargs):
     rotation = _alignment_matrix(to_align=disk_axis, target=beam)
     mesh.setRotationFromMatrix(rotation.flatten())
 
-    text_position = tuple(position.value + np.array([0, bounding_box[1], 0]))
+    text_position = tuple(position.value + np.array([0, 0.7 * bounding_box[1], 0]))
```
nvaytet, line 77: There seems to be some duplication between the different component types (setting the material, the position, creating the text).
Can we move the common functionality to a single function?

python/src/scippneutron/instrument_view.py
OwenArnold: How would we enable this in the API? it makes other features invalid, for example fill color settings. It looks like a nice setting, but I'm not sure how to get this to sit alongside the other input arguments we have.

nvaytet: the color could just be the color of the wireframe?

python/src/scippneutron/instrument_view.py
```diff
@@ -187,6 +214,18 @@ def instrument_view(scipp_obj=None,
...
+    This can be any desired name, it does not have to relate to the input
+    `scipp_obj` naming. The value for each entry is itself a dictionary that
+    provides the display settings and requires an `at`, `size`, `type` and
+    optionally a `color` and `wireframe` flag. `at` should correspond to the
+    meta item where the position to place the component is taken from. `size`
+    is a scipp scalar vector describing the bounding box to use. `type` is one
+    of any known shape types including box, cylinder and disk. The `color`
+    is a hexadecimal color code such as #F00000 to use as fill or line color,
+    which depends on the `wireframe` flag. `wireframe` is a bool that defaults
+    to False, but if set True, provides a wireframe to represent the shape.
```
SimonHeybrock, line 228: - Can you use newlines after each sentence, for a nicer diff?
- Use bullet point list to describe the various keys?
- Should describe `size` better, i.e.,that it should be a scalar variable with length unit.
- `at` sounds a bit odd to my ears, how about `center`?
  - We should also support accepting a (scalar) variable defining the center here. 
- How to get the list of supported shape types? You list a couple, but it is worded as if there were more?
- You say `wireframe` can be set to `True`, but also that it "provides a wireframe to represent the shape", which of the two is it?

python/src/scippneutron/instrument_view.py
```diff
@@ -112,47 +126,59 @@ def _unpack_to_scene(scene, items):
...
             supported_shapes = ", ".join(shapes.keys())
-            raise ValueError(f"Unknown shape: {type} requested for {item}. "
+            raise ValueError(f"Unknown shape: {type} requested for {name}. "
                              f"Allowed values are: {supported_shapes}")
-        component_position = scipp_obj.meta[item]
+        component_position = scipp_obj.meta[at]
+        if not component_position.unit == size.unit:
+            raise ValueError(
+                f"item at {at} has a position with unit {component_position.unit},"
+                f"but size has unit {size.unit}")
```
SimonHeybrock, line 169: I suggest to not fail here, just use:
```python
component_position = sc.to_unit(component_position, positions.unit, copy=False)
size = sc.to_unit(size,  positions.unit, copy=False)
```
Note that we also need the former because:
- `sample_position` may have different unit from (detector) `position`
- `component_position` may be user-provided (see my other comment).

SimonHeybrock (changes requested):

python/src/scippneutron/instrument_view.py
nvaytet: I think `at` is ok, but i also don't mind `center`.

I don't think there is too much of an issue with the wireframe description. Maybe just something like
"`wireframe` is a bool that defaults to `False`. If set to `True`, the returned geometrical shape is a wireframe instead of a shape with opaque faces." ?

python/src/scippneutron/instrument_view.py
OwenArnold: For scipp 0.7, we cannot use `copy` and `to_unit` does not work on vector types. Otherwise I would implement this. I think it would be well worth doing this when we can develop scippneutron against 0.8 scipp.

python/src/scippneutron/instrument_view.py
OwenArnold: > How to get the list of supported shape types? You list a couple, but it is worded as if there were more?
I wrote this in a way such that we would not have to constantly update the docs when new shape types were added. The only way currently to see the supported shape types is to read the exception message for a failed type parameter - I think having a separate function to list the shape types would be useful.

python/src/scippneutron/instrument_view.py
SimonHeybrock: Given that the release is imminent: I was discussing with @jl-wynen last week that we will likely split the scippneutron release:
1. Release changes that have gone in so far + fixes from `scipp-0.8` branch.
2. Shortly after, release again with refactor of `convert` to use `sc.transform_coords`.

So if this PR can wait a couple days you could include it only in the second release.

python/src/scippneutron/instrument_view.py
SimonHeybrock: I am fine with the exception message, but maybe reformulate such that it is clear that this is the full list?

SimonHeybrock: (or maybe I mix this up with the formulation of the docstring, nevermind)

OwenArnold: I'm not sure the best pythonic way to do this, so i've added a `instrument_view_shape_types` function. `instrument_view` shares implementation with this, but it allows the user to query the supported shape types.

python/src/scippneutron/instrument_view.py
OwenArnold: Yes it can wait. Sounds reasonable to me.

SimonHeybrock: Since `main` is now compatible with `scipp-0.8` these changes can be implemented, even without `scippneutron` release.

python/src/scippneutron/instrument_view.py
```diff
@@ -112,47 +126,61 @@ def _unpack_to_scene(scene, items):
...
+        center = settings["center"]
+        color = settings.get("color", "#808080")
+        wireframe = settings.get("wireframe", False)
         if type not in shapes:
             supported_shapes = ", ".join(shapes.keys())
-            raise ValueError(f"Unknown shape: {type} requested for {item}. "
+            raise ValueError(f"Unknown shape: {type} requested for {name}. "
                              f"Allowed values are: {supported_shapes}")
-        component_position = scipp_obj.meta[item]
+        component_position = scipp_obj.meta[center]
```
SimonHeybrock, line 169: `center` should take a variable as an argument, not the name of an coord/attr (or in addition to). Otherwise you will not be able to display, say, monitor positions where you need to access `da.attrs['monitor'].value.meta['position']`.

python/src/scippneutron/instrument_view.py
```diff
@@ -112,47 +126,61 @@ def _unpack_to_scene(scene, items):
...
               **kwargs))
 
 
 def _furthest_component(det_center, scipp_obj, additional):
-    distances = [(key, sc.norm(scipp_obj.meta[key] - det_center).value)
-                 for key in additional.keys()]
+    distances = [(settings["center"],
+                  sc.norm(scipp_obj.meta[settings["center"]] - det_center).value)
+                 for settings in list(additional.values())]
     item, max_displacement = sorted(distances, key=lambda x: x[1])[-1]
```
SimonHeybrock, line 145: Do we actually want to do this? Typically we want to just show the detectors, including the source 160 m away from a small detector in the view will make it almost unusable.

OwenArnold: The only thing this does is set the furthest distance the camera will support. Without this you will not ever be able to see the furthest components when you zoom out. Nevertheless, I'll refactor this slightly as some of the functionality it supports is no longer required.

SimonHeybrock: 👍 (I thought this would affect the default viewport)

python/src/scippneutron/instrument_view.py
nvaytet: I agree that `center` should just be a Variable, like `size`.

python/src/scippneutron/__init__.py
```diff
@@ -24,6 +24,6 @@
 from ._scippneutron import convert
 from ._scippneutron import position, source_position, sample_position, incident_beam, scattered_beam, Ltotal, L1, L2, two_theta
 from .mantid import from_mantid, to_mantid, load, fit
-from .instrument_view import instrument_view
+from .instrument_view import instrument_view, instrument_view_shape_types
```
nvaytet, line 27: I think it's a bit overkill to make this function available to the user.
Could we just list the shapes in the docs, and in the instruement_view docstring?
I know it makes it more prone to having the code and the docs getting out of sync at some point, but I think it's not providing enough functionality to warrant a top-level function?
Once we put the function out there, it's more difficult to remove it later.

docs/user-guide/instrument-view.ipynb
```diff
@@ -110,7 +110,52 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sample_settings = {'center': 'sample_position', 'color': '#000000', 'wireframe': True,\n",
+    "                   'size': sc.vector(value=[0.5, 0.5, 0.5], unit=sc.units.m), 'type': 'box'}\n",
+    "source_settings = {'center': 'source_position', 'color': '#FFC133',\n",
+    "                   'size': sc.vector(value=[1000, 2000, 1000], unit=sc.units.m), 'type': 'cylinder'}\n",
+    "scn.instrument_view(gem_da, pixel_size=0.05, \n",
+    "                    components={'sample': sample_settings, 'source': source_settings})"
```
nvaytet, line 158: Could you also add a couple of tests in the python tests (either in the `test_neutron.py` or move instrument view tests to a new file?) that test the new features (e.g. all the shapes, `wireframe=True` etc...).
As we probably don't want to test all the combinations in the docs...

nvaytet (changes requested):

python/src/scippneutron/instrument_view.py
```diff
@@ -151,13 +151,20 @@ def _instrument_view_shape_types():
     return {"box": _box, "cylinder": _cylinder, "disk": _disk_chopper}
 
 
+def _as_vector(center):
+    if center.dtype == sc.dtype.vector_3_float64:
+        return center
+    else:
+        return sc.geometry.position(x=center, y=center, z=center)
```
SimonHeybrock, line 158: `center` is a bit misleading as a variable name? Isn't it `size`?

docs/user-guide/instrument-view.ipynb
```diff
@@ -110,7 +110,52 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sample_settings = {'center': 'sample_position', 'color': '#000000', 'wireframe': True,\n",
+    "                   'size': sc.vector(value=[0.5, 0.5, 0.5], unit=sc.units.m), 'type': 'box'}\n",
+    "source_settings = {'center': 'source_position', 'color': '#FFC133',\n",
+    "                   'size': sc.vector(value=[1000, 2000, 1000], unit=sc.units.m), 'type': 'cylinder'}\n",
```
nvaytet, line 156: 2km seems to be a very large size?

python/src/scippneutron/instrument_view.py
```diff
@@ -9,34 +9,52 @@
 
 
 def _text_mesh(position, display_text, x_width, y_width):
-    text_geometry = p3.PlaneGeometry(width=x_width,
-                                     height=y_width,
-                                     widthSegments=2,
-                                     heightSegments=2)
-
     text = p3.TextTexture(string=display_text, color='black', size=20)
```
nvaytet, line 12: The resolution on the text is quite bad, I would suggest to use `size=300`, as we do in the scipp 3d plotting.
Screenshot at 2021-09-21 11-25-42

python/src/scippneutron/instrument_view.py
```diff
@@ -9,34 +9,52 @@
 
 
 def _text_mesh(position, display_text, x_width, y_width):
```
nvaytet, line 11: I know i'm being annoying but it's not really making a `mesh` anymore, it's making a `Sprite`.
Rename to `_text_sprite` and `_create_text_sprite`? Or just `_create_text`?

python/src/scippneutron/instrument_view.py
```diff
@@ -9,34 +9,52 @@
...
+    size = 1.0
+    return p3.Sprite(material=text_material,
+                     position=position,
+                     scale=[size, size, size])
+
+
+def _create_text_mesh(position, bounding_box, display_text):
+    # Position offset in y
+    text_position = tuple(position.value + np.array([0, 0.8 * bounding_box[1], 0]))
+    text_mesh = _text_mesh(text_position,
```
nvaytet, line 23: Is `_text_mesh` being called anywhere else? If not, maybe we can drop the extra function and put everything in a single function named `_create_text`?

python/src/scippneutron/instrument_view.py
```diff
@@ -9,34 +9,52 @@
...
+                           display_text=display_text,
+                           x_width=max(bounding_box),
```
nvaytet, line 25: Also, I don't see `x_width` and `y_width` being used inside `_text_mesh`? (this is not picked up by `flake8`?)

python/src/scippneutron/instrument_view.py
```diff
@@ -112,47 +126,68 @@ def _unpack_to_scene(scene, items):
...
+                      display_text=name,
+                      bounding_box=tuple(size.values),
+                      color=color,
+                      wireframe=wireframe,
                       det_center=det_center)
     # Reset camera
     camera = _get_camera(scene)
     if camera:
-        camera.far = furthest_distance * 5.0
+        furthest_distance = _furthest_component(det_center, scipp_obj, components)
```
nvaytet, line 189: For some reason, this seems to not quite be working. In the GEM example in the notebook above, I can't see the source. Only the sample in the middle of the detectors.

As an aside, when we rescale this, do we also want to rescale the red/green/blue axes? I think we currently don't have easy control over this in the plot. Maybe something for later?

python/tests/test_instrument_view.py
```diff
@@ -0,0 +1,94 @@
...
+    scn.instrument_view(d["a"],
+                        components={'sample': _make_component_settings(wireframe=True)})
+
+
+def test_neutron_instrument_view_components_with_invalid_type():
+    d = make_dataset_with_beamline()
+    with pytest.raises(ValueError):
+        scn.instrument_view(
+            d["a"],
+            components={'sample': _make_component_settings(type='trefoil_knot')})
```
nvaytet, line 82: Maybe also test the `cylinder` and the `disk` to make sure these continue to work?

nvaytet (changes requested):

python/src/scippneutron/instrument_view.py
nvaytet: hmm maybe the fact that I can't see it is due to the gigantic size set in the notebook? (1x2x1 km)

nvaytet (approved):

nvaytet:
Thanks for all the changes. Looking good now :+1:

python/src/scippneutron/instrument_view.py
SimonHeybrock: @OwenArnold There is this comment left to address.

python/src/scippneutron/instrument_view.py
```diff
@@ -222,7 +221,7 @@ def instrument_view(scipp_obj=None,
     The value for each entry is itself a dictionary that provides the display
     settings and requires:
 
-    * `center` - name of meta item where the position to place the component is
+    * `center` - Variable (0D vector) position to place the component is
     taken from.
```
SimonHeybrock: This sentence does not parse anymore.

SimonHeybrock (approved):
Nice work, looks good. 👍

Issue #138, opened 2021-08-17
Add option to return intermediate detector data structures from `load_nexus` 


Tom-Willemsen:
It may be useful for some operations to load "intermediate" detector data structures such as the event index from `load_nexus`. Probably make the binning/mapping to pixel step optional.

SimonHeybrock:
Basically, this should return data as stored on disk in the `NXevent_data` groups. Note that these are already loaded, but there is no option to return this.

OwenArnold:
Would be nice if this work (or subsequent) could demo how these options could be used. It might be relevant for event filtering for example.

PR #163, opened 2021-09-10
Load intermediate detector structures

Tom-Willemsen:
Add option to load intermediate detector and event data structures directly as they appear in the nexus file, with no binning/preprocessing.

Closes https://github.com/scipp/scippneutron/issues/138
Closes https://github.com/scipp/scippneutron/issues/167


python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -160,8 +172,13 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
...
             f"{__name__}('my_file.nxs', '/entry_2')")
-    loaded_data = load_detector_data(groups[nx_event_data], groups[nx_detector],
-                                     nexus_file, nexus, quiet)
+
+    if raw_detector_data:
+        loaded_data = load_raw_detector_structures(groups[nx_event_data],
+                                                   groups[nx_detector], nexus)
+    else:
+        loaded_data = load_detector_data(groups[nx_event_data], groups[nx_detector],
+                                         nexus_file, nexus, quiet)
```
SimonHeybrock, line 181: I had expected that both of these use the same underlying code to load `NXevent_data`. As it is now there is a lot of duplication with `load_raw_detector_structure`, I think. Conceptually I thought this would look something like this:

```python
event_data = load_event_data(...)
if group_by_pixel:
  loaded_data = sc.bin(event_data, groups=[detector_id])
else:
  events = sc.DataArray(data=sc.ones(...), coords={'tof':event_data['tof'], 'pulse_time':event_data['pulse_time']})
  loaded_data = sc.bins(begin=event_data['event_index'], end=..., dim='pulse', data=events)
```
where the first three lines is what the existing implementation does.

Needs some extra thought w.r.t. how multiple banks get handled, but if all banks have the same pulses it can just be concatenated.

SimonHeybrock (commented):
Some questions below. Happy to discuss in a call to clarify a bit more.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -459,3 +461,36 @@ def _load_data_from_each_nx_detector(detector_groups: List[Group], file_root: h5
...
+        for structure in [
+                "event_id", "event_index", "event_time_bins", "event_time_offset",
+                "event_time_zero"
+        ]:
+            try:
+                data[structure] = nexus.load_dataset(group.group, structure,
+                                                     [structure])
+            except MissingDataset:
+                pass
+        loaded_data[name] = sc.scalar(value=data)
```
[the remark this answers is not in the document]
Tom-Willemsen: For symmetry with `load_data` without the `raw_detector_data` flag. I thought it would be confusing to return a `Dict` in one case and a `Dataset` in the other? As it is I think it is already slightly confusing as it can sometimes return a `DataArray` instead of a `Dataset`...

python/src/scippneutron/file_loading/load_nexus.py
Tom-Willemsen: Hmm, perhaps we had different ideas of what this ticket should accomplish? I thought we were trying to load the detector structures exactly as they appear in the nexus file, with no manipulations done at all. This way people can apply whatever specific binning steps they need manually, using the low-level structures if needed.

SimonHeybrock: > Hmm, perhaps we had different ideas of what this ticket should accomplish? I thought we were trying to load the detector structures exactly as they appear in the nexus file, with no manipulations done at all.

No, not actually different ideas: The structure of `NXevent_data` is essentially the same structure scipp uses for binned data, so not actual manipulation is necessary, just putting things into the right place. I can explain in a call tomorrow if you like.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -362,12 +375,25 @@ def get_detector_id(data: DetectorData):
...
+        if not raw:
+            # Events in the NeXus file are effectively binned by pulse
+            # (because they are recorded chronologically)
+            # but for reduction it is more useful to bin by detector id
+            return sc.bin(data.events, groups=[data.detector_ids], edges=[_tof_edges])
+        else:
+            # If loading "raw" data, leave binned by pulse.
+            begins = data.event_index
+            ends = sc.array(dims=[_event_dimension],
+                            values=np.append(data.event_index.values[1:],
```
OwenArnold, line 388: Why not the zeroth element?

python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -114,30 +117,38 @@ def _load_start_and_end_time(entry_group: Group, data: ScippData, nexus: LoadFro
...
+               quiet=True,
+               raw_detector_data: bool = False) -> Optional[ScippData]:
     """
     Load a NeXus file and return required information.
 
     :param data_file: path of NeXus file containing data to load
     :param root: path of group in file, only load data from the subtree of
       this group
     :param quiet: if False prints some details of what is being loaded
+    :param raw_detector_data: if True, loads detector and event data structures as they
```
OwenArnold, line 129: What happens to the other information (meta data) in this case?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -269,14 +262,32 @@ def _load_event_group(group: Group, file_root: h5py.File, nexus: LoadFromNexus,
...
-    pixel_positions_found, _ = nexus.dataset_in_group(detector_group, "x_pixel_offset")
-    # Checking for positions here is needed because, in principle, the standard
-    # allows not to always have them. ESS files should however always have
-    # them.
-    if pixel_positions_found:
-        detector_data.pixel_positions = _load_pixel_positions(
-            detector_group, detector_data.detector_ids.shape[0], file_root, nexus)
+    if raw:
+
+        detector_data.event_index = sc.Variable(dims=[_event_dimension],
```
OwenArnold, line 267:
```suggestion
        detector_data.event_index = sc.array(dims=[_event_dimension],
```

python/src/scippneutron/file_loading/load_nexus.py
Tom-Willemsen: The metadata is loaded as usual.

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: If the event index is
```
[0, 5, 10]
```

We want the first bin to be events `0-5`, i.e. the end of the first bin is the second element of the event index.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -269,14 +262,32 @@ def _load_event_group(group: Group, file_root: h5py.File, nexus: LoadFromNexus,
...
+
+        detector_group = group.parent
+        pixel_positions_found, _ = nexus.dataset_in_group(detector_group,
+                                                          "x_pixel_offset")
+        # Checking for positions here is needed because, in principle, the standard
+        # allows not to always have them. ESS files should however always have
+        # them.
+        if pixel_positions_found:
+            detector_data.pixel_positions = _load_pixel_positions(
+                detector_group, detector_data.detector_ids.shape[0], file_root, nexus)
```
SimonHeybrock, line 290: Even if we do not bin into pixels, shouldn't we still load the positions?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -362,12 +375,25 @@ def get_detector_id(data: DetectorData):
...
+            # Events in the NeXus file are effectively binned by pulse
+            # (because they are recorded chronologically)
+            # but for reduction it is more useful to bin by detector id
+            return sc.bin(data.events, groups=[data.detector_ids], edges=[_tof_edges])
+        else:
+            # If loading "raw" data, leave binned by pulse.
+            begins = data.event_index
+            ends = sc.array(dims=[_event_dimension],
+                            values=np.append(data.event_index.values[1:],
+                                             data.event_index.values[-1]))
```
SimonHeybrock, line 389: Isn't this dropping the events from the last pulse? I think we have to append the total number of events as the final end index.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -362,12 +375,25 @@ def get_detector_id(data: DetectorData):
...
-    # but for reduction it is more useful to bin by detector id
-    events = sc.bin(detector_data.events,
-                    groups=[detector_data.detector_ids],
-                    edges=[_tof_edges])
+    def _bin_events(data):
+        if not raw:
+            # Events in the NeXus file are effectively binned by pulse
+            # (because they are recorded chronologically)
+            # but for reduction it is more useful to bin by detector id
+            return sc.bin(data.events, groups=[data.detector_ids], edges=[_tof_edges])
```
SimonHeybrock, line 383: Why is it now `data.detector_ids`?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -376,16 +402,14 @@ def get_detector_id(data: DetectorData):
...
 
-        new_events = sc.bin(detector_data.events,
-                            groups=[detector_data.detector_ids],
-                            edges=[_tof_edges])
+        new_events = _bin_events(detector_data)
 
         if pixel_positions_loaded:
             new_events.coords['position'] = detector_data.pixel_positions
 
         events = sc.concatenate(events, new_events, dim=_detector_dimension)
```
SimonHeybrock, line 410: If `raw` then `dim` should not be `detector_id` (or similar) but `bank` (or similar).

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -376,16 +402,14 @@ def get_detector_id(data: DetectorData):
...
 
-    return events
+    return sc.DataArray(data=events) if raw else events
```
SimonHeybrock, line 412: Why the wrapping in a data array?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -376,16 +402,14 @@ def get_detector_id(data: DetectorData):
     while event_data:
         detector_data = event_data.pop(0)
 
-        new_events = sc.bin(detector_data.events,
-                            groups=[detector_data.detector_ids],
-                            edges=[_tof_edges])
+        new_events = _bin_events(detector_data)
 
         if pixel_positions_loaded:
             new_events.coords['position'] = detector_data.pixel_positions
```
SimonHeybrock, line 408: If `raw` then we should add the pulse times here instead. Note that we should *not* map pulse times into bins in the "raw" load (it wouldn't be raw any more). This happens in `_load_event_group`.

python/src/scippneutron/file_loading/load_nexus.py
```diff
@@ -211,11 +225,16 @@ def _load_nexus_json(
     streams = None
     if get_start_info:
         streams = get_streams_info(loaded_json)
-    return _load_data(loaded_json, None, LoadFromJson(loaded_json), True), streams
+    return _load_data(loaded_json,
+                      None,
+                      LoadFromJson(loaded_json),
+                      True,
+                      raw_detector_data=raw_detector_data), streams
```
SimonHeybrock, line 232: Would a better name for this parameter be `bin_by_pixel` or something like this?

SimonHeybrock (changes requested):

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: This is already being done as one of the manipulations in `_load_event_group` (see https://github.com/scipp/scippneutron/blob/main/python/src/scippneutron/file_loading/_detector_data.py#L219), unless I have misunderstood your comment?

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: Because this is a nested function which takes `data` as it's argument, I felt this was less ambiguous than shadowing the name from the outer scope?

SimonHeybrock: Sorry, I misread, you are right.

python/src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: But if that is the case, shouldn't you have:
```python
begins = data.event_index[:-1]
ends = data.event_index[1:]
```
?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -160,18 +162,24 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus,
...
-        raise BadSource(f"Event index in NXEvent at {group.path}/event_index was not"
-                        f"ordered. The index must be ordered to load pulse times.")
+        if any(_diffs < 0):
+            raise BadSource(
+                f"Event index in NXEvent at {group.path}/event_index was not"
+                f"ordered. The index must be ordered to load pulse times.")
+
+        _raw_pulse_times_np = np.repeat(_raw_pulse_times.values, _diffs)
+    else:
+        _raw_pulse_times_np = _raw_pulse_times.values
```
SimonHeybrock, line 175: Why do we need to convert to numpy just to convert back to scipp?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -366,23 +380,30 @@ def get_detector_id(data: DetectorData):
...
+            begins = data.event_index[_pulse_dimension, :-1]
+            ends = data.event_index[_pulse_dimension, 1:]
+
+            binned = sc.bins(data=data.events,
+                             dim=_event_dimension,
+                             begin=begins,
+                             end=ends)
+
+            da = sc.DataArray(data=binned,
+                              coords={"pulse_time": data.pulse_times["pulse", :-1]})
```
SimonHeybrock, line 400: Why is the last pulse dropped? There should be the same number of bins as pulse times, unless the index setup is wrong. You said earlier that `data.event_index` already had been extended by one element, so `begin` and `end` above should have the same length.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -366,23 +380,30 @@ def get_detector_id(data: DetectorData):
...
+            da.events.coords['pulse_time'] = sc.empty(sizes=da.events.sizes,
+                                                      dtype=data.pulse_times.dtype,
+                                                      unit=data.pulse_times.unit)
+            da.bins.coords['pulse_time'][...] = da.coords['pulse_time']
+            del da.coords["pulse_time"]
```
SimonHeybrock, line 405: Copy and paste error? We don't want to map the pulse times to events, when just loading the raw detector structure.

python/tests/test_load_nexus.py
```diff
@@ -168,47 +170,13 @@ def test_loads_pulse_times_with_combinations_of_offset_and_units(
...
     assert sc.identical(
-        loaded_data.values[0].attrs['pulse_time'],
+        loaded_data.values[0].coords['pulse_time'],
         sc.array(dims=["event"],
                  values=[_time],
                  unit=sc.units.s,
                  dtype=sc.dtype.datetime64))
 
 
-def test_does_not_load_events_if_time_zero_unit_not_convertible_to_s(
```
SimonHeybrock, line 178: Why is this removed?

SimonHeybrock (changes requested):

python/tests/test_load_nexus.py
Tom-Willemsen: This test mixed structures that did and didn't contain pulse times, which then caused `concatenate` to fail with mismatched dimensions if you try to concatenate those structures together.

SimonHeybrock: But this seems unrelated to the current change. Are you saying the test was passing but for the wrong reason?

Tom-Willemsen: I think the difference is probably that it is now a `coord`, where it was previously an `attr`? Not completely sure, I can look at this a bit further

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: I'm not sure I understand exactly what is meant by this - we don't currently load `bank` as a dim/attribute and it's not clear to me exactly what this refers to. Should `bank` be the name of the nexus group from which the events were loaded (something like `detector_1_events`), or is there some other property that I've missed here?

SimonHeybrock: If we load multiple `NXevent_data` groups you end up callong `concatenate` here, and thus you need a dimension to concatenate along. I don't think there is a named equivalent to this in the file, since the file just uses multiple groups, rather than an axis/dimension, so the naming is slightly arbitrary.

We don't necessarily need a coord for this dim, but adding the Nexus group name as you suggest (`detector_1_events`) could be an option. But maybe just don't for now, for simplicity.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -157,20 +165,25 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus,
...
 
     try:
-        pulse_times = sc.to_unit(sc.array(dims=[_event_dimension],
-                                          values=np.repeat(_raw_pulse_times.values,
-                                                           _diffs),
-                                          unit=_raw_pulse_times.unit,
-                                          dtype=sc.dtype.int64),
-                                 sc.units.s,
-                                 copy=False)
+        pulse_times = sc.to_unit(_raw_pulse_times_sc, sc.units.s, copy=False)
```
SimonHeybrock, line 186: Wondering why the pulse-times get converted to seconds? While this was not introduced in this PR this seems wrong, since we lose a lot of very significant and important digits. @OwenArnold I think you reviewed the original PR that added this, can you comment?

Tom-Willemsen: I also think this is wrong, was just about to create an issue for it and assign myself.

Tom-Willemsen: https://github.com/scipp/scippneutron/issues/167

OwenArnold: Agree this should not be converted to seconds.

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -153,8 +153,15 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus, number_of_event_ids: i
     _raw_pulse_times = nexus.load_dataset(
         group.group,
         time_zero_group,
-        dimensions=[_event_dimension if times_per_event else _pulse_dimension],
-        dtype=sc.dtype.int64)
+        dimensions=[_event_dimension if times_per_event else _pulse_dimension])
+
+    try:
+        pulse_times = sc.to_unit(_raw_pulse_times, sc.units.ns, copy=False)
```
SimonHeybrock, line 159: Why are we not simply leaving it in the original unit, as found in the file?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -173,24 +180,23 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus, number_of_event_ids: i
...
-    except sc.UnitError:
-        raise BadSource(f"Could not load pulse times: units attribute "
-                        f"'{_raw_pulse_times.unit}' in NXEvent at "
-                        f"{group.path}/{time_zero_group} is not convertible"
-                        f" to seconds.")
+        # Need to convert the values which were loaded as float64 into int64 to be able
+        # to do datetime arithmetic. This needs to be done after conversion to ns to
+        # avoid unnecessary loss of accuracy.
+        pulse_times = sc.array(
+            dims=[_event_dimension if times_per_event else _pulse_dimension],
```
SimonHeybrock, line 193: Aren't we in the `else:` branch of `if times_per_event` (unless I misread the diff)?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -173,24 +180,23 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus, number_of_event_ids: i
                 f"Event index in NXEvent at {group.path}/event_index was not"
                 f"ordered. The index must be ordered to load pulse times.")
 
-        _raw_pulse_times_np = np.repeat(_raw_pulse_times.values, _diffs)
-        _raw_pulse_times_sc = sc.array(
+        pulse_times = sc.array(
             dims=[_event_dimension if times_per_event else _pulse_dimension],
```
SimonHeybrock, line 184: Aren't we in `if times_per_event`?

python/src/scippneutron/file_loading/_detector_data.py
```diff
@@ -173,24 +180,23 @@ def _load_pulse_times(group: Group, nexus: LoadFromNexus, number_of_event_ids: i
...
-                        f"{group.path}/{time_zero_group} is not convertible"
-                        f" to seconds.")
+        # Need to convert the values which were loaded as float64 into int64 to be able
+        # to do datetime arithmetic. This needs to be done after conversion to ns to
+        # avoid unnecessary loss of accuracy.
+        pulse_times = sc.array(
+            dims=[_event_dimension if times_per_event else _pulse_dimension],
+            values=pulse_times.values,
+            unit=sc.units.ns,
+            dtype=sc.dtype.int64)
```
SimonHeybrock, line 196: Not sure I understand what is done here. Why are we extracting values from a variable, just to create a new one?

python/tests/test_load_nexus.py
```diff
@@ -140,22 +140,20 @@ def test_loads_pulse_times_from_single_event_with_different_units(
...
-        assert sc.identical(
-            loaded_data.values[event].coords['pulse_time'],
-            sc.array(dims=["event"],
-                     values=[_time],
-                     unit=sc.units.s,
-                     dtype=sc.dtype.datetime64))
+        # Allow 1ns difference for rounding errors between different routes
+        assert all(
+            np.abs(loaded_data.values[event].coords['pulse_time'].values -
+                   _time) <= np.array(1).astype("timedelta64[ns]"))
```
SimonHeybrock: Use `isclose` or `allclose`?

python/tests/test_load_nexus.py
```diff
@@ -178,14 +176,11 @@ def test_loads_pulse_times_with_combinations_of_offset_and_units(
...
 
-    assert sc.identical(
-        loaded_data.values[0].coords['pulse_time'],
-        sc.array(dims=["event"],
-                 values=[_time],
-                 unit=sc.units.s,
-                 dtype=sc.dtype.datetime64))
+    # Allow 1ns difference for rounding errors between different routes
+    assert np.abs(loaded_data.values[0].coords['pulse_time'].values[0] -
+                  _time) <= np.array(1).astype("timedelta64[ns]")
```
SimonHeybrock: Use `sc.allclose`?

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: Yes, exactly. The example files I have all use `s` in the nexus file,

python/src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: To convert to `int64` so that we can add it to a `datetime64`. We don't want to load it initially as `int64` as we wouldn't have done the unit conversion yet and so may lose precision if the unit in the file is e.g. seconds

python/tests/test_load_nexus.py
Tom-Willemsen: `isclose`/`allclose` don't support datetime64.

python/tests/test_load_nexus.py
Tom-Willemsen: `isclose`/`allclose` don't support datetime64.

python/src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Use `astype`?

python/tests/test_load_nexus.py
SimonHeybrock: Do we even have rounding issues then? Have you tried `sc.identical`?

Tom-Willemsen: Yes, I tried `sc.identical`. It fails on some tests because the numbers are not quite equal, hence the 1ns tolerance that I added in these tests.

SimonHeybrock (approved):

PR #164, opened 2021-09-14
Update to scipp-0.8

SimonHeybrock:


conda/meta.yaml
```diff
@@ -26,6 +26,7 @@ test:
     - neutron
   requires:
     - h5py
+    - ipykernel = 6.3.1  # see https://github.com/ipython/ipykernel/issues/771
     - ipywidgets
     - matplotlib-base
     - psutil
```
nvaytet, line 32: Probably need to pin `scipp` to `0.8` in the `conda_build_config.yaml` file.

.azure-pipelines/main.yml
```diff
@@ -43,7 +40,7 @@ extends:
         conda_env: 'scippneutron-developer.yml'
       osx:
         py_versions: ['3.7']
-        conda_env: 'scippneutron-developer.yml'
+        conda_env: 'scippneutron-developer-osx.yml'
```
nvaytet, line 43: Having all these different env files is becoming a nightmare to maintain. Is there really no alternative to having a new file which is 98% identical to the other env files? Aren't there tags you can add in env file if you want something install only on certain platforms? I thought we did this with some of the data streaming packages which don't install on windows.

nvaytet: See in the conda/meta.yaml file

SimonHeybrock: > Having all these different env files is becoming a nightmare to maintain.

Couldn't agree more. Added as item in scipp/scipp#2163.

OwenArnold: > See in the conda/meta.yaml file

I mainly made this change to ensure that we could complete the CI, i.e. scippneutron worked with 0.8. These only work with the recipe files as far as I can determine. https://conda-devenv.readthedocs.io/en/latest/usage.html is the only other solution i am aware of.

SimonHeybrock: See note I just added in https://github.com/scipp/scipp/issues/2163#issuecomment-919712295, might `conda debug` be a solution for creating the envs from `meta.yml`?

OwenArnold: > might conda debug be a solution for creating the envs from meta.yml?
It could work, but seems that you need to source the environment script from within a conda working directory, that is to say, it doesn't look like the most convenient thing to use as a developer, but I do like the fact that this might result in us being able to discard our environment files all-together.

SimonHeybrock: Yes, you need to have a base env with (I presume) `conda-build` installed. But then something like this may work:

```sh
eval "$(conda debug -c conda-forge -a .)"
```

I suppose there could be a helper script shipped with the repo that does this.

docs/tutorials/exploring-data.ipynb
```diff
@@ -401,7 +401,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.10"
+   "version": "3.8.10"
   }
  },
  "nbformat": 4,
```
jl-wynen, line 407: `instrument_view` seems to be broken. Try this notebook. It gives me `ValueError: Out of range float values are not JSON compliant`

SimonHeybrock: Are you using an old conda env? I fixed this a couple days ago by pinning `ipykernel`.

docs/tutorials/powder-diffraction.ipynb
```diff
@@ -331,12 +348,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dspacing_bins = sc.Variable(\n",
-    "    ['dspacing'],\n",
-    "    values=np.arange(0.3, 2.0, 0.001),\n",
-    "    unit=sc.units.angstrom)\n",
-    "hist = sc.Dataset({'sample':sc.histogram(dspacing_sample, dspacing_bins),\n",
-    "                   'vanadium':sc.histogram(dspacing_vanadium, dspacing_bins)})\n",
+    "dspacing_bins = sc.arange(dim='dspacing', start=0.3, stop=2.0, step=0.001, unit=sc.units.angstrom)\n",
```
jl-wynen, line 351: I think `arange` falls into the category of position-only arguments because the meaning of the numerical arguments changes depending on how many there are. (Not including `unit`, of course.)

I mention this because someone had posted a link to position-only arguments at some point and it sounded like there was interest in them,

docs/tutorials/powder-diffraction.ipynb
```diff
@@ -396,15 +410,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "two_theta = sc.Variable(['two_theta'],\n",
-    "                    unit=sc.units.rad,\n",
-    "                    values=np.linspace(0.0, np.pi, num=16))\n",
+    "two_theta = sc.linspace(dim='two_theta', unit='rad', start=0.0, stop=np.pi, num=16)\n",
```
jl-wynen, line 413: Same as `arange` above. And there are more cases in other notebooks.

docs/tutorials/powder-diffraction.ipynb
SimonHeybrock: Only `start` appears to be position-only: `arange(start, /, stop=None, step=1, *, dtype=None, device=None)` in https://data-apis.org/array-api/latest/API_specification/creation_functions.html.

docs/tutorials/powder-diffraction.ipynb
SimonHeybrock: `linspace(start, stop, /, num, *, dtype=None, device=None, endpoint=True)` from Array API standard.

But we should also keep in mind that scipp requires an extra `dim` argument, so we cannot necessarily adopt this 1:1

docs/about/release-notes.rst
```diff
@@ -3,9 +3,12 @@
 Release Notes
 =============
 
-Since v0.2.0
+Since v0.3.0
 ------------
 
+v0.3.0 (September 2021)
+------------------
```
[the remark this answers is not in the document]
SimonHeybrock:
```suggestion
-----------------------
```

docs/tutorials/exploring-data.ipynb
jl-wynen: That fixed it

docs/tutorials/powder-diffraction.ipynb
nvaytet: `start` is used as `stop` if it's the only one that is specified

jl-wynen: Ok. We need to decide how to handle those in our `arange` considering that we have the additional dimension arg. I personally, prefer not naming arguments of arange. But we can still make the docs reflect the current status of the API. So I am fine with leaving it.

docs/environments/scippneutron-latest.yml
```diff
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - ess-dmsc::ess-streaming-data-types
   - h5py
+  - ipykernel = 6.3.1  # see https://github.com/ipython/ipykernel/issues/771
```
nvaytet, line 9: I actually don't think these files are used anywhere? It's probably safe to remove them?

SimonHeybrock: On the contrary, this is what users will use to install. See https://scipp.github.io/scippneutron/getting-started/installation.html#with-the-provided-environment-file.

nvaytet: ah, that's what they are used for! Thanks!

nvaytet: But once again we risk those being out of sync with https://github.com/scipp/scippneutron/blob/main/scippneutron-developer.yml and https://github.com/scipp/scippneutron/blob/main/scippneutron-developer-no-mantid.yml

nvaytet: And these won't work for windows, etc...

SimonHeybrock: > But once again we risk those being out of sync with https://github.com/scipp/scippneutron/blob/main/scippneutron-developer.yml and https://github.com/scipp/scippneutron/blob/main/scippneutron-developer-no-mantid.yml

Not really, because the purpose is different: The user-envs do not include dev dependencies. And valid versions for actual dependencies should be setup correctly in `meta.yml` so we do *not* have to re-specify this here. The user env files are basically just "get me scippneutron and useful basics (jupyterlab, ...)". The fact that they contain `ipykernel` now is because of the lack of proper pinning in our package, I think (or maybe just redundant, but I wanted to be on the safe side)?

nvaytet (approved):

PR #169, opened 2021-09-22
Plotting dependencies are optional

OwenArnold:
scippneutron/__init__.py imports instrument_view with plotting dependencies considered optional

```
ImportError while importing test module '/write_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniconda3/envs/ess-developer-latest/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/reflectometry/write_test.py:15: in <module>
    from ess.reflectometry import write, data, orso
src/ess/reflectometry/data.py:13: in <module>
    import scippneutron as scn
../../miniconda3/envs/ess-developer-latest/lib/python3.7/scippneutron/__init__.py:27: in <module>
    from .instrument_view import instrument_view
../../miniconda3/envs/ess-developer-latest/lib/python3.7/scippneutron/instrument_view.py:6: in <module>
    import pythreejs as p3
E   ModuleNotFoundError: No module named 'pythreejs'
```

### DIFF
--- a/python/src/scippneutron/instrument_view.py
+++ b/python/src/scippneutron/instrument_view.py
@@ -3,9 +3,13 @@
 # @author Neil Vaytet and Owen Arnold
 
 import numpy as np
-import pythreejs as p3
 import scipp as sc
 from scipy.spatial.transform import Rotation as Rot
+try:
+    import pythreejs as p3
+except ImportError as ex:
+    p3 = None
+    _pythreejs_import_error = ex
 
 
 def _create_text_sprite(position, bounding_box, display_text):
@@ -235,6 +239,8 @@ def instrument_view(scipp_obj=None,
     the returned geometrical shape is a wireframe instead of a shape with
     opaque faces
     """
+    if not p3:
+        raise _pythreejs_import_error
 
     from scipp.plotting import plot
     from scipp.plotting.objects import PlotDict


python/src/scippneutron/instrument_view.py
```diff
@@ -3,12 +3,12 @@
 # @author Neil Vaytet and Owen Arnold
 
 import numpy as np
-import pythreejs as p3
 import scipp as sc
 from scipy.spatial.transform import Rotation as Rot
 
 
 def _text_mesh(position, display_text, x_width, y_width):
+    import pythreejs as p3
```
nvaytet, line 11: Is there a nice way we could do this without having to import in 5 different places?

OwenArnold: The other options I'm aware of:

* try-catch at the top of the module along side the other imports. This would allow you to import once, However this will give misleading exceptions at the point of use, so I'm not in favour of this.
* refactor how the instrument_view is imported upstream. We see the error when we `import scippneutron` because we have `instrument_view` in `__init__.py` but maybe we could look to remove it from the package construction?

nvaytet: Maybe a try and except that sets `p3 = None` in the `except` block, and then a check inside the `instrument_view` function that raises a sensible error message if `p3 is None`?

OwenArnold: What you suggest sounds like a modified version of my first suggestion. I'll draft out what this looks like and we can re-review.

nvaytet: > What you suggest sounds like a modified version of my first suggestion.

yes that's exactly what it is ;-)

nvaytet (approved):

OwenArnold:
@nvaytet Maybe we should handle all our optional dependencies in a similar way? We have the same problem with ipywidgets in scipp for example.

PR #174, opened 2021-09-27
Use pooch to get tutorial data

SimonHeybrock:
Use pooch instead of custom handling of sample data. They key advantage is that it makes notebooks portable, in particular also to our JupyterHub, and to users computers, without requiring manual file downloads.

For later:
- Use same approach for other files used in docs.
- Consider adding generated data, rather than just downloaded. Maybe more relevant for `scipp` itself.
- Are we happy with the naming?

We also need to figure out what to do with versions. The current setup will brake once we make a release. Do we add symlinks for new releases? Or do we version `scippneutron.data` independently from `scippneutron`?

### DIFF
--- a/docs/tutorials/exploring-data.ipynb
+++ b/docs/tutorials/exploring-data.ipynb
@@ -34,7 +34,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We  start by loading some data (download [here](https://github.com/ess-dmsc-dram/loki_tube_scripts/raw/master/test/test_data/LARMOR00049338.nxs)), in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
+    "We  start by loading some data, in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
    ]
   },
   {
@@ -43,7 +43,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "data = scn.load(filename='LARMOR00049338')"
+    "data = scn.data.tutorial_dense_data()"
    ]
   },
   {
--- a/docs/tutorials/working-with-masks.ipynb
+++ b/docs/tutorials/working-with-masks.ipynb
@@ -36,7 +36,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We  start by loading some data (download [here](https://github.com/ess-dmsc-dram/loki_tube_scripts/raw/master/test/test_data/LARMOR00049338.nxs)), in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
+    "We  start by loading some data, in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
    ]
   },
   {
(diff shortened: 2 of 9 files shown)


SimonHeybrock:
@nvaytet Should we just version `scippneutron.data` independently, i.e., just define a version in `data/__init__.py` that gets updated only if we need to replace a file?

docs/tutorials/exploring-data.ipynb
```diff
@@ -43,7 +43,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "data = scn.load(filename='LARMOR00049338')"
+    "data = scn.data.tutorial_dense_data()"
```
nvaytet, line 46: Can we make sub-groups in the `.data`, e.g. `scn.data.tutorials.dense_data()`?
Or something like a dict that would let us list the keys to see what's available?

python/src/scippneutron/data/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+_version = '1'
+
+__all__ = ['tutorial_dense_data', 'tutorial_event_data']
+
+
+def _make_pooch():
+    import pooch
+    return pooch.create(
+        path=pooch.os_cache('scippneutron'),
+        base_url='https://public.esss.dk/groups/scipp/scippneutron/{version}/',
```
nvaytet, line 15: I'm not sure exactly why we need the version here. Does `pooch` require versions to work?
Or is it because you think the files might change in the future and you want to maintain backward compatibility?

nvaytet (commented):
A couple of questions, otherwise looks fantastically simple!

docs/tutorials/exploring-data.ipynb
SimonHeybrock: Probably some combination of the above. I think it hard to come up with a good solution with only two examples, so I suggest to refactor once we have more.

python/src/scippneutron/data/__init__.py
SimonHeybrock: The pooch docs highly recommend using versions. And I think it does make sense. For example we might want to change tutorials from using data with fake artifacts/filtering to data where we can filter on something real. We still need to keep the old releases working.

docs/tutorials/exploring-data.ipynb
```diff
@@ -34,7 +34,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We  start by loading some data (download [here](https://github.com/ess-dmsc-dram/loki_tube_scripts/raw/master/test/test_data/LARMOR00049338.nxs)), in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
+    "We  start by loading some data, in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
```
nvaytet, line 37: Can you split long lines while you're at it?

docs/tutorials/working-with-masks.ipynb
```diff
@@ -36,7 +36,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We  start by loading some data (download [here](https://github.com/ess-dmsc-dram/loki_tube_scripts/raw/master/test/test_data/LARMOR00049338.nxs)), in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
+    "We  start by loading some data, in this case measured with a prototype of the [LoKI](https://europeanspallationsource.se/instruments/loki) detectors at the [LARMOR beamline](https://www.isis.stfc.ac.uk/Pages/Larmor.aspx):"
```
nvaytet, line 39: Split long line

nvaytet (changes requested):

docs/tutorials/exploring-data.ipynb
SimonHeybrock: What's wrong with once sentence per line?

nvaytet: One sentence per line is ok, but it's even easier to review if long sentences are split?

SimonHeybrock: You can split it next time you change the file ;)

nvaytet (approved):

PR #176, opened 2021-09-28
Tag solution cells

SimonHeybrock:
Add script to automatically strip solutions cells. This is used to create notebooks for courses.


docs/tutorials/2_working-with-masks.ipynb
```diff
@@ -285,13 +305,12 @@
    "source": [
```
nvaytet, line 305: As you mentioned yesterday during the course, in the cell further up where you compute `phi`, use `to_unit` instead of manual conversion from `rad` to `deg`?

docs/tutorials/3_understanding-event-data.ipynb
```diff
@@ -542,7 +550,9 @@
    "id": "77391eb4-d48a-4335-aa90-1587ff32865e",
    "metadata": {
     "jp-MarkdownHeadingCollapsed": true,
```
nvaytet, line 552: Delete this line?

nvaytet (changes requested):

PR #179, opened 2021-10-04
Install data for docs via pooch

jl-wynen:
Eliminates the need for downloading data and setting up paths manually.


docs/tutorials/powder-diffraction.ipynb
```diff
@@ -45,11 +45,11 @@
...
-    "sample = scn.load(filename='PG3_4844_event.nxs',\n",
-    "                  load_pulse_times=False,\n",
-    "                  mantid_args={'LoadMonitors': True})\n",
-    "vanadium = scn.load(filename='PG3_4866_event.nxs',\n",
-    "                    load_pulse_times=False)"
+    "sample = scn.data.example_data('PG3_4844_event.nxs',\n",
+    "                               load_pulse_times=False,\n",
+    "                               mantid_args={'LoadMonitors': True})\n",
+    "vanadium = scn.data.example_data('PG3_4866_event.nxs',\n",
+    "                                 load_pulse_times=False)"
```
SimonHeybrock, line 52: We may want to be able to chose how to load a file, e.g., using `scn.load` vs `scn.load_nexus`. Can we instead expose the path provided by pooch so we can chose here how to do so? This would also better illustrate and teach how to load files.

docs/user-guide/groupby.ipynb
```diff
@@ -36,7 +36,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "events = scn.load(filename='PG3_4844_event.nxs', load_pulse_times=False)"
+    "events = scn.data.example_data('PG3_4844_event.nxs', load_pulse_times=False)"
```
SimonHeybrock, line 39: I am not sure we should have arguments such as `load_pulse_times=False` on these functions? We do we start and stop in forwarding arguments? Maybe we should just get the path in case we want extra args, and can additionally have some dedicated named functions to returns specific data (without even exposing the filename)?

SimonHeybrock (commented):
Aren't we also using files for the unit tests? I was surprised to not see changes for that, where do they get the files now?

docs/tutorials/powder-diffraction.ipynb
```diff
@@ -45,10 +45,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sample = scn.load(filename='PG3_4844_event.nxs',\n",
+    "sample = scn.load(scn.data.locate('PG3_4844_event.nxs'),\n",
```
SimonHeybrock, line 48: Is `locate` misleading, maybe? Someone may think it is actually searching in the system filesystem?

jl-wynen: I was thinking of it in terms of the locator pattern. But that is of course no helpful to non-programmers. Do you have a better suggestion?

SimonHeybrock (approved):

PR #180, opened 2021-10-05
Store sample_position as attr and not coord in monitors

jl-wynen:
Storing `sample_position` as a coord of monitors means that monitor data cannot be combined with scattered data even after conversion to wavelength, d-spacing, etc. because conversions of monitors do not use the sample position.

### DIFF
--- a/python/src/scippneutron/mantid.py
+++ b/python/src/scippneutron/mantid.py
@@ -525,6 +525,11 @@ def convert_monitors_ws(ws, converter, **ignored):
                             WorkspaceIndexList=[index]) as monitor_ws:
             # Run logs are already loaded in the data workspace
             single_monitor = converter(monitor_ws, load_run_logs=False)
+        # Storing sample_position as a coord of monitors means that monitor
+        # data cannot be combined with scattered data even after conversion
+        # to wavelength, d-spacing, etc. because conversions of monitors do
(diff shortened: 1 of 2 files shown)


python/tests/test_mantid.py
```diff
@@ -343,7 +343,8 @@ def test_EventWorkspace_with_monitors(self):
             assert monitor.shape == [200001]
             assert 'position' in monitor.coords
             assert 'source_position' in monitor.coords
-            assert 'sample_position' in monitor.coords
+            assert 'sample_position' not in monitor.coords
```
SimonHeybrock, line 346: Maybe add a check in `test_Workspace2D_with_include_monitors` above to ensure that monitors there do not have this coord either?

python/src/scippneutron/mantid.py
```diff
@@ -526,6 +525,8 @@ def convert_monitors_ws(ws, converter, **ignored):
                             WorkspaceIndexList=[index]) as monitor_ws:
             # Run logs are already loaded in the data workspace
             single_monitor = converter(monitor_ws, load_run_logs=False)
```
SimonHeybrock, line 527: Add explanation you gave in PR description here as a comment, for future reference?

SimonHeybrock (approved):

PR #191, opened 2021-10-20
Load nxlog times as datetime64

Tom-Willemsen:
Loads `NXLog` timestamps as `datetime64` data type as opposed to `float64`.

The times are relative to the `start` attribute of the `NXLog` group if present, otherwise the current implementation falls back to the run start time. I'm unsure whether this is the correct approach or whether we should just reject any logs that don't contain a `start` attribute, despite it being optional in the nexus standard.

In the case where neither `start` nor `run_start_time` are present, loading the log is skipped with a warning as we wouldn't be able to construct a sensible `datetime`.

### DIFF
--- a/python/src/scippneutron/file_loading/_log_data.py
+++ b/python/src/scippneutron/file_loading/_log_data.py
@@ -51,33 +51,27 @@ def _convert_nxlog_time_to_datetime64(
             not provided, defaults to 1 (a no-op scaling factor).
     """
     try:
-        raw_times_ns = sc.to_unit(raw_times, sc.units.ns)
+        raw_times_ns = sc.to_unit(raw_times, sc.units.ns, copy=False)
     except sc.UnitError:
         raise BadSource(f"The units of time in the NXlog entry at "
                         f"'{group_path}/time{{units}}' must be convertible to seconds, "
                         f"but this cannot be done for '{raw_times.unit}'. Skipping "
                         f"loading NXLog at '{group_path}'.")
 
-    if log_start is not None:
-        try:
-            _start_ts = sc.scalar(value=np.datetime64(log_start),
-                                  unit=sc.units.ns,
-                                  dtype=sc.dtype.datetime64)
-        except ValueError:
-            raise BadSource(
-                f"The date string '{log_start}' in the NXLog entry at "
-                f"'{group_path}/time@start' failed to parse as an ISO8601 date. "
-                f"Skipping loading NXLog at '{group_path}'")
-    else:
-        _start_ts = sc.scalar(value=np.datetime64("1970-01-01T00:00:00Z"),
+    try:
+        _start_ts = sc.scalar(value=np.datetime64(log_start or "1970-01-01T00:00:00Z"),
                               unit=sc.units.ns,
                               dtype=sc.dtype.datetime64)
+    except ValueError:
+        raise BadSource(
+            f"The date string '{log_start}' in the NXLog entry at "
+            f"'{group_path}/time@start' failed to parse as an ISO8601 date. "
+            f"Skipping loading NXLog at '{group_path}'")
 
-    _scale = sc.scalar(value=scaling_factor if scaling_factor is not None else 1.,
-                       unit=sc.units.dimensionless,
-                       dtype=sc.dtype.float64)
+    _scale = sc.scalar(value=scaling_factor if scaling_factor is not None else 1,
+                       unit=sc.units.dimensionless)
 
-    return _start_ts + (raw_times_ns * _scale).astype(sc.dtype.int64)
+    return _start_ts + (raw_times_ns * _scale).astype(sc.dtype.int64, copy=False)
 
 
 def _add_log_to_data(log_data_name: str, log_data: sc.Variable, group_path: str,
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
> otherwise the current implementation falls back to the run start time. I'm unsure whether this is the correct approach or whether we should just reject any logs that don't contain a start attribute, despite it being optional in the nexus standard.

I would do neither: If there is no start attribute, just load the offsets, i.e., the final dtype would be int/float, not datetime. Only if there is a start attribute we return `start + offset`, i.e., dtype=datetime.

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -59,39 +51,33 @@ def _correct_nxlog_times(raw_times: sc.Variable,
...
+        _start_ts = sc.scalar(value=np.datetime64("1970-01-01T00:00:00Z"),
+                              unit=sc.units.ns,
+                              dtype=sc.dtype.datetime64)
 
     _scale = sc.scalar(value=scaling_factor if scaling_factor is not None else 1.,
                        unit=sc.units.dimensionless,
                        dtype=sc.dtype.float64)
 
-    return (raw_times_s * _scale) + (_log_start_ts - _run_start_ts)
+    return _start_ts + (raw_times_ns * _scale).astype(sc.dtype.int64)
```
SimonHeybrock, line 80:
```suggestion
    return _start_ts + (raw_times_ns * _scale).astype(sc.dtype.int64, copy=False)
```
to avoid redundant copy in case type is already correct.

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -59,39 +51,33 @@ def _correct_nxlog_times(raw_times: sc.Variable,
             not provided, defaults to 1 (a no-op scaling factor).
     """
     try:
-        raw_times_s = sc.to_unit(raw_times, sc.units.s)
+        raw_times_ns = sc.to_unit(raw_times, sc.units.ns)
```
SimonHeybrock, line 54:
```suggestion
        raw_times_ns = sc.to_unit(raw_times, sc.units.ns, copy=False)
```
to potentially avoid redundant copy.

But given that below we multiply be a `float64` scale, why is this converted at all?

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -59,39 +51,33 @@ def _correct_nxlog_times(raw_times: sc.Variable,
...
+                f"'{group_path}/time@start' failed to parse as an ISO8601 date. "
+                f"Skipping loading NXLog at '{group_path}'")
+    else:
+        _start_ts = sc.scalar(value=np.datetime64("1970-01-01T00:00:00Z"),
+                              unit=sc.units.ns,
+                              dtype=sc.dtype.datetime64)
 
     _scale = sc.scalar(value=scaling_factor if scaling_factor is not None else 1.,
                        unit=sc.units.dimensionless,
                        dtype=sc.dtype.float64)
```
SimonHeybrock, line 78: Maybe it would be better to not fix the dtype and use `1` (instead of `1.`), do avoid potential precision loss?

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -59,39 +51,33 @@ def _correct_nxlog_times(raw_times: sc.Variable,
...
+                                  dtype=sc.dtype.datetime64)
+        except ValueError:
+            raise BadSource(
+                f"The date string '{log_start}' in the NXLog entry at "
+                f"'{group_path}/time@start' failed to parse as an ISO8601 date. "
+                f"Skipping loading NXLog at '{group_path}'")
+    else:
+        _start_ts = sc.scalar(value=np.datetime64("1970-01-01T00:00:00Z"),
+                              unit=sc.units.ns,
+                              dtype=sc.dtype.datetime64)
```
SimonHeybrock, line 74: This is a bit of duplication, can just set `log_start` to fallback if None, and remove the `if`/`else`?

python/src/scippneutron/file_loading/_log_data.py
```diff
@@ -147,8 +133,9 @@ def _load_log_data_from_group(group: Group, nexus: LoadFromNexus,
     try:
         dimension_label = "time"
         is_time_series = True
-        raw_times = nexus.load_dataset(group.group, time_dataset_name,
-                                       [dimension_label])
+        raw_times = nexus.load_dataset(group.group,
+                                       time_dataset_name, [dimension_label],
+                                       dtype=sc.dtype.float64)
```
SimonHeybrock, line 138: Can you explain this change? I saw above the raw times are converted to `int64_t`, why do we now need an explicit type here?

python/src/scippneutron/file_loading/_log_data.py
Tom-Willemsen: The scale and the units are independent of each other I think... from the nexus standard:

> The scaling_factor, when present, has to be applied to the time values in order to arrive at the units specified in the units attribute

So we need to do both conversions (changing unit and applying scaling factor) to arrive at a consistent unit of `ns`.

Unless I have misunderstood your comment?

SimonHeybrock: No, my bad, since I didn't realize that the other was `astype`, not `to_unit`, ...

SimonHeybrock (approved):

Issue #34, opened 2021-03-12
Load log transformations with multiple values


matthew-d-jones:
Only NXlog transforms with a single value are currently loaded.
The full chain of transformations for each component, for example pixel positions, needs to be calculated for each value in the log and recorded with the timestamps.
This also happens in the case of streamed data.

Example file: https://project.esss.dk/owncloud/index.php/s/HlwUAoNY6lYrZiE
see sample group.

PR #200, opened 2021-10-27
Load transformations using new spatial transform data types

Tom-Willemsen:
Use the new spatial transform types in loading component positions.

This is related to https://github.com/scipp/scippneutron/issues/34 but does not yet fully fix that issue as we need to discuss how we want transforms to appear to the user in their loaded dataset (currently the code just uses the first transform, effectively throwing away any subsequent values).


src/scippneutron/file_loading/_transformations.py
```diff
@@ -229,23 +279,29 @@ def _get_transformation_magnitude_and_unit(
...
         magnitude = nexus.load_dataset_as_numpy_array(transform).astype(float).item()
         unit = nexus.get_unit(transform)
+
         if unit == sc.units.dimensionless:
             raise TransformationError(f"Missing units for transformation at "
                                       f"{nexus.get_name(transform)}")
-    return magnitude, sc.Unit(unit)
+
+        values = sc.array(dims=["value"], values=[magnitude], unit=unit)
+        times = sc.array(dims=["value"], values=[0.], unit=sc.units.s)
```
SimonHeybrock, line 288: - Same question about dim label as above.
- Returning a dummy time of `0` does seem quite error prone. Can't we just return what we have, i.e., a scalar `magnitude` (similar to the current code), and handle differences downstream?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -229,23 +279,29 @@ def _get_transformation_magnitude_and_unit(
...
                      group_name: str, nexus: LoadFromNexus):
-    angle, unit = _get_transformation_magnitude_and_unit(group_name, transform, nexus)
-    if unit == sc.units.deg:
-        angle = np.deg2rad(angle)
-    elif unit != sc.units.rad:
+    angles = _get_transformation_magnitude_and_unit(group_name, transform, nexus)
+    try:
+        angles = sc.to_unit(angles.astype(sc.DType.float64, copy=False),
+                            sc.units.rad,
+                            copy=False)
```
SimonHeybrock, line 300: Can use `to(unit=..., dtype=...)`

src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,18 +211,26 @@ def _append_translation(offset: np.ndarray, transform: GroupObject,
...
                         nexus: LoadFromNexus):
-    magnitude, unit = _get_transformation_magnitude_and_unit(group_name, transform,
-                                                             nexus)
+    loaded_transform = _get_transformation_magnitude_and_unit(
+        group_name, transform, nexus)
+
+    loaded_transform_m = sc.to_unit(loaded_transform.astype(sc.DType.float64,
+                                                            copy=False),
+                                    sc.units.m,
+                                    copy=False)
```
SimonHeybrock, line 220: Can use `to()`

src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,18 +211,26 @@ def _append_translation(offset: np.ndarray, transform: GroupObject,
...
-        magnitude_var = magnitude * unit
-        magnitude_var = sc.to_unit(magnitude_var, sc.units.m)
-        magnitude = magnitude_var.value
     # -1 as describes passive transformation
-    vector = direction_unit_vector * -1. * magnitude
-    offset_vector = vector + offset
-    matrix = np.block([[np.eye(3), offset_vector[np.newaxis].T], [0., 0., 0., 1.]])
-    transformations.append(matrix)
+    vectors = sc.vector(
+        value=direction_unit_vector) * sc.scalar(-1.0) * loaded_transform_m
```
SimonHeybrock, line 224:
```suggestion
    vectors = sc.vector(value=-direction_unit_vector) * loaded_transform_m
```
more readable?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,18 +211,26 @@ def _append_translation(offset: np.ndarray, transform: GroupObject,
...
+    translations = sc.spatial.translations(dims=loaded_transform_m.dims,
+                                           values=vectors.values,
+                                           unit=sc.units.m)
+
+    translations = translations * sc.spatial.translation(value=offset, unit=sc.units.m)
```
SimonHeybrock, line 229: Having trouble understanding what is done in this function:
1. We load translation magnitudes
2. We create a vector from a direction and the translation magnitudes
3. We convert the vector into a translation
4. We create another translation from "offset"
5. We combine this with the first translation

I don't if I misunderstood, but this is very hard to see through?

Why not use plain addition to add the two offset vectors (convert to `translation3` only at the very end)?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,18 +211,26 @@ def _append_translation(offset: np.ndarray, transform: GroupObject,
...
+
+    transformations.append(
+        sc.DataArray(data=translations,
+                     coords={"time": loaded_transform.coords["time"]}))
```
SimonHeybrock, line 233: Why is our math dropping the coord? Because it has no choice since we currently do not support operations from `spatial` with data arrays?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -67,10 +86,26 @@ def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> np.nda
...
-    total_transform_matrix = np.identity(4)
-    for transformation in transformations:
-        total_transform_matrix = np.matmul(transformation, total_transform_matrix)
-    return total_transform_matrix
+
+    total_transform = sc.DataArray(
+        data=sc.spatial.affine_transforms(dims=["value"],
+                                          values=[np.identity(4)],
+                                          unit=sc.units.m),
+        coords={"time": sc.array(dims=["value"], values=[0.], unit=sc.units.s)})
```
SimonHeybrock, line 94: - As written elsewhere, we should probably work with datetimes, such as for other `NXlog`?
- This this to arbitrarily introduce an additional time point with no transform. With the interpolation below, couldn't this lead to completely broken results, e.g., if there is an actual transform just below 0 that is overridden by this?

Can we just initialize the output transform to the first (in time) transform and only multiply the remaining ones?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -67,10 +86,26 @@ def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> np.nda
...
+
+    for transform in transformations:
+        xnew = sc.sort(
+            sc.concat([
+                total_transform.coords["time"].to(dtype=sc.DType.float64,
+                                                  unit=sc.units.s),
+                transform.coords["time"].to(dtype=sc.DType.float64, unit=sc.units.s),
+            ],
+                      dim="value"), "value")
```
SimonHeybrock, line 103: Doesn't this give bad results if there are multiple transforms at the same time? Should `xnew` avoid duplicates?

SimonHeybrock (changes requested):

src/scippneutron/file_loading/_transformations.py
```diff
@@ -47,7 +49,24 @@ def get_position_from_transformations(group: Group, nexus: LoadFromNexus) -> np.
     :return: Position of the component as a three-element numpy array
```
jl-wynen, line 49:
```suggestion
    :return: Position of the component as a three-element Variable
```

src/scippneutron/file_loading/_transformations.py
```diff
@@ -47,7 +49,24 @@ def get_position_from_transformations(group: Group, nexus: LoadFromNexus) -> np.
...
     """
     total_transform_matrix = get_full_transformation_matrix(group, nexus)
-    return np.matmul(total_transform_matrix, np.array([0, 0, 0, 1], dtype=float))[0:3]
+    return total_transform_matrix * sc.vector(value=[0, 0, 0], unit=sc.units.m)
```
jl-wynen, line 52: Is it correct that this multiplication extract the translation part of the transform? We should add a nicer way of doing that in scipp.

src/scippneutron/file_loading/_transformations.py
```diff
@@ -47,7 +49,24 @@ def get_position_from_transformations(group: Group, nexus: LoadFromNexus) -> np.
...
+                                        "value",
+                                        kind="previous",
+                                        fill_value="extrapolate")(xnew=xnew)
+
+    transform.coords["time"] = transform.coords["value"]
+    del transform.coords["value"]
+    return transform
 
 
 def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> np.ndarray:
```
jl-wynen, line 72:
```suggestion
def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> sc.DataArray:
```
And in docstring

src/scippneutron/file_loading/_transformations.py
Tom-Willemsen: Is there a convenient way to do this in scipp without converting to numpy and then back again? Or would it be worth me adding something like `sc.unique()` in scipp (acting like [`np.unique`](https://numpy.org/doc/stable/reference/generated/numpy.unique.html))?

SimonHeybrock: There is currently no way I think, i.e., you would indeed need to use `np.unique`. I think adding it in scipp would be ok (but that does not help you here, since you'd need to wait for the next release), but I'd use numpy for now.

src/scippneutron/file_loading/_transformations.py
Tom-Willemsen: > Why is our math dropping the coord? Because it has no choice since we currently do not support operations from `spatial` with data arrays?

Yes, we don't have a function which converts vector -> translation as a data array, we only have constructors for `Variable`s with the spatial transform types.

src/scippneutron/file_loading/_transformations.py
Tom-Willemsen: >     * Returning a dummy time of `0` does seem quite error prone. Can't we just return what we have, i.e., a scalar `magnitude` (similar to the current code), and handle differences downstream?

I'm not sure I understand what you're suggesting here. Do you mean that as a user I would have some transforms loaded as scalars and others as arrays along the `time` dimension? Isn't that confusing for the users - as a user I think I would expect a consistent return format?

Or did you mean something else by "handle the differences downstream"... in which case where is "downstream" in this instance?

src/scippneutron/file_loading/_transformations.py
Tom-Willemsen: I'll create a separate issue for this in scipp - we'll need to come up with a good non-ambiguous name for it.

Tom-Willemsen: Actually, on second thoughts, I'm not sure this makes sense? If we had a function like `sc.spatial.translation_component(transform)`, I think it would make sense for it to be returning `translation3` and not `vector3`, so we would still need to do another conversion from `translation3`->`vector3`? I'm now thinking it probably does make more sense to think of this as "appyling the transform to the zero vector" rather than "extracting the translation component". 

Happy to discuss if you disagree though?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -18,24 +20,24 @@ class TransformationError(Exception):
...
+        angles: a dataset containing the angles
+    Returns:
+        A dataset of rotation matrices.
+    """
+    rotvec = sc.vector(value=axis)
+    # We multiply by -1*angle to get a "passive transform"
+    rotvecs = rotvec * sc.scalar(-1.0) * angles.astype(sc.DType.float64, copy=False)
+    matrices = sc.spatial.rotations_from_rotvecs(dims=angles.dims,
+                                                 values=rotvecs.values,
+                                                 unit=sc.units.rad)
```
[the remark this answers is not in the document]
Tom-Willemsen: I think we did it like this to be consistent with all the other creation functions in the table at https://github.com/scipp/scipp/issues/2252#issuecomment-990864357 which take numpy types and return `Variable`. I think the consistency is good - but perhaps we should additionally have some kind of wrapper/helper that takes `Variable` as well to avoid the error-prone api? What do you think?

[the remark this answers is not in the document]
SimonHeybrock: I think we want consistency for the *direct constructors* (`vectors`, `affine_transform`, ...), but `rotations_from_rotvecs` is somewhat "higher level" functionality (we could almost think of it as a variant of `astype()` -- I am not at all suggesting to combine it though). I think we should generally force users of scipp to use `Variable` wherever possible, eliminating as many sources of bugs as possible (while keeping the code and design simple).

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: > Do you mean that as a user I would have some transforms loaded as scalars and others as arrays along the time dimension? Isn't that confusing for the users - as a user I think I would expect a consistent return format?

I think that is what I mean. I don't think it would add confusion (mutating what is in the file *does*). If some transform does not have `time` dimension then that's fine since scipp's automatic broadcasting can take care of it, can't it?

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: We should probably have a look if that can be alleviated.

SimonHeybrock (commented):
Looks pretty good, a couple more comments below.

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -92,23 +92,14 @@ def _load_pixel_positions(detector_group: Group, detector_ids_size: int,
...
-        # Get and apply transformation matrix
-        transformation = get_full_transformation_matrix(detector_group, nexus)
-        for row_index in range(n_rows):
-            array[row_index, :] = np.matmul(transformation, array[row_index, :])
-
-        # Now the transformations are done we do not need the 4th
-        # element in each position
-        array = array[:, :3]
+        data = (get_full_transformation_matrix(detector_group, nexus) * data)["time",
+                                                                              0].data
```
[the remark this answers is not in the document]
Tom-Willemsen: Yes, this is basically a "hack" for the moment that keeps this code compatible with what we are currently returning. The next PR (not opened yet) removes this "hack" and returns the full transform dataarray + base position to the user.

src/scippneutron/file_loading/_transformations.py
```diff
@@ -85,30 +88,26 @@ def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> sc.Dat
...
+                ],
+                          dim="time").values),
+                                dims=["time"])
+            total_transform = _interpolate_transform(
+                transform, xnew) * _interpolate_transform(total_transform, xnew)
+        else:
+            if isinstance(transform, sc.DataArray):
+                xnew = transform.coords["time"]
+                total_transform = _interpolate_transform(
+                    transform, xnew) * _interpolate_transform(total_transform, xnew)
```
SimonHeybrock, line 108: Why do we need to interpolate `total_transform`? If I understand correctly it is a variable (without time dim), so shouldn't this just be something like:
```suggestion
            if isinstance(transform, sc.DataArray):
                xnew = transform.coords["time"]
                total_transform = _interpolate_transform(transform, xnew) * total_transform
```
would that also obviate the branch in `_interpolate_transform` above?

SimonHeybrock (approved):

PR #203, opened 2021-11-02
Add coverage build support

SimonHeybrock:
This setup should be completely symmetric to the one in scipp. I have improved `scipp-coverage` actions to not make explicit reference to the repo under test, so we should be able to make a 1:1 copy of that for `scippneutron-coverage` (once this is merged).

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,31 @@
+{
+  "version": 2,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 16,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "coverage",
+      "displayName": "coverage",
+      "description": "Options for coverage build (Linux only)",
+      "binaryDir": "${sourceDir}/build",
+      "generator": "Unix Makefiles",
+      "cacheVariables": {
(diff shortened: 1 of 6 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,31 @@
...
+      "displayName": "coverage",
+      "description": "Options for coverage build (Linux only)",
+      "binaryDir": "${sourceDir}/build",
+      "generator": "Unix Makefiles",
+      "cacheVariables": {
+        "COVERAGE": "ON",
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "OFF",
+        "DISABLE_MULTI_THREADING": "ON"
```
jl-wynen, line 20: Does gcov not work with multi threaded programs?

jl-wynen (approved):

CMakePresets.json
SimonHeybrock: It does, but the build is about 2x slower with TBB, and they way we do threading in scipp this should not make a difference for coverage results.

PR #204, opened 2021-11-02
Allow setting cache directory with env variable

jl-wynen:


src/scippneutron/data/__init__.py
```diff
@@ -13,6 +13,7 @@ def _make_pooch():
     import pooch
     return pooch.create(
         path=pooch.os_cache('scippneutron'),
+        env='SCIPPNEUTRON_DATA_DIR',
```
SimonHeybrock, line 16: Add some documentation for this flag?

jl-wynen: Good point

SimonHeybrock (approved):

PR #207, opened 2021-11-04
Fix invisible plots in docs

nvaytet:
There was a bad syntax for the `Session.metadata` in the `nbsphinx_execute_arguments` inside the `docs/conf.py` file.

I also changed `Scippneutron version` to `Version`, as I felt it was a bit too long.


jl-wynen (approved):
Works

PR #208, opened 2021-11-04
Fix docs for scipp v0.9

nvaytet:
Do not merge before scipp v0.9 is released

PR #210, opened 2021-11-10
Use concat for scipp-0.9

SimonHeybrock:
This should also significantly (or massively, if many banks) speed up `load_nexus` for Nexus files with multiple banks.


tests/test_load_nexus.py
```diff
@@ -1815,7 +1815,9 @@ def test_load_raw_detector_data_from_nexus_file(load_function: Callable):
         end=sc.array(dims=["pulse"], values=[3, 3, 5], dtype=sc.dtype.int64),
     )
 
-    expected = sc.DataArray(data=binned,
+    # Even if there is just a single NXevent_data entry in the file the
+    # output has a `bank` dimension, for consistency.
+    expected = sc.DataArray(data=sc.concat([binned], 'bank'),
```
jl-wynen, line 1820: Not sure if this helps with usability? It means that people with a single bank always have to slice out that dimension.

SimonHeybrock: I think there are two cases:
1. Load a specific NXevent_data (not supported yet). Then indeed there should not be a `bank` dimension.
2. Load entire file. Typically files have an unknown number of banks, and even the same instrument may have a fluctuating number of banks, e.g., when one is broken and removed, or new ones added.

For the latter case, writing scripts that don't break is easier if we consistently include the `bank` dimension here?

jl-wynen:
There might be overlap with  #208

SimonHeybrock:
> There might be overlap with #208

There shouldn't be, I checked before.

nvaytet (approved):

PR #211, opened 2021-11-10
Detect weighted events properly in `from_mantid`

### LINKED ISSUE #201: Conversion from mantid truncates events to int
Loading event data written by MCStas via `scan.load` turns float events into ints. The reason is that mantid marks the data as being non weighted events. But loading the same file in MantidFramework and saving to nexus produces a file that can be loaded correctly by scippneutron.

jl-wynen:
Fixes #201

### DIFF
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -17,7 +17,7 @@ Features
 Bugfixes
 ~~~~~~~~
 
-* Fix bug that loaded weighted events without their weights `#211 <https://github.com/scipp/scippneutron/pull/211>`_.
+* Fix bug in ``load`` that loaded weighted events without their weights if the first spectrum is empty `#211 <https://github.com/scipp/scippneutron/pull/211>`_.
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -14,6 +14,11 @@ Features
 
 * Add ``tof.conversions`` module with building blocks for custom coordinate transformation graphs `#187 <https://github.com/scipp/scipp/pull/187>`_.
 
+Bugfixes
+~~~~~~~~
+
+* Fix bug that loaded weighted events without their weights `#211 <https://github.com/scipp/scippneutron/pull/211>`_.
+
 Breaking changes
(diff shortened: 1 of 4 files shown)


src/scippneutron/mantid.py
```diff
@@ -613,10 +619,7 @@ def convert_EventWorkspace_to_data_array(ws,
     pulse_times = sc.empty(
         dims=['event'], shape=[n_event], dtype=sc.dtype.datetime64,
         unit=sc.units.ns) if load_pulse_times else None
-
-    evtp = ws.getSpectrum(0).getEventType()
-    contains_weighted_events = ((evtp == EventType.WEIGHTED)
-                                or (evtp == EventType.WEIGHTED_NOTIME))
+    contains_weighted_events = _contains_weighted_events(ws)
```
SimonHeybrock, line 622: I would recommend to move this check into the loop, and just copy the weights if they exist? `getSpectrum()` is relatively cheap, but there is *some* cost to it.

docs/about/release-notes.rst
```diff
@@ -14,6 +14,11 @@ Features
 
 * Add ``tof.conversions`` module with building blocks for custom coordinate transformation graphs `#187 <https://github.com/scipp/scipp/pull/187>`_.
 
+Bugfixes
+~~~~~~~~
+
+* Fix bug that loaded weighted events without their weights `#211 <https://github.com/scipp/scippneutron/pull/211>`_.
```
SimonHeybrock, line 20:
```suggestion
* Fix bug in ``load`` that loaded weighted events without their weights if there was any empty spectrum `#211 <https://github.com/scipp/scippneutron/pull/211>`_.
```

src/scippneutron/mantid.py
```diff
@@ -631,7 +627,7 @@ def convert_EventWorkspace_to_data_array(ws,
         if load_pulse_times:
             pulse_times['event',
                         current:current + size].values = sp.getPulseTimesAsNumpy()
-        if contains_weighted_events:
+        if _contains_weighted_events(ws.getSpectrum(i)):
```
SimonHeybrock, line 630:
```suggestion
        if _contains_weighted_events(sp):
```

SimonHeybrock (approved):

PR #214, opened 2021-11-11
Change version selector to 0.4.1

jl-wynen:
Is this all or do I need to make more changes for the release?


SimonHeybrock (approved):
I think that is it, yes.

PR #218, opened 2021-11-12
Use same tempdir for html and doctest builders

nvaytet:


docs/build_and_test.py
```diff
@@ -21,8 +21,6 @@
         ],
                               stderr=subprocess.STDOUT,
                               shell=shell)
-
-    with tempfile.TemporaryDirectory() as work_dir:
         subprocess.check_call([
             'python', 'build_docs.py', '--builder=doctest', f'--prefix={build_dir}',
             f'--work_dir={work_dir}', '--no-setup'
```
SimonHeybrock, line 26: Didn't you say this option should be removed?

SimonHeybrock (approved):

PR #223, opened 2021-11-16
Fix linkcheck errors

jl-wynen:
Needed for https://github.com/scipp/scipp/issues/2290

### DIFF
--- a/docs/build_docs.py
+++ b/docs/build_docs.py
@@ -2,9 +2,9 @@
 # Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
 # @author Neil Vaytet
 
-import os
 import argparse
-from pathlib import Path
+import os
+import pathlib
 import subprocess
 import sys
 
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
@nvaytet Is just setting up a common docs template, how does this affect that work (or vice versa)?

jl-wynen:
> @nvaytet Is just setting up a common docs template, how does this affect that work (or vice versa)?

We should unify the blacklist in `conf.py`. But otherwise, the two efforts should be independent.

nvaytet:
> > @nvaytet Is just setting up a common docs template, how does this affect that work (or vice versa)?
> 
> We should unify the blacklist in `conf.py`. But otherwise, the two efforts should be independent.

See https://github.com/scipp/templates/pull/1/commits/a085c808f72ece38b17048929baf21cbbddf495d

nvaytet (approved):

PR #224, opened 2021-11-16
Remove .events property from load_nexus

nvaytet:


src/scippneutron/file_loading/_detector_data.py
```diff
@@ -350,7 +348,7 @@ def _bin_events(data: DetectorData):
     events = sc.concat([_bin_events(item) for item in detectors], _dim)
 
     if bin_by_pixel:
-        event_tofs = events.events.coords[_time_of_flight]
+        event_tofs = events.bins.constituents['data'].coords[_time_of_flight]
         _min_tof = event_tofs.min()
         _max_tof = event_tofs.max()
```
SimonHeybrock, line 353:
```suggestion
        _min_tof = events.bins.coords[_time_of_flight].min()
        _max_tof = events.bins.coords[_time_of_flight].max()
```
should be used instead.

jl-wynen (commented):
~How did this not surface earlier? Are we missing tests?~

Never mind. The event property was only removed on `0.10`.

SimonHeybrock (approved):

PR #231, opened 2021-12-02
Update llnl/units and build conan recipe in cmake configure stage

### LINKED ISSUE #177: Update for new mechanism of building 'units' conan package in scipp
See scipp/scipp#2201. This needs to be done after next scipp release.

SimonHeybrock:
Closes #177.

There is some new (and old!) duplication between scipp and scippneutron
here. Most conan dependencies of scippneutron are just needed by scipp,
so the conanfile and the cmake setup are identical. The new llnl-units
recipe is just another example.

For the recipe, the intention was actually to ship it with scipp, find
scipp, and use that recipe here. However, there is a chicken-egg
problem, since finding the scipp package has LLNL-Units as a dependency,
i.e., it cannot be fulfilled *after* calling find_package.

One solution could be to put conan setup into scipp-config.cmake.in and
ship the conanfile with scipp. Provided that this work, the only
duplication left would be that of the conan cmake function calls.

The disadvantages here would be that:
- Libraries using scip would be forced to use conan (not actuall issue
  for us now)
- scippneutron would "rely" and some dependencies of scipp getting
  install that are actually not runtime requirements, in particular
  GTest and pybind11.


jl-wynen:
> One solution could be to put conan setup into scipp-config.cmake.in and
ship the conanfile with scipp. Provided that this work, the only
duplication left would be that of the conan cmake function calls.

Would this allow users of scipp to install additional packages through conan? It sounds like this approach would completely encapsulate conan in scipp.

lib/conanfile.txt
```diff
@@ -3,7 +3,7 @@ benchmark/1.5.3
 boost/1.76.0
 eigen/3.3.9
 gtest/1.11.0
-LLNL-Units/0.5.0@scipp/stable
+LLNL-Units/0.5.0.1
```
jl-wynen, line 6: What does the suffix mean? The recipe specifies `0.5.0`.

jl-wynen (approved):

lib/conanfile.txt
SimonHeybrock: Is see 0.5.0.1 in line 14?

We have this extra digit in the version because we are not (yet) using an upstream release, but our own fork.

SimonHeybrock:
> Would this allow users of scipp to install additional packages through conan? It sounds like this approach would completely encapsulate conan in scipp

Yes and no:
- We should be able to run `find_package` for any package that is defined in the scipp conanfile.
- I don't think we can add new things to the conanfile?

lib/conanfile.txt
jl-wynen: Ah, right. I looked at the wrong line.

jl-wynen:
> 

I guess, this is yet another reason to make a conan package for the C++ part of scipp.

Issue #228, opened 2021-11-17
`load_nexus` yields `ZeroDivisionError`


SimonHeybrock:
Malcolm reports this issue. It may be a similar problem as fixed in https://github.com/scipp/scipp/pull/2216.

```python
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
/var/folders/q4/m1ldym055j50dkwt_5ypq17rwkh6dq/T/ipykernel_78437/2965978507.py in <module>
----> 1 data = scn.load_nexus('/Users/66j/Documents/ORNL/CISissues/VULCAN/VULCAN_195855.nxs.h5')

~/miniconda3/envs/scippneutron/lib/python3.9/site-packages/scippneutron/file_loading/load_nexus.py in load_nexus(data_file, root, quiet, bin_by_pixel)
    136 
    137     with _open_if_path(data_file) as nexus_file:
--> 138         loaded_data = _load_data(nexus_file,
    139                                  root,
    140                                  LoadFromHdf5(),

~/miniconda3/envs/scippneutron/lib/python3.9/site-packages/scippneutron/file_loading/load_nexus.py in _load_data(nexus_file, root, nexus, quiet, bin_by_pixel)
    174             f"{__name__}('my_file.nxs', '/entry_2')")
    175 
--> 176     loaded_data = load_detector_data(groups[nx_event_data], groups[nx_detector], nexus,
    177                                      quiet, bin_by_pixel)
    178     # If no event data are found, make a Dataset and add the metadata as

~/miniconda3/envs/scippneutron/lib/python3.9/site-packages/scippneutron/file_loading/_detector_data.py in load_detector_data(event_data_groups, detector_groups, nexus, quiet, bin_by_pixel)
    289                        bin_by_pixel: bool) -> Optional[sc.DataArray]:
    290     detectors = _load_data_from_each_nx_detector(detector_groups, nexus)
--> 291     detectors = _load_data_from_each_nx_event_data(detectors, event_data_groups, nexus,
    292                                                    quiet)
    293 

~/miniconda3/envs/scippneutron/lib/python3.9/site-packages/scippneutron/file_loading/_detector_data.py in _load_data_from_each_nx_event_data(detector_data, event_data_groups, nexus, quiet)
    414         parent_path = "/".join(group.name.split("/")[:-1])
    415         try:
--> 416             new_event_data = _load_event_group(
    417                 group, nexus, detector_data.get(parent_path, DetectorData()), quiet)
[...]

PR #233, opened 2021-12-02
Fix loading datasets with zero length

Tom-Willemsen:
Fixes https://github.com/scipp/scippneutron/issues/228

I don't have a nexus file that reproduces this, but the test added in this PR did reproduce the same error trace.

This is equivalent to the fix in scipp at https://github.com/scipp/scipp/pull/2216

### DIFF
--- a/tests/test_load_nexus.py
+++ b/tests/test_load_nexus.py
@@ -1827,3 +1827,20 @@ def test_load_raw_detector_data_from_nexus_file(load_function: Callable):
                             })
 
     assert sc.identical(loaded_data, expected)
+
+
+def test_load_nexus_file_containing_empty_arrays(load_function: Callable):
+    event_data = EventData(
+        event_id=np.array([]),
+        event_time_offset=np.array([]),
+        event_time_zero=np.array([]),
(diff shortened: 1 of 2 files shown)


tests/test_load_nexus.py
```diff
@@ -1827,3 +1827,20 @@ def test_load_raw_detector_data_from_nexus_file(load_function: Callable):
...
+        event_index=np.array([]),
+    )
+
+    builder = NexusBuilder()
+    builder.add_event_data(event_data)
+    builder.add_log(Log("test_log", np.array([0]), np.array([0])))
+
+    # Empty datasets should not stop other data (e.g. metadata) from being loaded.
+    loaded_data = load_function(builder)
+    assert "test_log" in loaded_data
```
jl-wynen, line 1846: Please also check that the loaded log has the same values that you put in.

jl-wynen (approved):
Looks good, only a minor comment.

We should ask Malcolm if this fixes the bug for him. But we should not add a new file just for this one test, the artificial data should be enough.

jl-wynen (approved):

PR #234, opened 2021-12-07
Pip packaging

SimonHeybrock:


SimonHeybrock:
As we have decided(?) to move ahead with #342 this is no longer relevant (as making the wheels can just use what we do, e.g., in scippnexus), so I am closing this.

Issue #236, opened 2021-12-16
Transformations from `event_time_zero` and `event_time_offset` to `pulse_time` and `tof`


SimonHeybrock:
Relating to #225: When this change is implemented, we will need functionality to transform to `pulse_time` and `tof`. It is interesting that this is actually a family of transforms:

1. Simple definition of `t_0`, yielding offset w.r.t. `event_time_zero` and `event_time_offset` from file to directly compute `pulse_time` and `tof`
2. WFM, where we have multiple "t_0" within a pulse and need to *split* pulses.
3. Pulse-skipping on instruments with simultaneous study of wide energy ranges. Choppers block, e.g., every other pulse, but as far as I understand data is still written with the normal (14 Hz at ESS) `event_time_zero` definition. We thus need to *merge* pulses after `t_0` definition, i.e., kind of the reverse of the WFM case.

In all cases parameters for the transformation may be computable from choppers, but this most likely involves instrument-specifics. Here we therefore would restrict ourselves to generic functions to facility such transformations.

We should consider whether performing these transformation would work better before binning into pixels, i.e., with binning as in the Nexus file (by pulse).

We should also not forget implications for meta-data (such as proton charge or sample temperatures). We need to ensure that we are able to correlate these to events, after transformation of the times. It will typically be irrelevant for slow meta data, but everything else will require special attention. Is there a sensible way to handle this automatically? Should all attributes that hold time-series logs be rewritten to `pulse_time`?

SimonHeybrock:
Case 3 is hard or impossible to do nicely and consistently without rounding problems and potentially drift problems for data that has been binned into pixels, since then we typically do not have the original list of pulse times. We should therefore perform this transformation in the "Nexus" layout (binned by pulses) before binning into pixels.

SimonHeybrock:
> Case 3 is hard or impossible to do nicely and consistently without rounding problems and potentially drift problems for data that has been binned into pixels, since then we typically do not have the original list of pulse times. We should therefore perform this transformation in the "Nexus" layout (binned by pulses) before binning into pixels.

This is actually not possible (unless we want to do per-event geometry calculations): The offset depends on the detector pixel (L2, also energy in inelastic), so it is best performed *after* binning to pixels.

SimonHeybrock:
See https://scipp.github.io/scippneutron/reference/frame-unwrapping.html for some early work on this.

PR #242, opened 2022-01-10
Frame handling

SimonHeybrock:
See #236.

Work in progress: What interface to we want/need?


docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+   "cell_type": "markdown",
+   "id": "aaf4cac3-6a54-44a2-95f7-a0518474589e",
+   "metadata": {},
+   "source": [
+    "In the figure above, we define for a source frame (pulse) with index `i`\n",
+    "\n",
+    "- $T_0^i$ is the `event_time_zero` recorded in an `NXevent_data` group.\n",
+    "  These times are indicated by the vertical dotted lines.\n",
+    "- $T_0^{i+1} = T_0 + L_0$ with the frame length $L_0 = 1/f_0$ given by the source frequency $f_0$.\n",
+    "- $\\Delta T_0$ is the offset from $T_0^i$ at which the neutrons are \"emitted\".\n",
```
nvaytet, line 48: The definition of `Delta_T0` would benefit from zooming in onto the pulse to show better what we mean

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+    "  This may be zero if the full pulse length is used, but choppers such as resolution choppers may extract a section of the pulse that is not aligned with the start of the full pulse.\n",
+    "  This offset can also be used to take into account a potential difference between the timing system's definition of the pulse time and the actual beginning of the neutron pulse exiting, e.g., the moderator.\n",
+    "\n",
+    "For every event recorded in a detector, `event_time_offset` ($\\Delta t$) in an `NXevent_data` group is the offset from the latest (previous) `event_time_zero` ($T_0^j$), i.e., the time difference to the previous vertical dotted line.\n",
+    "\n",
+    "To compute the time-of-flight for a neutron, we need to identify which source pulse it orginated from.\n",
+    "Consider the wavelength band for frame `i`, i.e., neutrons produced by the second source pulse in the figure above:\n",
```
nvaytet, line 55:
```suggestion
    "Consider the wavelength band for frame `i`, i.e., neutrons produced by the second (orange) source pulse in the figure above:\n",
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+    "\n",
+    "- The fastest neutrons from this frame reach detector 1 one frame later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+1}$.\n",
+    "- The slowest neutrons from this frame reach detector 1 two frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+2}$.\n",
+    "- The fastest neutrons from this frame reach detector 2 two frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+2}$.\n",
+    "- The slowest neutrons from this frame reach detector 2 three frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+3}$.\n",
```
nvaytet, line 60: I think these would need names or annotations on the figure, but keeping the figure non-messy will be difficult..

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+    "\n",
+    "We thus need to add an integer multiple of the frame length to `event_time_offset`.\n",
+    "Within a given frame (indicated above by an area between two dotted vertical lines) there is a *pivot point*:\n",
+    "Neutrons with `event_time_offset` *below* the pivot point originated one source frame *before* neutrons *above* the pivot point.\n",
```
nvaytet, line 64: Be careful with using the words `below` and `above` when talking about this diagram, because instinctively, the reader will think "below" in the figure, while time offsets are left to right.

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+    "To compute the time-of-flight for a neutron, we need to identify which source pulse it orginated from.\n",
+    "Consider the wavelength band for frame `i`, i.e., neutrons produced by the second source pulse in the figure above:\n",
+    "\n",
+    "- The fastest neutrons from this frame reach detector 1 one frame later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+1}$.\n",
+    "- The slowest neutrons from this frame reach detector 1 two frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+2}$.\n",
+    "- The fastest neutrons from this frame reach detector 2 two frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+2}$.\n",
+    "- The slowest neutrons from this frame reach detector 2 three frames later, i.e., the `event_time_offset` for those fast neutrons will be relative to $T_0^{i+3}$.\n",
+    "\n",
+    "We thus need to add an integer multiple of the frame length to `event_time_offset`.\n",
+    "Within a given frame (indicated above by an area between two dotted vertical lines) there is a *pivot point*:\n",
```
nvaytet, line 63: I don't see the 'area' you are talking about

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+    "In the figure above, we define for a source frame (pulse) with index `i`\n",
+    "\n",
+    "- $T_0^i$ is the `event_time_zero` recorded in an `NXevent_data` group.\n",
+    "  These times are indicated by the vertical dotted lines.\n",
+    "- $T_0^{i+1} = T_0 + L_0$ with the frame length $L_0 = 1/f_0$ given by the source frequency $f_0$.\n",
+    "- $\\Delta T_0$ is the offset from $T_0^i$ at which the neutrons are \"emitted\".\n",
+    "  This may be zero if the full pulse length is used, but choppers such as resolution choppers may extract a section of the pulse that is not aligned with the start of the full pulse.\n",
+    "  This offset can also be used to take into account a potential difference between the timing system's definition of the pulse time and the actual beginning of the neutron pulse exiting, e.g., the moderator.\n",
+    "\n",
+    "For every event recorded in a detector, `event_time_offset` ($\\Delta t$) in an `NXevent_data` group is the offset from the latest (previous) `event_time_zero` ($T_0^j$), i.e., the time difference to the previous vertical dotted line.\n",
```
nvaytet, line 52: I think you need to label `event_time_offset` on the figure somewhere

docs/reference/frame-unwrapping.ipynb
nvaytet, line 64: Are you trying to say that the pivot point is half-way between the blue and orange areas? If so, can you draw a diagonal line that travels between the frames? Would this help?
Right now I don't really understand what the arrows are pointing at.

nvaytet: hmm ok, below you say that the pivot point depends on `lambda_min`, so it's not half-way.
are the arrows supposed to be pointing at the intersection between the left edge of the orange area and the detector lines?

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,132 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "aaf4cac3-6a54-44a2-95f7-a0518474589e",
+   "metadata": {},
+   "source": [
+    "In the figure above, we define for a source frame (pulse) with index `i`\n",
+    "\n",
+    "- $T_0^i$ is the `event_time_zero` recorded in an `NXevent_data` group.\n",
+    "  These times are indicated by the vertical dotted lines.\n",
+    "- $T_0^{i+1} = T_0 + L_0$ with the frame length $L_0 = 1/f_0$ given by the source frequency $f_0$.\n",
```
nvaytet, line 47:
```suggestion
    "- $T_0^{i+1} = T_0 + L_0$, where the frame length $L_0$ is defined by $L_0 = 1/f_0$, given a source frequency $f_0$.\n",
```

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: I mean vertical band.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: > Are you trying to say that the pivot point is half-way between the blue and orange areas? 

It does not really matter where exactly, it just has to be somewhere between the two.

SimonHeybrock: Right now the code computes it based on `lambda_min`, so it is defined as the left edge (border line) of a frame.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: Maybe I should invert the logic: Talk about a specific frame (shade a vertical band between two dotted lines) and say:
- First (small `event_time_offset`) we see the slowest neutrons from N frames earlier.
- Then (larger `event_time_offset`) we see the slowest neutrons from N-1 frames earlier.

... and possibly avoid repeating this for both detectors, i.e., just give the example of 1 o f them?

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: Changed to `before` and `after`, and also renamed `pivot point` -> `pivot time`.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: I made the pulse length (unrealistically) large instead.
image

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,151 @@
...
+    "  This may be zero (or half the pulse length) if the full pulse length is used, but choppers such as resolution choppers may extract a section of the pulse that is not aligned with the start of the full pulse.\n",
+    "  This offset can also be used to take into account a potential difference between the timing system's definition of the pulse time and the actual beginning of the neutron pulse exiting, e.g., the moderator.\n",
+    "- The black solid line within the first pulse (blue) indicates a neutron detected at $T_0^{i+1} + \\Delta t$.\n",
+    "  $\\Delta t$ is the `event_time_offset` in an `NXevent_data` group.\n",
+    "  This value is recorded for every neutron and gives the offset from the latest (previous) `event_time_zero` ($T_0^j$), i.e., the time difference to the previous vertical dotted line.\n",
+    "\n",
+    "To compute the time-of-flight for a neutron, we need to identify which source pulse it orginated from.\n",
+    "Consider the shaded vertical band above, indicating the time during which arriving neutrons are associated with $T_0^{i+1}$.\n",
+    "For, e.g., detector 1 when then observe:\n",
```
jl-wynen, line 58:
```suggestion
    "For, e.g., detector 1 we then observe:\n",
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,151 @@
...
+    "\n",
+    "- First (small `event_time_offset` $\\Delta t$, to the left of the dashed black line) we see the slowest neutrons from N (in this case N=2) source pulses earlier.\n",
+    "- Then (larger `event_time_offset` $\\Delta t$, to the right of the dashed black line) we see the slowest neutrons from N-1 (in this case N-1=1) source pulses earlier.\n",
```
jl-wynen, line 61:
```suggestion
    "- Then (larger `event_time_offset` $\\Delta t$, to the right of the dashed black line) we see the fastest neutrons from N-1 (in this case N-1=1) source pulses earlier.\n",
```
?

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,151 @@
...
+    "- Typically there is is an intermediate region where no neutrons should be able to traverse the chopper cascade.\n",
+    "  Neutrons detected in this time interval must thus be background from other sources.\n",
+    "\n",
+    "To compute the time-of-flight we add an integer multiple of the frame length to `event_time_offset`.\n",
+    "Within a given frame (indicated above by a band between two dotted vertical lines, such as the grey shaded band) there is a *pivot time*:\n",
+    "Neutrons with `event_time_offset` *before* the pivot time originated one source frame *before* neutrons *after* the pivot time.\n",
```
jl-wynen, line 67: In principle, there could be more than 2 pulses in a frame, right? This would require chopping them narrowly and a long beamline. Is this excluded in all beamline designs?

SimonHeybrock: Are you referring to WFM?

jl-wynen: No. Event without that, pulses get wide enough over a long distance to have contributions from more than one in a frame. If they are then chopped up in a particular way, we can end up with non-overlapping regions of 3 or more pulses in a single frame.

SimonHeybrock: You mean, e.g., interleaving bands from different pulses? I don't think this is done anywhere.

jl-wynen: No, much simpler. Say you have three consecutive pulses and you take the slow neutrons from the first, medium fast neutron from the second, and fast neutron from the third. Those three sections could all be in the same frame.

SimonHeybrock: I don't think it is possible to design a reliable chopper cascade that does this.

SimonHeybrock: Well, maybe it is, but I don't think anyone does this?

docs/reference/frameunwrapping/__init__.py
```diff
@@ -0,0 +1,83 @@
+import scipp as sc
```
nvaytet, line 1: Add missing licence identifier & copyright?

docs/reference/frameunwrapping/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    ax.annotate(r'$T_0^i$',
+                xy=(x0.value, 0),
+                xytext=(x0.value - 20, 3),
+                arrowprops=props)
+    ax.annotate(r'$T_0^i+\Delta T_0$',
+                xy=(x1.value, 0),
+                xytext=(x1.value - 20, 5),
+                arrowprops=props)
+    ax.annotate(r'$T_0^{i+1}+t_{\mathrm{pivot}}(\mathrm{det1})$',
+                xy=(205, det1.value),
```
nvaytet, line 36: Would be nice if these numbers (`205` etc.) could also have units.
Could you make `diagram.annotate()`?

docs/reference/frameunwrapping/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+                xytext=(150, 10),
+                arrowprops=props)
+    ax.annotate(r'$T_0^{i+2}+t_{\mathrm{pivot}}(\mathrm{det2})$',
+                xy=(235, det2.value),
+                xytext=(220, 15),
+                arrowprops=props)
+
+    # Pivot line
+    ax.plot([x1.value, x1.value + 177], [0, 1.1 * det2.value],
```
nvaytet, line 45: Could this just be `diagram.addneutron()`?

src/scippneutron/tof/diagram.py
```diff
@@ -0,0 +1,108 @@
+import scipp as sc
```
nvaytet, line 1: license & copyright

src/scippneutron/tof/diagram.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @property
+    def frame_length(self):
+        return self._frame_length
+
+    def to_time(self, time):
+        return time.to(unit=self._time_unit)
+
+    def add_source_pulse(self, pulse_length=3.0 * sc.Unit('ms')):
+        ls = 'dotted'
```
nvaytet, line 23: Can you make the styling part of the arguments, if you want to change to a dashed line?

src/scippneutron/tof/diagram.py
```diff
@@ -0,0 +1,108 @@
...
+    def add_neutron(self,
+                    *,
+                    time_offset: sc.Variable,
+                    wavelength: sc.Variable,
+                    L: sc.Variable,
+                    label=None):
+        tof = self.to_time(_tof_from_wavelength(wavelength=wavelength, Ltotal=L))
+        t0 = self.to_time(time_offset).value
+        self._ax.plot([t0, tof.value], [0, L.value],
+                      marker='',
```
nvaytet, line 47: Same as above regarding styling?

tests/tof/frames_test.py
```diff
@@ -0,0 +1,239 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+import pytest
+import numpy as np
+import scipp as sc
+from scippneutron.tof import frames
+
+
+def array(*,
```
nvaytet, line 10: better name than `array`, which is close to `sc.array`? e.g. `make_data_array()`?

docs/reference/frame-unwrapping.ipynb
```diff
@@ -0,0 +1,151 @@
...
+   "id": "aaf4cac3-6a54-44a2-95f7-a0518474589e",
+   "metadata": {},
+   "source": [
+    "In the figure above the index `i` labels source pulses.\n",
+    "We define:\n",
+    "\n",
+    "- $T_0^i$ is the `event_time_zero` recorded in an `NXevent_data` group.\n",
+    "  These times are indicated by the vertical dotted lines.\n",
+    "- $T_0^{i+1} = T_0^i + L_0$ where the frame length $L_0$ is defined by $L_0 = 1/f_0$, given a source frequency $f_0$.\n",
+    "- $\\Delta T_0$ is the offset from $T_0^i$ at which the neutrons are \"emitted\".\n",
```
nvaytet, line 49: How do you actually know what `Delta_T0` for a given neutron is during an experiment?
And don't you need to know that to compute your `tof`?
Should that be added as an uncertainty on the `tof` values?

docs/reference/frame-unwrapping.ipynb
jl-wynen: I'd still prefer to at least detect such a case and raise an exception. Just in case someone is doing something fancy.

docs/reference/frameunwrapping/__init__.py
SimonHeybrock: That has a different line style and would be harder to distinguish?

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: `Delta_T0` would likely be determined from chopper phases, similar to how you do it for WFM. Yes, it is needed to compute `tof`.

Not sure I get your question regarding uncertainty? It is an offset, not an uncertainty.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: How would you detect this?

jl-wynen: From the chopper settings, I guess.

SimonHeybrock: Parsing the chopper logs is unrelated to this functionality.

jl-wynen: Sure. So you would check for this when loading the choppers and warn / raise if the setup is unsupported ?

SimonHeybrock: I actually slightly doubt that such a check is feasible in practice? You would need to look at all the cutouts of all the choppers and do some sort of ray tracing? But I may be wrong and it could be simpler than I imagine.

docs/reference/frame-unwrapping.ipynb
nvaytet: cf the notebook Introduction to WFM: https://scipp.github.io/ess/techniques/wfm/introduction-to-wfm.html

with a long pulse, you can have 2 neutrons with different speeds originating from a different `Delta_T0` hitting a detector at the same time.

This is the whole reason for doing WFM, because you don't know the `Delta_T0`.
and when you compute the `tof`, it is an approximation, because you don't actually know the exact time the neutron started its journey down the beamline.
By using WFM choppers, you reduce this uncertainty.

Basically the time the neutron left the source would be `[T_0 + (0.5 * pulse_length)] +/- 0.5 * pulse_length` ?

docs/reference/frameunwrapping/__init__.py
nvaytet: but now you added the possibility to change the neutron line style?

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: Depends. Many instrument have resolution choppers that select part of the raw pulse, so the uncertainty can be (much) less. But of course some "uncertainty" remains.

If someone plans to track instrument-resolution via uncertainties of the coordinate that is likely a bigger undertaking that requires more thought? Pulse-length is far from the only contributor.

nvaytet (approved):

PR #245, opened 2022-01-13
Move to github workflows

SimonHeybrock:


lib/cmake/scipp-test.cmake
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# ~~~
+function(scipp_test target prefix)
+  if(${CTEST_DISCOVER_TESTS})
+    gtest_discover_tests(${target} TEST_PREFIX scipp/${prefix}/)
+  else()
+    add_test(NAME ${target} COMMAND ${target} --gtest_brief=1)
+  endif()
+endfunction()
```
jl-wynen, line 11: Could this conflict with the file in scipp? I.e. does scipp install this file?

jl-wynen (commented):
I didn't read the files in detail. Are they just copied over from scipp or is there something in particular I should pay attention to?

lib/cmake/scipp-test.cmake
SimonHeybrock: No this is not installed.

SimonHeybrock:
> I didn't read the files in detail. Are they just copied over from scipp or is there something in particular I should pay attention to?

Copied, with minor edits. Will wait until #239 is merged and rebase.

lib/neutron/test/CMakeLists.txt
```diff
@@ -14,3 +14,9 @@ set_property(
   TARGET ${TARGET_NAME} PROPERTY EXCLUDE_FROM_ALL $<NOT:$<BOOL:${FULL_BUILD}>>
 )
 scipp_test(${TARGET_NAME} neutron)
+
+if (WIN32)
+    # Cmake uses ; as a separator and won't set the path correctly unless we manually escape the path first.
+    STRING(REPLACE ";" "\\;" PATH_STRING "$ENV{PATH}")
+    set_tests_properties(scippneutron-test PROPERTIES ENVIRONMENT "PATH=${PATH_STRING}\;$ENV{CONDA_PREFIX}\\Lib\\site-packages\\scipp")
+endif(WIN32)
```
SimonHeybrock, line 28: @Tom-Willemsen Can this be done in a nicer manner using https://cmake.org/cmake/help/git-stage/command/cmake_path.html?

jl-wynen (approved):
Seems to work. Let's test a release

lib/neutron/test/CMakeLists.txt
Tom-Willemsen: I couldn't see a way to do this without manually escaping the semicolons using string replace... the cmake_path commands you linked don't help with this particular issue.

SimonHeybrock: I presume you did look at https://cmake.org/cmake/help/git-stage/command/cmake_path.html#native-path?

Tom-Willemsen: Yes, I did - but it doesn't do any semicolon escaping, which means that when we use the value in `set_tests_properties` it is interpreted as a cmake list rather than a string, which fails.

SimonHeybrock: 👍

PR #247, opened 2022-01-17
Add channels to conda meta.yml

jl-wynen:
They weren't in the original meta.yml used on Azure. But the error in the 0.4.2 package build looks like conda build fails to find scipp and mantid-framework. Is this the right fix?

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -13,7 +13,7 @@ requirements:
     - git
     - ninja
     - python {{ python }}
-    - scipp {{ scipp }}
+    - scipp::scipp {{ scipp }}
     - tbb-devel
     # doc build requirements
     - appdirs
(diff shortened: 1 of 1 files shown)


conda/meta.yaml
```diff
@@ -13,7 +13,7 @@ requirements:
     - git
     - ninja
     - python {{ python }}
-    - scipp {{ scipp }}
+    - scipp::scipp {{ scipp }}
```
SimonHeybrock, line 16: I think Owen advised against hard-coding channels in `meta.yaml`. Instead, they should be set when running `conda build`. That is, change the workflow script instead?

jl-wynen: I see. That is why it broke now. The channels were added in the Azure pipeline.

PR #249, opened 2022-01-17
Lower-level Nexus access utilities

SimonHeybrock:
This adds `nexus.Group`, which can used as follows:

```python
from scippneutron import nexus
from scippneutron.nexus import NX_class
f = nexus.File(filename)  # should use contextmanager, but that is cumbersome in a Jupyter notebook

f.by_nx_class()[NX_class.NXlog]['SampleTemp']  # nexus.Group (wrapping h5py.Group), nothing loaded yet
f.by_nx_class()[NX_class.NXlog]['SampleTemp']['time', :3]
f.by_nx_class()[NX_class.NXlog]['proton_charge'][:1000]  # dim label redundant if single dim
f.by_nx_class()[NX_class.NXmonitor]['monitor1'][...]  # load everything
f.by_nx_class()[NX_class.NXevent_data]['bank42_events']['pulse', :20000]
```

The implementation right now is a draft. I am considering the following changes:

- `nexus.Group` implementation has cumbersome branching bases on `NX_class`. My current idea would be to turn this into a base-class, `NXobject`, which can be implemented, e.g., by `NXlog`, `NXmonitor`, and `NXevent_data`.
- The child classes provide properties such as `dims, `shape`, and `unit`.
- Child classes provide `__getitem__` to actually load the class. This would forward to the current internal functions for loading such groups.

Does it make sense to introduce such an inheritance-based design? Or is there a better option?

### DIFF
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -86,9 +86,8 @@ def to_plain_index(dims, select, ignore_missing=False):
     """
     def check_1d():
         if len(dims) != 1:
-            raise ValueError(
-                f"Dataset has multiple dimensions {dims}, specify the dimension to index."
-            )
+            raise ValueError(f"Dataset has multiple dimensions {dims}, "
+                             "specify the dimension to index.")
 
     if select is Ellipsis:
         return select
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -8,6 +8,7 @@ from scipp.spatial import affine_transform, linear_transform, \
     rotation, translation
 import scipp as sc
 import numpy as np
+from typing import Dict
 
 
 class BadSource(Exception):
@@ -42,6 +43,7 @@ class JSONGroup(dict):
         self.file = file
 
 
+Dataset = Union[h5py.Dataset, Dict]
 Group = Union[h5py.Group, JSONGroup]
 
 
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -76,3 +76,38 @@ def _add_attr_to_loaded_data(attr_name: str,
             data[attr_name] = sc.scalar(value=value, unit=unit)
     except KeyError:
         pass
+
+
+def to_plain_index(dims, select, ignore_missing=False):
+    """
+    Given a valid "scipp" index 'select', return an equivalent plain numpy-style index.
+    """
+    def check_1d():
+        if len(dims) != 1:
+            raise ValueError(
+                f"Dataset has multiple dimensions {dims}, specify the dimension to index."
+            )
+
+    if select is Ellipsis:
+        return select
+    if isinstance(select, tuple) and len(select) == 0:
+        return select
+    if isinstance(select, tuple) and isinstance(select[0], str):
+        key, sel = select
+        select = {key: sel}
+
+    if isinstance(select, tuple):
+        check_1d()
+        return select
+    elif isinstance(select, int) or isinstance(select, slice):
+        check_1d()
+        return select
+    elif isinstance(select, dict):
+        index = [slice(None)] * len(dims)
+        for key, sel in select.items():
+            if not ignore_missing and key not in dims:
+                raise ValueError(
+                    f"'{key}' used for indexing not found in dataset dims {dims}.")
+            index[dims.index(key)] = sel
+        return tuple(index)
+    raise ValueError("Cannot process index {select}.")
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -208,7 +208,7 @@ def _load_event_group(group: Group,
     _check_for_missing_fields(group, nexus)
 
     def shape(name):
-        return nexus.get_shape(nexus.get_dataset_from_group(group,name))
+        return nexus.get_shape(nexus.get_dataset_from_group(group, name))
 
     max_index = shape("event_index")[0]
     single = False
@@ -219,12 +219,18 @@ def _load_event_group(group: Group,
         # TODO unused key, should this take plain indices, like load_dataset?
         key, index = select
         if isinstance(index, int):
-            index = slice(index, None if index + 2 > max_index else index + 2)
             single = True
-        elif isinstance(index.stop, int):
-            index = slice(index.start,
-                          None if index.stop + 1 > max_index else index.stop + 1)
-        last_loaded = isinstance(index.stop, int)
+            start, stop, _ = slice(index, None).indices(max_index)
+            if start == stop:
+                raise IndexError('')
+            index = slice(start, start + 1)
+        start, stop, stride = index.indices(max_index)
+        if stop + stride > max_index:
+            last_loaded = False
+        else:
+            stop += stride
+            last_loaded = True
+        index = slice(start, stop, stride)
 
     event_index = nexus.load_dataset_from_group_as_numpy_array(
         group, "event_index", index)
@@ -234,7 +240,8 @@ def _load_event_group(group: Group,
     event_index[event_index < 0] = num_event
 
     if len(event_index) > 0:
-        event_select = slice(event_index[0], event_index[-1] if last_loaded else num_event)
+        event_select = slice(event_index[0],
+                             event_index[-1] if last_loaded else num_event)
     else:
         event_select = slice(None)
 
@@ -295,7 +302,7 @@ def _load_event_group(group: Group,
 
     if not quiet:
         print(f"Loaded {len(event_id)} events from "
-              f"{group.name} containing {number_of_event_ids} events")
+              f"{group.name} containing {num_event} events")
 
     return detector_data
 
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -210,7 +210,7 @@ def _load_event_group(group: Group,
     def shape(name):
         return nexus.get_shape(nexus.get_dataset_from_group(group,name))
 
-    max_index = shape("event_time_offset")[0]
+    max_index = shape("event_index")[0]
     single = False
     if select is Ellipsis:
         index = select
@@ -230,11 +230,11 @@ def _load_event_group(group: Group,
         group, "event_index", index)
     pulse_times = _load_pulse_times(group, nexus, index)
 
-    number_of_event_ids = shape("event_id")[0]
-    event_index[event_index < 0] = number_of_event_ids
+    num_event = shape("event_time_offset")[0]
+    event_index[event_index < 0] = num_event
 
     if len(event_index) > 0:
-        event_select = slice(event_index[0], event_index[-1] if last_loaded else max_index)
+        event_select = slice(event_index[0], event_index[-1] if last_loaded else num_event)
     else:
         event_select = slice(None)
 
@@ -266,7 +266,7 @@ def _load_event_group(group: Group,
         detector_data.detector_ids.dtype, event_id.dtype)
 
     if not last_loaded:
-        event_index = np.append(event_index, number_of_event_ids)
+        event_index = np.append(event_index, num_event)
 
     event_index = sc.array(dims=[_pulse_dimension],
                            values=event_index,
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -226,6 +226,7 @@ class NXevent_data(NXobject):
         data = detector_data.event_data
         data.bins.coords['id'] = data.bins.coords.pop('detector_id')
         data.bins.coords['time_offset'] = data.bins.coords.pop('tof')
+        data.coords['time_zero'] = data.coords.pop('pulse_time')
         return data
 
 
@@ -266,6 +267,8 @@ def _load_event_group(group: Group,
     pulse_times = _load_pulse_times(group, nexus, index)
 
     num_event = shape("event_time_offset")[0]
+    # Some files contain uint64 "max" indices, which turn into negatives during
+    # conversion to int64. This is a hack to get arround this.
     event_index[event_index < 0] = num_event
 
     if len(event_index) > 0:
@@ -274,8 +277,6 @@ def _load_event_group(group: Group,
     else:
         event_select = slice(None)
 
-    # Some files contain uint64 "max" indices, which turn into negatives during
-    # conversion to int64. This is a hack to get arround this.
     event_id = nexus.load_dataset(group,
                                   "event_id", [_event_dimension],
                                   index=event_select)
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -16,8 +16,8 @@ from ._nexus import LoadFromNexus
 _bank_dimension = "bank"
 _detector_dimension = "detector_id"
 _event_dimension = "event"
-_pulse_dimension = "pulse"
-_pulse_time = "pulse_time"
+_pulse_dimension = "time"
+_pulse_time = "time"
 _time_of_flight = "tof"
 
 
@@ -149,12 +149,13 @@ def _create_empty_events_data_array(tof_dtype: Any = np.int64,
                         })
 
 
-def _load_pulse_times(group: Group, nexus: LoadFromNexus) -> sc.Variable:
+def _load_pulse_times(group: Group, nexus: LoadFromNexus, index=...) -> sc.Variable:
     time_zero_group = "event_time_zero"
 
     event_time_zero = nexus.load_dataset(group,
                                          time_zero_group,
-                                         dimensions=[_pulse_dimension])
+                                         dimensions=[_pulse_dimension],
+                                         index=index)
 
     try:
         pulse_times = sc.to_unit(event_time_zero, sc.units.ns, copy=False)
@@ -199,19 +200,46 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
     return DetectorData(detector_ids=detector_ids, pixel_positions=pixel_positions)
 
 
-def _load_event_group(group: Group, nexus: LoadFromNexus, detector_data: DetectorData,
-                      quiet: bool) -> DetectorData:
+def _load_event_group(group: Group,
+                      nexus: LoadFromNexus,
+                      detector_data: DetectorData,
+                      quiet: bool,
+                      select=...) -> DetectorData:
     _check_for_missing_fields(group, nexus)
 
-    event_id = nexus.load_dataset(group, "event_id", [_event_dimension])
-    number_of_event_ids = event_id.sizes[_event_dimension]
-    event_index = nexus.load_dataset_from_group_as_numpy_array(group, "event_index")
-    # Some files contain uint64 "max" indices, which turn into negatives during
-    # conversion to int64. This is a hack to get arround this.
+    max_index = group["event_id"].shape[0]
+    single = False
+    if select is Ellipsis:
+        index = select
+        last_loaded = False
+    else:
+        key, index = select
+        if isinstance(index, int):
+            index = slice(index, None if index + 2 > max_index else index + 2)
+            single = True
+        elif isinstance(index.stop, int):
+            index = slice(index.start,
+                          None if index.stop + 1 > max_index else index.stop + 1)
+        last_loaded = isinstance(index.stop, int)
+
+    event_index = nexus.load_dataset_from_group_as_numpy_array(
+        group, "event_index", index)
+    pulse_times = _load_pulse_times(group, nexus, index)
+
+    number_of_event_ids = max_index
     event_index[event_index < 0] = number_of_event_ids
 
-    event_time_offset = nexus.load_dataset(group, "event_time_offset",
-                                           [_event_dimension])
+    event_select = slice(event_index[0], event_index[-1] if last_loaded else max_index)
+
+    # Some files contain uint64 "max" indices, which turn into negatives during
+    # conversion to int64. This is a hack to get arround this.
+    event_id = nexus.load_dataset(group,
+                                  "event_id", [_event_dimension],
+                                  index=event_select)
+
+    event_time_offset = nexus.load_dataset(group,
+                                           "event_time_offset", [_event_dimension],
+                                           index=event_select)
 
     # Weights are not stored in NeXus, so use 1s
     weights = sc.ones(dims=[_event_dimension],
@@ -230,31 +258,36 @@ def _load_event_group(group: Group, nexus: LoadFromNexus, detector_data: Detecto
         event_id.dtype if detector_data.detector_ids is None else
         detector_data.detector_ids.dtype, event_id.dtype)
 
+    if not last_loaded:
+        event_index = np.append(event_index, number_of_event_ids)
+
     event_index = sc.array(dims=[_pulse_dimension],
                            values=event_index,
                            dtype=sc.DType.int64)
-    pulse_times = _load_pulse_times(group, nexus)
 
-    begins = event_index
+    event_index -= event_index.min()
+
     # There is some variation in the last recorded event_index in files from different
     # institutions. We try to make sure here that it is what would be the first index of
     # the next pulse. In other words, ensure that event_index includes the bin edge for
     # the last pulse.
-    ends = sc.concat(
-        [event_index[_pulse_dimension, 1:],
-         sc.scalar(number_of_event_ids)], _pulse_dimension)
+    begins = event_index[_pulse_dimension, :-1]
+    ends = event_index[_pulse_dimension, 1:]
+    if single:
+        begins = begins[_pulse_dimension, 0]
+        ends = ends[_pulse_dimension, 0]
 
     try:
         binned = sc.bins(data=events, dim=_event_dimension, begin=begins, end=ends)
     except sc.SliceError:
         raise BadSource(f"Event index in NXEvent at {group.name}/event_index was not"
-                        f"ordered. The index must be ordered to load pulse times.")
+                        f" ordered. The index must be ordered to load pulse times.")
 
     detector_data.event_data = sc.DataArray(data=binned,
                                             coords={"pulse_time": pulse_times})
 
     if not quiet:
-        print(f"Loaded event data from "
+        print(f"Loaded {len(event_id)} events from "
               f"{group.name} containing {number_of_event_ids} events")
 
     return detector_data
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -13,6 +13,7 @@ import scipp as sc
 from warnings import warn
 from ._transformations import get_full_transformation_matrix
 from ._nexus import LoadFromNexus
+from .nxobject import NXobject
 
 _bank_dimension = "bank"
 _detector_dimension = "detector_id"
@@ -199,6 +200,25 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
         pixel_positions = _load_pixel_positions(group, detector_ids.shape[0], nexus)
 
     return DetectorData(detector_ids=detector_ids, pixel_positions=pixel_positions)
+
+
+class NXevent_data(NXobject):
+    def __init__(self, group: Group, loader: LoadFromNexus):
+        super().__init__(group, loader)
+
+    @property
+    def shape(self):
+        return self._loader.get_shape(
+            self._loader.get_dataset_from_group(self._group, "event_index"))
+
+    @property
+    def dims(self):
+        return [_pulse_dimension]
+
+    @property
+    def unit(self):
+        # Binned data, bins do not have a unit
+        return sc.units.one
 
 
 def _load_event_group(group: Group,
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -16,8 +16,8 @@ from ._nexus import LoadFromNexus
 _bank_dimension = "bank"
 _detector_dimension = "detector_id"
 _event_dimension = "event"
-_pulse_dimension = "time"
-_pulse_time = "time"
+_pulse_dimension = "pulse"
+_pulse_time = "pulse_time"
 _time_of_flight = "tof"
 
 
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -207,12 +207,16 @@ def _load_event_group(group: Group,
                       select=...) -> DetectorData:
     _check_for_missing_fields(group, nexus)
 
-    max_index = group["event_index"].shape[0]
+    def shape(name):
+        return nexus.get_shape(nexus.get_dataset_from_group(group,name))
+
+    max_index = shape("event_index")[0]
     single = False
     if select is Ellipsis:
         index = select
         last_loaded = False
     else:
+        # TODO unused key, should this take plain indices, like load_dataset?
         key, index = select
         if isinstance(index, int):
             index = slice(index, None if index + 2 > max_index else index + 2)
@@ -226,10 +230,13 @@ def _load_event_group(group: Group,
         group, "event_index", index)
     pulse_times = _load_pulse_times(group, nexus, index)
 
-    number_of_event_ids = group["event_id"].shape[0]
+    number_of_event_ids = shape("event_id")[0]
     event_index[event_index < 0] = number_of_event_ids
 
-    event_select = slice(event_index[0], event_index[-1] if last_loaded else max_index)
+    if event_index:
+        event_select = slice(event_index[0], event_index[-1] if last_loaded else max_index)
+    else:
+        event_select = slice(None)
 
     # Some files contain uint64 "max" indices, which turn into negatives during
     # conversion to int64. This is a hack to get arround this.
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -229,7 +229,6 @@ class NXevent_data(NXobject):
         return data
 
 
-
 def _load_event_group(group: Group,
                       nexus: LoadFromNexus,
                       detector_data: DetectorData,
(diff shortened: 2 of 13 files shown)


src/scippneutron/nexus/nexus.py
```diff
@@ -0,0 +1,91 @@
...
+    NXevent_data = auto()
+
+
+class Group:
+    def __init__(self, group: h5py.Group):
+        self._group = group
+        self._loader = LoadFromHdf5()
+
+    def __getattr__(self, name):
+        return getattr(self._group, name)
```
jl-wynen, line 29: Seems a bit odd. What is this used for?

src/scippneutron/nexus/nexus.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class NX_class(Enum):
+    NXentry = auto()
+    NXlog = auto()
+    NXmonitor = auto()
+    NXevent_data = auto()
+
+
+class Group:
```
jl-wynen, line 23: Putting this here so we have a dedicated thread.

> nexus.Group implementation has cumbersome branching bases on NX_class

We need this branching somewhere. Are you saying it is cumbersome because it needs to be repeated in several methods?

> Child classes provide __getitem__ to actually load the class. This would forward to the current internal functions for loading such groups.

So `NXobject.__getitem__` would return an instance one of those child classes if the underlying HDF5 object is not a group? I think this makes sense.

> Does it make sense to introduce such an inheritance-based design? Or is there a better option?

It would certainly make sense for establishing a common interface. At least until we can use https://docs.python.org/3/library/typing.html#typing.Protocol (Introduced in Python 3.8. Can't be long until we drop support for 3.7...)
But what else would the base class contain? Default implementations of `dims` etc. that raise an exception? I m not sure there is much value in using inheritance here unless there are methods which actually do something that can be reused by the child classes.

src/scippneutron/nexus/nexus.py
SimonHeybrock: To forward to methods of `h5py.Group`, for things we are not implementing ourselves. Might not be what we want for the final product, but useful while implementation is still incomplete.

src/scippneutron/nexus/nexus.py
SimonHeybrock: > We need this branching somewhere. Are you saying it is cumbersome because it needs to be repeated in several methods?

Yes.

> So NXobject.__getitem__ would return an instance one of those child classes if the underlying HDF5 object is not a group? I think this makes sense.

Rather: Would return an instance of the HDF5 object is a group with known `NX_class` attribute. The `NXobject` class map to HDF5 groups, not datasets. But yes otherwise.

> But what else would the base class contain? Default implementations of dims etc. that raise an exception? I m not sure there is much value in using inheritance here unless there are methods which actually do something that can be reused by the child classes.

I think there will likely be a number of common methods, e.g., simply for tree traversal. Basically everything that is in `h5py.Group` would be in the base class. Will see while implementing if, e.g., composition makes more sense.

jl-wynen: > Rather: Would return an instance of the HDF5 object is a group with known NX_class attribute. The NXobject class map to HDF5 groups, not datasets. But yes otherwise.

I don't understand. Where do the derived classes come in?

SimonHeybrock: As you say: You access a group entry, if it is an `h5py.Group` (not a `h5py.Dataset`) then it should usually have a `NX_class` attribute, and instead of returning the base class `NXobject` (wrapping `h5py.Group`) we return, e.g., an `NXlog` instance if `NX_class == 'NXlog'`.

jl-wynen: Ok. Then I misunderstood your earlier reply.

I would say that inheritance is a good solution at first order. Like you said, we should see what the base class would actually be like.

Tom-Willemsen: I think I agree, the inheritance-based solution sounds ok to me, but I think we need to see what the code would actually look like to make more detailed comments...

Tom-Willemsen:
Perhaps I've not understood some aspect of the API, but why do we want:

```
f.by_nx_class()[NX_class.NXlog]['SampleTemp']
```

rather than the more natural looking

```
f.by_nx_class(NX_class.NXlog)['SampleTemp']
```

?

jl-wynen:
We could also 'overload' `__getitem__` and use
```.py
f[NX_class.NXlog]['SampleTemp']
```
Since the argument of the first brackets is of type `NX_class`, we could identify what the user means and call `by_nx_class` internally.
@SimonHeybrock, this would correspond to your initial suggestion, I think. Why did you move away from it?

SimonHeybrock:
> We could also 'overload' `__getitem__` and use
> 
> ```python
> f[NX_class.NXlog]['SampleTemp']
> ```
> 
> Since the argument of the first brackets is of type `NX_class`, we could identify what the user means and call `by_nx_class` internally. @SimonHeybrock, this would correspond to your initial suggestion, I think. Why did you move away from it?



> We could also 'overload' `__getitem__` and use
> 
> ```python
> f[NX_class.NXlog]['SampleTemp']
> ```
> 
> Since the argument of the first brackets is of type `NX_class`, we could identify what the user means and call `by_nx_class` internally. @SimonHeybrock, this would correspond to your initial suggestion, I think. Why did you move away from it?

`by_nx_class` does a recursive walk of the tree, it felt odd that `__getitem__` would operate recursively.

SimonHeybrock:
@jl-wynen @Tom-Willemsen Please have a look at the new `nxobject.py`, and subclasses of `NXobject` that I added. Essentially some of the lower-level functions become implementation details of the classes (and they could each and individually be refactored to do this in a nicer way than the current plain wrapper).

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,74 @@
...
+    NXmonitor = auto()
+    NXevent_data = auto()
+
+
+class NXobject:
+    _registry = {}
+
+    def __init_subclass__(cls):
+        super().__init_subclass__()
+        cls._registry[cls.__name__] = cls
```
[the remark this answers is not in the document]
SimonHeybrock: > to the nexus class names from a data file. 

What data file are you referring to?

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,74 @@
...
+            names = [self._loader.get_name(group) for group in groups]
+            if len(names) != len(set(names)):  # fall back to full path if duplicate
+                names = [group.name for group in groups]
+            out[NX_class[nx_class]] = {n: self.make(g) for n, g in zip(names, groups)}
+        return out
+
+    @property
+    def NX_class(self):
+        nx_class = self._loader.get_string_attribute(self._group, 'NX_class')
+        return NX_class[nx_class]
```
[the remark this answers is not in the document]
SimonHeybrock: I had considered getting rid of the enum, using the actual classes as dict keys. That works, but when printing them, the class names are displayed in full, including their "path" within the package and submodules, which was hard to read. If I moved them, e.g., to `scippneutron.nexus` it would be reasonable, maybe? Or do you know of a way to change the `__repr__` of a type?

src/scippneutron/file_loading/_common.py
```diff
@@ -76,3 +76,38 @@ def _add_attr_to_loaded_data(attr_name: str,
...
+        check_1d()
+        return select
+    elif isinstance(select, dict):
+        index = [slice(None)] * len(dims)
+        for key, sel in select.items():
+            if not ignore_missing and key not in dims:
+                raise ValueError(
+                    f"'{key}' used for indexing not found in dataset dims {dims}.")
+            index[dims.index(key)] = sel
+        return tuple(index)
```
jl-wynen, line 112: We don't support indexing by dict elsewhere. I think we should either not do it here either or support it in Variable and DataArray.

src/scippneutron/file_loading/nxobject.py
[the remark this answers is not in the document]
jl-wynen: `make` uses the content of the 'NX_class' attribute to look up the subclass which is registered by its name. So the name of the subclass is coupled to how the nexus class is stored in the file.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,74 @@
...
+    def __getattr__(self, name):
+        return getattr(self._group, name)
+
+    def __getitem__(self, index):
+        if isinstance(index, str):
+            item = self._group[index]
+            if hasattr(item, 'visititems'):
+                return self.make(item)
+            else:
+                return item
```
jl-wynen, line 49: Does this return `h5.Dataset` objects? I would have expected that we wrap them.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,74 @@
...
+        return self._getitem(index)
+
+    def _getitem(self, index):
+        # TODO Is it better to fall back to returning h5py.Group?
+        # distinguish classes not implementing _getitem, vs missing classes!
+        print(f'Cannot load unsupported class {self.NX_class}')
+
+    @functools.lru_cache()
+    def by_nx_class(self):
+        classes = self._loader.find_by_nx_class(tuple(self._registry), self._group)
```
jl-wynen, line 59: This behaves differently depending on what modules were imported because that can influence the registry. We have to make sure that we always import all modules containing subclasses of `NX_object`.
That is, this approach seems fragile.

src/scippneutron/file_loading/nxobject.py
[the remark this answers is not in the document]
jl-wynen: > Or do you know of a way to change the __repr__ of a type?

This would make it hard to analyse and debug some code. And I don't think it is possible. At least not without a custom metaclass.

src/scippneutron/file_loading/nxobject.py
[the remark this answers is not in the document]
SimonHeybrock: Well, not the file, but by which `NX_class` values the NeXus standard defines? Not sure I understand the comments, can you elaborate?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Well, Nexus does not really define much for "fields" (their name for HDF5 datasets), so such a class would have limited use, apart from hiding differences between and underlying HDF5 vs. Json, which is not the main aim of this PR. We can't do much with an individual dataset (it cannot be loaded as a scipp object). Many NXobject subclasses map to `sc.DataArray`, but Nexus fields have insufficient information to load them, e.g., as a `sc.Variable`.

But a wrapper could certainly be added later if required?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: If it is not imported in cannot be created either? Not sure how that could be done differently.

src/scippneutron/file_loading/_common.py
SimonHeybrock: We have to support it here, since we cannot use `group['x', 4]['y', 5]`, since the first `__getitem__` would already load the entire column.

SimonHeybrock: But yes, the plan is to support this in scipp sooner or later.

src/scippneutron/file_loading/nxobject.py
jl-wynen: Why even support loading it at all if it is not really useful?

src/scippneutron/file_loading/nxobject.py
jl-wynen: We could create the registry in a specific file so that it is always fully available if that file is imported. But this would likely mean moving the subclass construction logic out of the base class. Which I am in favour of anyway.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: I am not saying that it is not useful, just that there is nothing/little Nexus-specific about it, so users could fall back to plain h5py.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Doesn't this lead to some cyclic import issue?

jl-wynen: We could have `nx_registry.py` with `registry=None` and in `__init__.py` import the various `NX_xxx` classes and do `nx_registry.registry = [...]`. `NX_object` could then use `nx_registry.py` to instantiate a concrete object because that module never imports anything itself.

src/scippneutron/file_loading/nxobject.py
jl-wynen: Than I would still wrap it in order to generalise over HDF5 and JSON. A user can then use a property to get the underlying `h5.Dataset` or json dict. This would make it more explicit what is happening.

src/scippneutron/file_loading/nxobject.py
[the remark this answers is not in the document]
jl-wynen: The names of the Python classes must be the same as the `NX_class` names in NeXus. We could decouple the two by using separately constructed strings to look up the registry. E.g. `cls._registry[cls.NX_class] = cls` (assuming a class property `NX_class`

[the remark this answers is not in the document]
SimonHeybrock: > The names of the Python classes must be the same as the NX_class names in NeXus.

Yes, isn't that good?

> We could decouple the two by using separately constructed strings to look up the registry

We could, but why? Isn't that adding confusion?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: @jl-wynen I switched to a manual registry, is this along the lines of what you wanted?

jl-wynen: Not quite. This has a circular dependency between the modules. It only avoids problems by delaying the import. I was thinking of moving the registry into a module for `NX_object` to depend on. (Or leaving it in the same module, that part is not so important.) And populating the registry in a place that depends on `NX_object`, `NX_log`, etc. This way, there are no circular dependencies between modules. A bit like your original solution but with a centralised code for populating the registry.

SimonHeybrock: Sounds more complicated. Is it really that much better?

jl-wynen: Not sure. I think it would be cleanest because it has only one way dependencies between modules. But the current solution should work just fine as long as no one calls `_nx_class_registry` during import.

SimonHeybrock:
I am encountering some weirdness with the existing implementation:

- `group.name` is used in some cases, which works only if `group` is either a `h5py.Group` or a `JSONGroup`.
  - Some operations with `LoadFromJson` do not appear to return a `JSONGroup`, so we end up with an exception (previously untested code path?). I replaced this with a `get_path` method, which for now just returns the current name in `LoadFromJson`. In this particular case it only affects a log message.
- `find_by_nx_class` included the current group for `LoadFromHdf5` but not in `LoadFromJson`. I changed the former, tests pass, but I don't know if I broke loading of some files?

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,132 @@
...
+class Field:
+    """NeXus field.
+
+    In HDF5 fields are represented as dataset.
+    """
+    def __init__(self, dataset: Dataset, loader: LoadFromNexus = LoadFromHdf5()):
+        self._dataset = dataset
+        self._loader = loader
+
+    def __getitem__(self, index):
```
jl-wynen, line 29: It would be good to have more type annotations.
(Putting the comment here because this is the first place where I noticed it. There are a bunch more.)

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,132 @@
...
+
+class NX_class(Enum):
+    NXroot = auto()
+    NXentry = auto()
+    NXlog = auto()
+    NXmonitor = auto()
+    NXevent_data = auto()
+
+
+class Field:
```
jl-wynen, line 20: In `h5py`, groups and datasets have a ref to their file. This can be useful for navigation and comprehensive error messages. I think it would be a good idea to have something similar in all nexus objects.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,132 @@
...
+    def name(self) -> str:
+        return self._loader.get_path(self._dataset)
+
+    @property
+    def shape(self):
+        return self._loader.get_shape(self._dataset)
+
+    @property
+    def unit(self):
+        # TODO Prefer to return None if no such attr, provided that scipp supports this
```
jl-wynen, line 49: Add a test to make sure we don't forget when scipp is updated?

src/scippneutron/file_loading/nxobject.py
jl-wynen, line 20: Is `Field` not a subtype of `NXobject` in NeXus? That is, are `NXobjects` always HDF5 groups?

src/scippneutron/file_loading/nxobject.py
```diff
@@ -0,0 +1,132 @@
...
+            names = [self._loader.get_name(group) for group in groups]
+            if len(names) != len(set(names)):  # fall back to full path if duplicate
+                names = [group.name for group in groups]
+            out[NX_class[nx_class]] = {n: self._make(g) for n, g in zip(names, groups)}
+        return out
+
+    @property
+    def nx_class(self):
+        key = self._loader.get_string_attribute(self._group, 'NX_class')
+        return NX_class[key]
```
jl-wynen, line 104: How about turning this into a class attribute? It would mean all derived classes have to define it, but there no need to load it dynamically, right?

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -199,19 +202,88 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
...
+            self._loader.get_dataset_from_group(self._group, "event_index"))
+
+    @property
+    def dims(self):
+        return [_pulse_dimension]
+
+    @property
+    def unit(self):
+        # Binned data, bins do not have a unit
+        return sc.units.one
```
jl-wynen, line 218: Can we already return `None` here?

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -199,19 +202,88 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
...
+    def _getitem(self, index):
+        detector_data = _load_event_group(self._group,
+                                          self._loader,
+                                          DetectorData(),
+                                          quiet=False,
+                                          select=index)
+        data = detector_data.event_data
+        data.bins.coords['id'] = data.bins.coords.pop('detector_id')
+        data.bins.coords['time_offset'] = data.bins.coords.pop('tof')
+        data.coords['time_zero'] = data.coords.pop('pulse_time')
```
jl-wynen, line 229: What are these names based on?

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -199,19 +202,88 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
...
+
+
+def _load_event_group(group: Group,
+                      nexus: LoadFromNexus,
+                      detector_data: DetectorData,
+                      quiet: bool,
+                      select=...) -> DetectorData:
     _check_for_missing_fields(group, nexus)
+    index = to_plain_index([_pulse_dimension], select)
+    if isinstance(index, tuple):
```
jl-wynen, line 240: `index` can be an empty tuple. I think that should be equivalent to an ellipsis and not raise `IndexError`. Maybe `to_plain_index` should handle it in order to not complicate the code here.

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -199,19 +202,88 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
...
+            if start == stop:
+                raise IndexError('')
+            index = slice(start, start + 1)
+        start, stop, stride = index.indices(max_index)
+        if stop + stride > max_index:
+            last_loaded = False
+        else:
+            stop += stride
+            last_loaded = True
+        index = slice(start, stop, stride)
```
jl-wynen, line 263: This seems very complicated. Could it be simplified if `to_plain_index` returned a slice instead of int or ellipsis?
Also, it does not handle the case when index is a dict.

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -199,19 +202,88 @@ def _load_detector(group: Group, nexus: LoadFromNexus) -> DetectorData:
...
+    max_index = shape("event_index")[0]
+    single = False
+    if index is Ellipsis:
+        last_loaded = False
+    else:
+        if isinstance(index, int):
+            single = True
+            start, stop, _ = slice(index, None).indices(max_index)
+            if start == stop:
+                raise IndexError('')
```
jl-wynen, line 255: Message?

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -230,32 +302,42 @@ def _load_event_group(group: Group, nexus: LoadFromNexus, detector_data: Detecto
...
+
     # There is some variation in the last recorded event_index in files from different
     # institutions. We try to make sure here that it is what would be the first index of
     # the next pulse. In other words, ensure that event_index includes the bin edge for
     # the last pulse.
-    ends = sc.concat(
-        [event_index[_pulse_dimension, 1:],
-         sc.scalar(number_of_event_ids)], _pulse_dimension)
+    begins = event_index[_pulse_dimension, :-1]
+    ends = event_index[_pulse_dimension, 1:]
```
jl-wynen, line 323: Put into `else` block?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Yes, `NXobject` is always a group.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Erm, not sure: Thinks such as `NXmonitor` and `NXdetector` "kind of" inherit `NXdata` (though there is no actualy inheritance defined). I am just working on the former to create `NXdata` (or `NXevent_data`) to deal with concrete loading. Not sure if we want `nx_class` to refer the actual attribute from the file, or the current instance being used to process the group?

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Names from `NXevent_data`, without the `event_` prefix. Not sure this is a good choice, could keep the full name?

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: I don't think `to_plain_index` can simplify much here. The problem is that we need to deal with loading one more index (if possible), since `NXevent_data` does not define "end" indices. This is different from general slicing applications.

src/scippneutron/nexus/nexus.py
```diff
@@ -0,0 +1,17 @@
...
+
+from contextlib import AbstractContextManager
+import h5py
+
+from ..file_loading.nxobject import NXroot
+
+
+class File(AbstractContextManager, NXroot):
+    def __init__(self, *args, **kwargs):
+        self._file = h5py.File(*args, **kwargs)
```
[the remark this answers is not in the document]
SimonHeybrock: Not sure. How would you do it?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Have a ref to their file? Isn't that in the group, i.e., not need to duplicate?

src/scippneutron/nexus/nexus.py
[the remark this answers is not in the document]
jl-wynen:
```.py
def __enter__(self) -> File:
    self._file.__enter__()
    return self
```

src/scippneutron/file_loading/nxobject.py
jl-wynen: How would a user access it, though?

src/scippneutron/file_loading/nxobject.py
jl-wynen: Are you saying that there will be, .e.g, an `NXdata` object in python processing an `NXmonitor` group in HDF5?
It seems odd to then use `nx_class == NXmonitor` in this case because, using the `NXdata` object, the monitor-specific contents are not accessible. But I guess, I am not sure what `nx_class` is used for.

src/scippneutron/file_loading/_detector_data.py
jl-wynen: I think it makes the most sense if these low level functions load the data as faithfully as possible.

src/scippneutron/nexus/nexus.py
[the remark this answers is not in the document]
SimonHeybrock: But we are not using the `h5py.File` as context manager? What would calling `__enter__` achieve?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: We would add a `file` property, just like `h5py.Group` has one? Not sure I understand the question.

src/scippneutron/nexus/nexus.py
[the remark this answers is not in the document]
jl-wynen: Then why does it inherit from `AbstractContextManager`?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: We could probably just remove the `nx_class` property. I have added an `attrs` property in a follow up branch, so the user could just do `group.attrs['NX_class']` instead of `group.nx_class`. Should I remove this convenience property?

src/scippneutron/file_loading/nxobject.py
jl-wynen: I was referring to

> Isn't that in the group, i.e., not need to duplicate?

The information is present but not accessible by the user.

src/scippneutron/file_loading/nxobject.py
jl-wynen: Yes, I think that makes sense.

tests/test_nexus.py
```diff
@@ -0,0 +1,196 @@
...
+
+def test_nxobject_by_nx_class_contains_everything(nexus_group: Tuple[Callable,
+                                                                     LoadFromNexus]):
+    resource, loader = nexus_group
+    with resource(builder_with_events_monitor_and_log())() as f:
+        root = nexus.NXroot(f, loader)
+        classes = root.by_nx_class()
+        assert list(classes[nexus.NX_class.NXentry]) == ['entry']
+        assert list(classes[nexus.NX_class.NXmonitor]) == ['monitor']
+        assert list(classes[nexus.NX_class.NXlog]) == ['log']
```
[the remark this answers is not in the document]
SimonHeybrock: It can be used to load something with a know specific name, without having to manually search the entire tree structure. Someone may be interested in a specific log, but not know in which group and subgroup it is stored.

src/scippneutron/nexus/nexus.py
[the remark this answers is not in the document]
SimonHeybrock: Because it is implementing a context manager? But it is not using the `h5py.File` context manager. But maybe that is possible, I just don't know how?

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: The interface is *very* incomplete right now, there are potentially dozens of others. I suggest we add things when we need them.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Ran into an "issue" right away when trying to remove: Most files to not correctly set `NX_class` for the root (to `NXroot`), so the convenience property does actually add value here (avoiding special case handling in client code). I suggest we keep it?

tests/test_nexus.py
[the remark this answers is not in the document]
jl-wynen: Ah, I checked the implementation again. It returns a dict of dicts. The test made me think it returns a dict of lists.

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Done in follow up branch.

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Fixed in follow-up

src/scippneutron/file_loading/nxobject.py
jl-wynen: Ok. But then add a docstring, please.

SimonHeybrock: done in follow-up

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Done in follow-up

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Done in follow-up

src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -148,19 +153,24 @@ def load_dataset(self,
...
+                                dtype=dtype,
+                                unit=self.get_unit(dataset))
+            if variable.values.flags["C_CONTIGUOUS"] and variable.values.size > 0:
+                dataset.read_direct(variable.values)
+            else:
+                variable.values = dataset
+            return variable
+        return sc.array(dims=dimensions,
+                        unit=self.get_unit(dataset),
+                        values=dataset[index].astype(dtype))
```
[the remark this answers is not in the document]
SimonHeybrock: Would be nice, can't think of a solution right now. Maybe there is a helper in numpy? I am adding a TODO note for now, unless you have a better idea?

src/scippneutron/file_loading/_monitor_data.py
```diff
@@ -1,42 +1,76 @@
...
+            for dim in dims
+        }
         return sc.DataArray(data=data, coords=coords)
     else:
         return None
 
 
+def load_monitor(group: Group, nexus: LoadFromNexus, select=tuple()) -> sc.DataArray:
+    # Look for event mode data structures in NXMonitor. Event-mode data takes
+    # precedence over histogram-mode-data if available.
```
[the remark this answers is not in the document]
SimonHeybrock: done in follow-up

tests/test_nexus.py
```diff
@@ -0,0 +1,196 @@
...
+    builder.add_log(Log("log", np.array([1.1, 2.2, 3.3]), np.array([4.4, 5.5, 6.6])))
+    return builder
+
+
+def test_nxobject_tree_traversal(nexus_group: Tuple[Callable, LoadFromNexus]):
+    resource, loader = nexus_group
+    with resource(builder_with_events_monitor_and_log())() as f:
+        root = nexus.NXroot(f, loader)
+        assert root.nx_class == nexus.NX_class.NXroot
+        assert set(root.keys()) == set(['entry', 'monitor'])
```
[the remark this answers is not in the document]
SimonHeybrock: done in follow-up

tests/test_nexus.py
```diff
@@ -0,0 +1,196 @@
...
+def test_nxobject_grandchild_can_be_accessed_using_path(
+        nexus_group: Tuple[Callable, LoadFromNexus]):
+    resource, loader = nexus_group
+    with resource(builder_with_events_monitor_and_log())() as f:
+        root = nexus.NXroot(f, loader)
+        assert root['entry/log'].name == '/entry/log'
+        assert root['/entry/log'].name == '/entry/log'
+
+
+def test_nxobject_by_nx_class_contains_everything(nexus_group: Tuple[Callable,
```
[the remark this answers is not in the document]
SimonHeybrock: done in follow-up

tests/test_nexus.py
```diff
@@ -0,0 +1,196 @@
...
+    builder.add_event_data(event_data)
+    builder.add_monitor(
+        Monitor("monitor",
+                data=np.array([1.]),
+                axes=[("time_of_flight", np.array([1.]))]))
+    builder.add_log(Log("log", np.array([1.1, 2.2, 3.3]), np.array([4.4, 5.5, 6.6])))
+    return builder
+
+
+def test_nxobject_tree_traversal(nexus_group: Tuple[Callable, LoadFromNexus]):
```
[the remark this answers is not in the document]
SimonHeybrock: done in follow-up

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: I changed it directly now, to return `None`.

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: Done in follow-up

src/scippneutron/nexus/nexus.py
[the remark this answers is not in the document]
jl-wynen: >> Then why does it inherit from AbstractContextManager?
>
> Because it is implementing a context manager?

I misread your comment, I though you said, we are not using `nexus.File` as a context manager.
In this case, we should properly enter and exist `h5py.File`, even though `h5py.File.__enter__` is probably a noop, just like for `nexus.File` right now.

My suggestion for `__enter__` above would do that. Also, `__exit__` should exit `self._file` instead of `self._group`. And in order to let users use `nexus.File` outside of context managers, we can also provide
```.py
def close():
    self._file.close()
```
This way, we are really wrapping the `h5py.File`.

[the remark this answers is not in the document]
SimonHeybrock: Done, seems to work.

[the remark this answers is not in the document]
jl-wynen: On the other branch?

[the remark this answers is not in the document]
SimonHeybrock: Ah, yes, sorry.

jl-wynen (approved):
Assuming the next PR will address my comments in this one.

PR #255, opened 2022-01-19
Mantid convert test refactor

### LINKED ISSUE #27: scippneutron.convert, mantid.simpleapi.ConvertUnits differences
See https://github.com/scipp/scippneutron/pull/16#discussion_r587222620

Going to tighter tolerances causes test failure where we cross-compare coordinates. We need to understand the origin of these differences. Easy to reproduce simply by changing the tolerance factor in the line linked above.

jl-wynen:
Fixes #27

This PR removes the common fixture class `MantidScippComparison` because it seemed to be more complicated than necessary. Especially because every test module used a different specialised behaviour of it.

All tolerances are now fairly tight. And individually adjusted.


tests/mantid_scipp_comparison/test_neutron_convert_units.py
```diff
@@ -1,131 +1,144 @@
...
+def test_mantid_convert_tof_to_wavelength():
+    in_ws = make_workspace('tof')
+    out_mantid = mantid_convert_units(in_ws, 'wavelength')
+
+    in_da = scn.mantid.from_mantid(in_ws)
+    out_scipp = scn.convert(data=in_da, origin='tof', target='wavelength', scatter=True)
+
+    assert sc.allclose(out_scipp.coords['wavelength'],
+                       out_mantid.coords['wavelength'],
+                       rtol=1e-8 * sc.units.one)
```
SimonHeybrock, line 78: Wonder if it would be clearer if we also gave `atol` explicitly?

tests/mantid_scipp_comparison/test_neutron_convert_units.py
```diff
@@ -1,131 +1,144 @@
...
+
+    in_da = scn.mantid.from_mantid(in_ws)
+    out_scipp = scn.convert(data=in_da, origin='tof', target='dspacing', scatter=True)
+
+    assert sc.allclose(out_scipp.coords['dspacing'],
+                       out_mantid.coords['dspacing'],
+                       rtol=1e-8 * sc.units.one)
+    assert sc.allclose(out_scipp.coords['spectrum'],
+                       out_mantid.coords['spectrum'],
+                       rtol=1e-8 * sc.units.one)
```
SimonHeybrock, line 96: This are integers, why do we need `allclose`?

tests/mantid_scipp_comparison/test_neutron_convert_units.py
```diff
@@ -1,131 +1,144 @@
...
+    in_ws = make_workspace('tof')
+    out_mantid = mantid_convert_units(in_ws, 'energy')
+
+    in_da = scn.mantid.from_mantid(in_ws)
+    out_scipp = scn.convert(data=in_da, origin='tof', target='energy', scatter=True)
+
+    # Mantid reverses the order of the energy dim.
+    mantid_energy = sc.empty_like(out_mantid.coords['energy'])
+    assert mantid_energy.dims[1] == 'energy'
+    mantid_energy.values = out_mantid.coords['energy'].values[..., ::-1]
```
SimonHeybrock, line 109: Maybe we should implement https://github.com/scipp/scipp/issues/860.

SimonHeybrock (commented):
Looks good, just minor comments.

tests/mantid_scipp_comparison/test_histogram_events.py
```diff
@@ -1,37 +1,45 @@
...
+except ImportError:
+    pytestmark = pytest.mark.skip('Mantid framework is unavailable')
+    sapi = None
+    kernel = None
+
 
+@pytest.fixture
+def in_ws():
+    filename = scn.data.get_path("CNCS_51936_event.nxs")
+    print('Loading', filename)
```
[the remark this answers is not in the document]
jl-wynen: Yes. Copied over from the old implementation. But it seems less relevant now that the loading code is closer to the actual test.

tests/mantid_scipp_comparison/test_neutron_convert_units.py
jl-wynen: Good point. Will change.

tests/mantid_scipp_comparison/test_neutron_convert_units.py
jl-wynen: Yes. But I don't this this PR needs to wait for it.

tests/mantid_scipp_comparison/test_neutron_convert_units.py
jl-wynen: Don't think so. I can set `atol=1e-20` and the test passes. The relative tolerance is so large that the absolute tolerance does not really matter.

SimonHeybrock (approved):

PR #259, opened 2022-01-21
Use metatoenv script in CI

nvaytet:
- remove `developer.yml` files
- add `metatoenv.py` script from [templates](https://github.com/scipp/templates/blob/main/tools/metatoenv.py)
- run script during PR CI to generate conda env


conda/meta.yaml
```diff
@@ -12,8 +12,8 @@ requirements:
     - gxx_linux-64 11.1.* [linux64]
     - git
     - ninja
-    - python {{ python }}
-    - scipp {{ scipp }}
+    - python
+    - scipp 0.11.*
```
SimonHeybrock, line 16: I believe Owen told be to not set versions in `meta.yaml`, but do it in `conda_build_config.yaml`, but I cannot remember why.

tools/metatoenv.py
```diff
@@ -0,0 +1,226 @@
...
+parser = argparse.ArgumentParser(
+    description='Generate a conda environment file from a conda recipe meta.yaml file')
+parser.add_argument('--dir', default='.')
+parser.add_argument('--meta-file', default='meta.yaml')
+parser.add_argument('--env-file', default='environment.yml')
+parser.add_argument('--env-name', '--name', default='')
+parser.add_argument('--channels', default='conda-forge')
+parser.add_argument('--platform', default=None)
+parser.add_argument('--extra', default='')
+parser.add_argument('--merge-with', default='')
```
SimonHeybrock, line 24: Some help messages explaining what the options do would be useful?

SimonHeybrock (commented):
Looks pretty good. I believe the `metatoenv.py` has been reviewed previously so I did not study it in-depth.

conda/meta.yaml
nvaytet: I think this is the reason: https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html
But the only thing we are ever making a matirix for is the python version, which is set via the `--python`` command line argument anyway, because we are parallelising our builds.

I have not tried this matrix building in `conda-build` directly, but I suspect it has to perform the lib solving for each version, and I predict that what we are doing on multiple workers on our CI is faster than building all python versions in a single `conda-build` step.

SimonHeybrock: 👍 

Can you put a comment explaining why we are not using `conda_build_config.yaml` along these lines?

jl-wynen (commented):
Can you add a note in the dev docs about how do generate environment files?

tools/metatoenv.py
nvaytet: Can you look at this PR? Once ok, I will copy the new file over to here.
https://github.com/scipp/templates/pull/4

SimonHeybrock: Merged. Looks good.

docs/developer/getting-started.rst
```diff
@@ -14,6 +14,125 @@ Code contributions to scippneutron should meet the following criteria:
...
+  cmake --build . --target all-benchmarks
+  cmake --build . --target install
+
+
+To use the ``scippneutron`` Python module:
+
+.. code-block:: bash
+
+  cd ../python
+  PYTHONPATH=$PYTHONPATH:../install python3
```
jl-wynen, line 100: Or alternatively, `conda develop ../install` to make it persistent.

docs/developer/getting-started.rst
```diff
@@ -14,6 +14,125 @@ Code contributions to scippneutron should meet the following criteria:
...
+
+In Python:
+
+.. code-block:: python
+
+  import scippneutron as scn
+
+Building using a local build of Scipp
```
jl-wynen, line 108: Make subsection of 'Getting the code, building, and installing' and make build with conda scipp another subsection?

docs/developer/getting-started.rst
```diff
@@ -14,6 +14,125 @@ Code contributions to scippneutron should meet the following criteria:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Instead of using a published Scipp package as part of your ``scippneutron-developer`` conda environment,
+it is also possible to link ``scippneutron`` against a local build of Scipp.
+To avoid conflicts, you will first need to remove the ``scipp`` entry from your generated ``scippneutron-developer.yml`` file.
+Then, use the ``CMAKE_PREFIX_PATH`` to tell ``cmake`` where to find the Scipp C++ libraries:
```
jl-wynen, line 114: Didn't we have `scipp_DIR` at some point? I always used that but can't seem to find it in any CMake files.

SimonHeybrock: That is just standard cmake, not something we have defined ourselves. But yes, refering to that could be useful, since it is specific to this package, whereas prefix-path affects all packages.

docs/developer/getting-started.rst
nvaytet: This is copied from Scipp, so I'll leave it for now. We can change it in both later.

docs/developer/getting-started.rst
nvaytet: ok for making it a subsection. Do we really need the other subsection with conda scipp?

SimonHeybrock (approved):

PR #260, opened 2022-01-24
Extend nexus loading utilities, add NXdata and NXdetector

### LINKED ISSUE #141: load_nexus: support loading individual nexus classes
To avoid the mistakes that accompanied growth of the mantid loaders, we should consider restructuring some parts of the nexus loading utilities. To support the expanded information obtainable from Nexus the loaders will need to do more and we should pause to consider this.

For example,  I think it would be important to introduce testing for the low-level concepts and building blocks such as the methods of `LoadFromHdf5` which are used as utilities onto which the nexus building is structure but could easily disintegrate into quite a mess.

### LINKED ISSUE #226: Loading slices of Nexus files or classes
Based on a discussion regarding use of dask, but also what will be required when working with large files produced from long runs in general (processed with dask or not):

- We will need to be able to load "slices" of data.
- A syntax similar to what `h5py` provides would appear convenient.

Example:

```python
from scippneutron import nexus

f = nexus.File(filename)
bank = f['bank7_events]
bank['pulse_time', 1000:2000]
```
would return pulses 1000 through 2000 as a data array (on this level, probably no binning into pixels would be performed, and no other meta data loaded).

```python
bank['pulse_time', start_time:stop_time]
```
could use label-based indexing (would load `event_time_zero` dataset to use for slicing).

A related but in principle orthogonal issues is #141.

Note also #225, using `pulse_time` above is likely not good. Need a different dimension name? Maybe just `time_zero` or `event_time_zero`?

### LINKED ISSUE #257: Support loading NXdetector with data
`load_nexus` loads parts of `NXdetector`, in particular detector numbers, as needed for loading event data. However, beyond that there appears to be no support for actually loading `NXdetector`, in particular data, if present.

This requirement came up for loading spectra stored by the NUrF sample environment, but it is very likely required more generally.

- Add an `NXdetector` subclass of `NXobject` to support loading this directly.
- Unclear how to handle this in `load_nexus`, since it is not obvious how to deal with multiple/conflicting data entries. Should probably postpone this part.

SimonHeybrock:
There are a number of question marks attached to some of the functionality, pending answers on how to interpret the Nexus standard.

- Closes #257.
- Closes #226, at least for the basics. No support in `load_nexus` added, but this might not be what we want anyway.
- Closes #141, even though we may want to wrap things up further in the future. For now this should be a start and enables users to load individual components without forcing them into specific transformations and assembly of pieces from a file.

### DIFF
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -37,7 +37,6 @@ def _check_for_missing_fields(group: Group, nexus: LoadFromNexus):
     required_fields = (
         "event_time_zero",
         "event_index",
-        "event_id",
         "event_time_offset",
     )
     for field in required_fields:
@@ -219,7 +218,8 @@ class NXevent_data(NXobject):
 
     def _getitem(self, index):
         data = _load_event_group(self._group, self._loader, quiet=False, select=index)
-        data.bins.coords['event_id'] = data.bins.coords.pop('detector_id')
+        if 'detector_id' in data.bins.coords:
+            data.bins.coords['event_id'] = data.bins.coords.pop('detector_id')
         data.bins.coords['event_time_offset'] = data.bins.coords.pop('tof')
         data.coords['event_time_zero'] = data.coords.pop('pulse_time')
         return data
@@ -267,9 +267,12 @@ def _load_event_group(group: Group, nexus: LoadFromNexus, quiet: bool,
     else:
         event_select = slice(None)
 
-    event_id = nexus.load_dataset(group,
-                                  "event_id", [_event_dimension],
-                                  index=event_select)
+    if nexus.dataset_in_group(group, "event_id")[0]:
+        event_id = nexus.load_dataset(group,
+                                      "event_id", [_event_dimension],
+                                      index=event_select)
+    else:
+        event_id = None
 
     event_time_offset = nexus.load_dataset(group,
                                            "event_time_offset", [_event_dimension],
@@ -277,16 +280,14 @@ def _load_event_group(group: Group, nexus: LoadFromNexus, quiet: bool,
 
     # Weights are not stored in NeXus, so use 1s
     weights = sc.ones(dims=[_event_dimension],
-                      shape=event_id.shape,
+                      shape=event_time_offset.shape,
                       unit='counts',
                       dtype=np.float32,
                       with_variances=True)
 
-    events = sc.DataArray(data=weights,
-                          coords={
-                              _time_of_flight: event_time_offset,
-                              _detector_dimension: event_id,
-                          })
+    events = sc.DataArray(data=weights, coords={_time_of_flight: event_time_offset})
+    if event_id is not None:
+        events.coords[_detector_dimension] = event_id
 
     if not last_loaded:
         event_index = np.append(event_index, num_event)
@@ -320,7 +321,7 @@ def _load_event_group(group: Group, nexus: LoadFromNexus, quiet: bool,
                         f" ordered. The index must be ordered to load pulse times.")
 
     if not quiet:
-        print(f"Loaded {len(event_id)} events from "
+        print(f"Loaded {len(event_time_offset)} events from "
               f"{nexus.get_name(group)} containing {num_event} events")
 
     return sc.DataArray(data=binned, coords={"pulse_time": pulse_times})
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -218,22 +218,14 @@ class NXevent_data(NXobject):
         return None
 
     def _getitem(self, index):
-        detector_data = _load_event_group(self._group,
-                                          self._loader,
-                                          DetectorData(),
-                                          quiet=False,
-                                          select=index)
-        data = detector_data.event_data
+        data = _load_event_group(self._group, self._loader, quiet=False, select=index)
         data.bins.coords['event_id'] = data.bins.coords.pop('detector_id')
         data.bins.coords['event_time_offset'] = data.bins.coords.pop('tof')
         data.coords['event_time_zero'] = data.coords.pop('pulse_time')
         return data
 
 
-def _load_event_group(group: Group,
-                      nexus: LoadFromNexus,
-                      detector_data: DetectorData,
-                      quiet: bool,
+def _load_event_group(group: Group, nexus: LoadFromNexus, quiet: bool,
                       select=tuple()) -> DetectorData:
     _check_for_missing_fields(group, nexus)
     index = to_plain_index([_pulse_dimension], select)
@@ -296,10 +288,6 @@ def _load_event_group(group: Group,
                               _detector_dimension: event_id,
                           })
 
-    _check_event_ids_and_det_number_types_valid(
-        event_id.dtype if detector_data.detector_ids is None else
-        detector_data.detector_ids.dtype, event_id.dtype)
-
     if not last_loaded:
         event_index = np.append(event_index, num_event)
     else:
@@ -331,14 +319,11 @@ def _load_event_group(group: Group,
         raise BadSource(f"Event index in NXEvent at {group.name}/event_index was not"
                         f" ordered. The index must be ordered to load pulse times.")
 
-    detector_data.event_data = sc.DataArray(data=binned,
-                                            coords={"pulse_time": pulse_times})
-
     if not quiet:
         print(f"Loaded {len(event_id)} events from "
               f"{nexus.get_name(group)} containing {num_event} events")
 
-    return detector_data
+    return sc.DataArray(data=binned, coords={"pulse_time": pulse_times})
 
 
 def _check_event_ids_and_det_number_types_valid(detector_id_type: Any,
@@ -491,9 +476,17 @@ def _load_data_from_each_nx_event_data(detector_data: Dict,
     for group in event_data_groups:
         parent_path = "/".join(group.name.split("/")[:-1])
         try:
-            new_event_data = _load_event_group(
-                group, nexus, detector_data.get(parent_path, DetectorData()), quiet)
-            event_data.append(new_event_data)
+            new_detector_data = detector_data.get(parent_path, DetectorData())
+            new_event_data = _load_event_group(group, nexus, quiet)
+
+            event_id_dtype = new_event_data.bins.constituents['data'].coords[
+                _detector_dimension].dtype
+            _check_event_ids_and_det_number_types_valid(
+                event_id_dtype if new_detector_data.detector_ids is None else
+                new_detector_data.detector_ids.dtype, event_id_dtype)
+            new_detector_data.event_data = new_event_data
+
+            event_data.append(new_detector_data)
             # Only pop from dictionary if we did not raise an
             # exception when loading events
             detector_data.pop(parent_path, DetectorData())
--- a/src/scippneutron/file_loading/_detector_data.py
+++ b/src/scippneutron/file_loading/_detector_data.py
@@ -224,9 +224,9 @@ class NXevent_data(NXobject):
                                           quiet=False,
                                           select=index)
         data = detector_data.event_data
-        data.bins.coords['id'] = data.bins.coords.pop('detector_id')
-        data.bins.coords['time_offset'] = data.bins.coords.pop('tof')
-        data.coords['time_zero'] = data.coords.pop('pulse_time')
+        data.bins.coords['event_id'] = data.bins.coords.pop('detector_id')
+        data.bins.coords['event_time_offset'] = data.bins.coords.pop('tof')
+        data.coords['event_time_zero'] = data.coords.pop('pulse_time')
         return data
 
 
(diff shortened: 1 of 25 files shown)


src/scippneutron/file_loading/nxobject.py
```diff
@@ -1,20 +1,55 @@
...
 import functools
+from typing import List, Union, Tuple, NoReturn, Any, Dict
 
 from ._nexus import LoadFromNexus
-from ..file_loading._hdf5_nexus import LoadFromHdf5
-from ._common import Group, Dataset
+from ._hdf5_nexus import LoadFromHdf5
+from ._common import Group, Dataset, MissingAttribute
+
+ScippIndex = Tuple[str, Union[int, slice]]
```
jl-wynen, line 14: Missing dict and ellipsis? A type checker should complain if you use `ScippIndex` in `to_plain_index`.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -1,20 +1,55 @@
...
+        try:
+            _ = self[name]
+            return True
+        except MissingAttribute:
+            return False
+
+    def __getitem__(self, name: str) -> Any:
+        attr = self._loader.get_attribute(self._node, name)
+        # Is this check for string attributes sufficient? Is there a better way?
+        if isinstance(attr, str) or isinstance(attr, bytes):
```
jl-wynen, line 47:
```suggestion
        if isinstance(attr, (str, bytes)):
```

I think this is sufficient. There are no other string types to my knowledge. (Unless you count OPAQUE datatypes which we cannot decode but could unwrap and store as `bytes`. See https://docs.h5py.org/en/stable/strings.html#how-to-store-raw-binary-data)

src/scippneutron/file_loading/nxobject.py
```diff
@@ -1,20 +1,55 @@
...
+            return self._loader.get_string_attribute(self._node, name)
```
jl-wynen, line 48: Looks like we are loading the attribute twice. Can we avoid that?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -0,0 +1,87 @@
...
+        if self._is_events and self._detector_number is not None:
+            return self._detector_number.shape
+        # If event data but no detector_number then this gives the underlying
+        # shape of NXevent_data
+        return self._nxbase.shape
+
+    @property
+    def dims(self) -> List[str]:
+        if self._is_events and self._detector_number is not None:
+            return self.attrs.get('axes', ['detector_number'])
```
jl-wynen, line 26: I don't see an 'axes' attr in NXdetector. Is this code correct?
Also, what if `detector_number` is multi dimensional? I would expect to see something like `[dim_0, ..., dim_n]`.

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -0,0 +1,87 @@
...
+        return self._nxbase.unit
+
+    @property
+    def _is_events(self) -> bool:
+        # The standard is unclear on whether the 'data' field may be NXevent_data or
+        # whether the fields of NXevent_data should be stored directly within this
+        # NXdetector. Both cases are observed in the wild.
+        if 'data' in self:
+            return isinstance(self['data'], NXevent_data)
+        return 'event_time_offset' in self
```
jl-wynen, line 42: I don't see 'event_time_offset' this specified in the docs. Is this something that people simply do or did I miss something?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -0,0 +1,87 @@
...
+        # as needed for event data loading
+        if self._is_events:
+            if self.detector_number is None:
+                # If there is no 'detector_data' then self.dims is the same as that of
+                # the underlying event data. Slicing here thus corresponds to slicing
+                # NXevent_data.
+                return self._nxbase[select]
+            # If there is a 'detector_number' field it is used to bin events into
+            # detector pixels. Note that due to the nature of NXevent_data, which stores
+            # events from all pixels and random order we, always have to load the entire
```
jl-wynen, line 79:
```suggestion
            # events from all pixels and random order, we always have to load the entire
```

tests/nexus_helpers.py
```diff
@@ -281,13 +282,16 @@ def add_dataset(parent: Dict, name: str, data: Union[str, bytes,
...
+    def add_attribute(parent: Dict, name: str, value: Union[str, bytes, list,
+                                                            np.ndarray]):
         if isinstance(value, (str, bytes)):
             attr_info = {"string_size": len(value), "type": "string"}
         elif isinstance(value, float):
             attr_info = {"size": 1, "type": "float64"}
         elif isinstance(value, int):
             attr_info = {"size": 1, "type": "int64"}
+        elif isinstance(value, list):
+            attr_info = {"size": len(value), "type": "string"}
```
jl-wynen, line 294: Do lists always contain strings?

tests/test_nexus.py
```diff
@@ -194,3 +226,42 @@ def test_negative_event_index_converted_to_num_event(nexus_group: Tuple[Callable
...
+                                                                      LoadFromNexus]):
+    resource, loader = nexus_group
+    with resource(builder_with_events_and_events_monitor_without_event_id())() as f:
+        event_data = nexus.NXroot(f, loader)['entry/events_0']
+        da = event_data[...]
+        assert len(da.bins.coords) == 1
+        assert 'event_time_offset' in da.bins.coords
+
+
+def test_event_mode_monitor_without_event_id_can_be_loaded(
```
jl-wynen, line 260: I really like the long test names :+1:

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: > I don't see an 'axes' attr in NXdetector. Is this code correct?

Hah, good question. Apparently `NXdetector` (and `NXmonitor`) are conceptually a bit "inheriting" `NXdata` so you will see use of some of the same attributes... but this is based on unclear personal communication on Slack and I have not gotten anyone to update/clarify the standard.

You are right that we should use better defaults for a multi-dim case.

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: See above, apparently this is one of the (standard?!?) ways of storing event data: Either the fields are contained directly, or `data` is an `NXevent_data`. Again, standard is very incomplete, waiting for better confirmation from personal communication.

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -215,44 +214,37 @@ def dims(self):
...
-                                          DetectorData(),
-                                          quiet=False,
-                                          select=index)
-        data = detector_data.event_data
-        data.bins.coords['id'] = data.bins.coords.pop('detector_id')
-        data.bins.coords['time_offset'] = data.bins.coords.pop('tof')
-        data.coords['time_zero'] = data.coords.pop('pulse_time')
+        data = _load_event_group(self._group, self._loader, quiet=True, select=index)
+        # Map back to original field names from NXevent_data since _load_event_group
+        # imposes names that based on assumptions and history.
```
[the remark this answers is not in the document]
jl-wynen: I think so. The loading functions should use the NX* classes when they need to load sub objects. But we still need to put the actual loading code somewhere and the existing functions are as good a place as any.

src/scippneutron/file_loading/nxdata.py
```diff
@@ -0,0 +1,82 @@
...
+            if self[name].attrs.get('signal') == 1:
+                return name
+        return None
+
+    @property
+    def _errors_name(self) -> str:
+        if self._signal_name_default is None:
+            return f'{self._signal_name_default}_errors'
+        else:
+            return 'errors'
```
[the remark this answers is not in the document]
SimonHeybrock: Good catch.

src/scippneutron/file_loading/nxdata.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @property
+    def _signal(self) -> Dataset:
+        return self[self._signal_name]
```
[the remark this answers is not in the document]
SimonHeybrock: Isn't this premature optimization? All of this is just tiny metadata. Also I believe HDF5 does a lot of caching anyway (on top of the OS caching).

Furthermore, for the latter we must be careful, since sooner or later we will have to handle files that are still being written to. I am not sure how to do that (does it just work automagically?), but caching would likely interfere?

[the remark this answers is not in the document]
SimonHeybrock: See also https://docs.h5py.org/en/stable/swmr.html.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Probably, but extracting some implementation details from `LoadFromHdf5`. But maybe we can postpone until it is clear if we are refactoring that?

tests/nexus_helpers.py
SimonHeybrock: Current yes. Since Python lists could also contain mixed types this whole mechanism would break anyway in the most generic case?

src/scippneutron/file_loading/nxdata.py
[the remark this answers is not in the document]
jl-wynen: Ok. Didn't think h5py would cache attributes.

Skimming the docs, it looks like we have to call `refresh` manually to read the latest version in case stuff was cached.

src/scippneutron/file_loading/nxobject.py
jl-wynen: OK

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -0,0 +1,94 @@
...
+    def _getitem(self, select: ScippIndex) -> sc.DataArray:
+        # Note that ._detector_data._load_detector provides a different loading
+        # facility for NXdetector but handles only loading of detector_number,
+        # as needed for event data loading
+        if self._is_events:
+            if self.detector_number is None:
+                # If there is no 'detector_data' then self.dims is the same as that of
+                # the underlying event data. Slicing here thus corresponds to slicing
+                # NXevent_data.
+                return self._nxbase[select]
```
SimonHeybrock, line 83: @jl-wynen I think I don't agree with what I did here any more. I feel that `NXdetector` should imply that the result is "pixel-based" and we should *always* `bin` the `NXevent_data` result. If there is no `detector_number` field we will simply have to fall back to using, e.g., the range of indices that show up in `event_id`.

jl-wynen: What does `event_id` contain in this case? It doesn't looks like the standard specifies it.

Have you seen files where the detector number is missing? It seems like there has to be some information elsewhere in the file that lets us identify pixels.

SimonHeybrock: > What does event_id contain in this case? It doesn't looks like the standard specifies it.

Just some integers.

> Have you seen files where the detector number is missing? It seems like there has to be some information elsewhere in the file that lets us identify pixels.

Yes, SNS files. I think they specify detector numbers in auxiliary files.

jl-wynen: Then I think, I agree that we should bin in `event_id`.

SimonHeybrock: ... and rename to `detector_number`, for consistency?

jl-wynen: Yes

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -0,0 +1,112 @@
...
+            # If there is a 'detector_number' field it is used to bin events into
+            # detector pixels. Note that due to the nature of NXevent_data, which stores
+            # events from all pixels and random order, we always have to load the entire
+            # bank. Slicing with the provided 'select' is done while binning.
+            event_data = self._nxbase[...]
+            if self.detector_number is None:
+                # Ideally we would prefer to use np.unique, but a quick experiment shows
+                # that this can easily be 100x slower, so it is not an option. In
+                # practice most files have contiguous event_id values within a bank
+                # (NXevent_data).
```
jl-wynen, line 95: Using `arange` still produces a correct result, right? It might simply add extra bins compared to using unique event id's?
I think I am fine with it, although it took me a moment to realise.

tests/test_nxdetector.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+    builder = NexusBuilder()
+    builder.add_detector(
+        Detector(detector_numbers=np.array([1, 2, 3, 4]), event_data=event_data))
+    resource, loader = nexus_group
+    with resource(builder)() as f:
+        detector = nexus.NXroot(f, loader)['entry/detector_0']
+        loaded = detector[...]
+        assert sc.identical(loaded.bins.size().data,
+                            sc.array(dims=['detector_number'], values=[2, 3, 1, 0]))
```
jl-wynen, line 72: Where does the order of `values` come from?

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: Correct. There may be gaps, and in theory someone could even define event IDs such that the gaps exceed our memory.

tests/test_nxdetector.py
SimonHeybrock: You have to count the occurrences of particular IDs above:
```
       event_id=np.array([1, 2, 3, 1, 2, 2]),
```

jl-wynen (approved):

PR #262, opened 2022-01-27
Summarize current interpretation of the NeXus format

SimonHeybrock:


src/scippneutron/nexus/docs/our-interpretation-of-the-nexus-format.md
```diff
@@ -0,0 +1,70 @@
...
+More concretely this means that, e.g., for loading an `NXdetector` from a NH, the implementation will:
+
+1. Find all `NXdata` children.
+2. Find all `NXevent_data` children.
+3. Search the group for fields defined in `NXdata`.
+   This includes looking for the `signal` attributes on the group and on fields.
+4. Search the group for fields defined in `NXevent_data`.
+5. Search the group for fields pre-defined in the class that are equivalents of what is defined in `NXdata`, even if the `NXdata` requirements (such as `signal` attributes) are not met.
+
+If the above results in not more than one item, the group can be loaded.
```
nvaytet, line 48:
```suggestion
If the above yields no more than one item, the group can be loaded.
```

src/scippneutron/nexus/docs/our-interpretation-of-the-nexus-format.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `axes` attribute uses `'.'` to define an axis without coordinate field.
+The implication of the above is that there is no way to define the *name* of an axis.
+For example, an `NXdata` group storing a stack of images may not define a coordinate for the "image" dimension of the signal dataset.
+The NF does not specify any other way to define such an axis name.
+We therefore have to fall back to a generic and meaningless dimension label.
+
+Note that the `axes` attribute of the signal field could in principle be used to define such names and the `axes` attribute of the group would then only define the field names of coordinates.
+That is, both attributes would work in conjunction.
+It is unclear whether this is allowed or actually considered invalid in the NF.
```
nvaytet, line 70: Since you have something about missing axes, should we also add something about missing units?

nvaytet (commented):
Nicely written :+1:

src/scippneutron/nexus/docs/our-interpretation-of-the-nexus-format.md
SimonHeybrock: Actually I don't know if we have reached a conclusion on this one?

zjttoefs:
Few comments:
* In practice for ESS you will see NXdata and bin edges mostly for ODIN. 
* I would like to discuss the relationship of `NXdetector` and the corresponding `NXevent_data`. I find that may be too indirect. And we can cross check some of the other content.
* For the axes I am not sure why the field name should not be a good first guess for the label. 
* The axes units should be covered in the units attribute.

SimonHeybrock:
@zjttoefs 
>     * I would like to discuss the relationship of `NXdetector` and the corresponding `NXevent_data`. I find that may be too indirect. And we can cross check some of the other content.

I suppose this is best done in a call?

>     * For the axes I am not sure why the field name should not be a good first guess for the label.

Which field? The problem is the axes that are defined using `"."` in the `axes` attribute, i.e., there is no field?

>     * The axes units should be covered in the units attribute.

If you are referring to the question @nvaytet posed, that was about potentially *missing* units attributes (in general, not specific to axes), and whether we should treat this as "dimensionless", "no unit", "default unit" (scipp defines this depending on the dtype &mdash; "dimensionless" for numbers, "no unit" in other cases), or raise an error.

SimonHeybrock:
@nvaytet Can you have a brief look at the clarifications I just pushed? Cheers!

nvaytet (approved):

PR #265, opened 2022-02-04
Refactor loading of NXlog to use NXdata

SimonHeybrock:
- Multi-dim logs can now be loaded
- Length-0 logs can now be loaded
- Consistent handling of unit strings that cannot be parsed. I am not sure this is the correct choice. For now this is returning `dimensionless`, may want to switch to `None` after next scipp release.
- NXlog improved (dims, shape, .. working).
- NXtransformations now used NXlog, reducing code duplication.


SimonHeybrock:
@Tom-Willemsen Could you have a look specifically at the changes to `_transformations.py`. Cheers!

src/scippneutron/file_loading/_log_data.py
```diff
@@ -53,95 +53,50 @@ def _add_log_to_data(log_data_name: str, log_data: sc.Variable, group_path: str,
 class NXlog(NXobject):
```
jl-wynen, line 53: Rename the file to `nxlog.py`?

src/scippneutron/file_loading/nxdata.py
```diff
@@ -22,8 +24,8 @@ def shape(self) -> List[int]:
     def dims(self) -> List[str]:
         # Apparently it is not possible to define dim labels unless there are
         # corresponding coords. Special case of '.' entries means "no coord".
-        if 'axes' in self.attrs:
-            axes = self.attrs['axes']
+        if self.attrs.get('axes', self._axes_default) is not None:
+            axes = self.attrs.get('axes', self._axes_default)
```
jl-wynen, line 28: Looks like you want to use the walrus operator here....

jl-wynen, line 28: The `axes` attr overrides `_axes_default`. This makes sense in general but is this what we want in `NXlog`? Your comment there makes it look like we always want the first axis to be 'time'.

src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -234,7 +234,14 @@ def get_unit(node: Union[h5py.Dataset, h5py.Group]) -> str:
...
             return "dimensionless"
-        return _ensure_str(units, LoadFromHdf5.get_attr_encoding(node, "units"))
+        units = _ensure_str(units, LoadFromHdf5.get_attr_encoding(node, "units"))
+        try:
+            sc.Unit(units)
+        except sc.UnitError:
+            warnings.warn(f"Unrecognized unit '{units}' for value dataset "
+                          f"in '{node.name}'; setting unit as 'dimensionless'")
+            return "dimensionless"
+        return units
```
jl-wynen, line 244: Why don't you return `sc.Unit(units)`?

src/scippneutron/file_loading/_transformations.py
```diff
@@ -249,42 +236,24 @@ def _get_transformation_magnitude_and_unit(group_name: str,
...
-        if unit == sc.units.dimensionless:
-            # See if the value unit is on the NXLog itself instead
-            unit = nexus.get_unit(transform)
-            if unit == sc.units.dimensionless:
-                raise TransformationError(f"Missing units for transformation at "
-                                          f"{nexus.get_name(transform)}")
     else:
         magnitude = nexus.load_dataset_as_numpy_array(transform).astype(float).item()
         unit = nexus.get_unit(transform)
+        da = sc.scalar(value=magnitude, unit=unit, dtype=sc.DType.float64)
```
Tom-Willemsen, line 251: Maybe we should name this variable something different? `da` is a little confusing if it might not be a dataarray.

src/scippneutron/file_loading/_transformations.py
```diff
@@ -249,42 +236,24 @@ def _get_transformation_magnitude_and_unit(group_name: str,
...
-                raise TransformationError(f"Found empty NXlog as a "
-                                          f"transformation for {group_name}")
-            if len(values) != len(times):
-                raise TransformationError(f"Mismatched time and value dataset lengths "
-                                          f"for transformation at {group_name}")
-
+            da = NXlog(transform, nexus)[()]
+        except sc.DimensionError:
+            raise TransformationError(f"Mismatched time and value dataset lengths "
+                                      f"for transformation at {group_name}")
```
jl-wynen, line 242: There are a lot of very specific error messages raised at places where you cannot be completely sure that this is actually the error that was encountered. (Was already the case before this PR.) I have seen cases where the messages got out of sync and reported wrong errors. Is it possible to move them closer to the code that actually triggers the error?

tests/test_nxdetector.py
```diff
@@ -1,4 +1,4 @@
-from .test_nexus import open_nexus, open_json
+from .nexus_test import open_nexus, open_json
```
Tom-Willemsen, line 1: I don't really have a strong opinion on whether `test_nexus` or `nexus_test` is better, but I do think it would worth being consistent in our test naming scheme? At the moment most test files follow the `test_*` style.

src/scippneutron/file_loading/_transformations.py
```diff
@@ -249,42 +236,24 @@ def _get_transformation_magnitude_and_unit(group_name: str,
...
-                                      dim="time",
-                                      group_name=group_name)
-            if len(values) == 0:
-                raise TransformationError(f"Found empty NXlog as a "
-                                          f"transformation for {group_name}")
-            if len(values) != len(times):
-                raise TransformationError(f"Mismatched time and value dataset lengths "
-                                          f"for transformation at {group_name}")
-
+            da = NXlog(transform, nexus)[()]
```
Tom-Willemsen, line 239: I find this notation a bit unintuitive (and hard to search for in docs). Could we have a utility method on `NXLog` that returns a dataarray rather than relying on this somewhat obscure slicing pattern?

Tom-Willemsen (commented):
I like the approach in general - this seems to have got rid of a lot of duplicate/nearly-duplicate code.

tests/test_nxdetector.py
jl-wynen: We renamed all files in scipp to have `_test` as a postfix because that makes it easier to use tab-completion in a terminal. We should do the same in scippneutron. We just didn't get around to doing it.

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: This is what `h5py` uses, and the `NX*` classes are trying to mimic that. When we add `NXtransformations` this will probably be modified to support a slice index, so it will look less obscure.

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: Not sure, you'd get the wrong exception type then?

src/scippneutron/file_loading/_hdf5_nexus.py
SimonHeybrock: Probably just historic reasons, this method may be expected to return a string?

src/scippneutron/file_loading/nxdata.py
SimonHeybrock: I think if someone puts the `axes` attribute on the group we should follow that? The standard does not specify this.

SimonHeybrock: Can we drop 3.7 support? 😬

SimonHeybrock: I improved the comment and documented the parameter in NXdata.

SimonHeybrock:
I think I addressed all comments. Since there is follow-up work I'd like to start, I'll merge this. Happy to address things elsewhere if there are more comments.

PR #267, opened 2022-02-07
Load multi valued transforms

Tom-Willemsen:
Fixes https://github.com/scipp/scippneutron/issues/34


src/scippneutron/file_loading/_detector_data.py
```diff
@@ -396,7 +403,10 @@ def _bin_events(data: DetectorData):
         if pixel_positions_loaded:
             # TODO: the name 'position' should probably not be hard-coded but moved
             # to a variable that cah be changed in a single place.
-            da.coords['position'] = data.pixel_positions
+            da.coords['base_position'] = data.pixel_positions
+            if data.pixel_position_transforms is not None:
+                da.attrs['position_transformations'] = sc.scalar(
+                    value=data.pixel_position_transforms)
```
SimonHeybrock, line 409: I presume this is wrapped in a scalar since it may have an extra time dimension?

tests/test_load_nexus.py
```diff
@@ -632,10 +632,11 @@ def test_loads_pixel_positions_with_event_data(load_function: Callable):
         np.concatenate((y_pixel_offset_1, y_pixel_offset_2.flatten())),
         np.concatenate((z_pixel_offset_1, z_pixel_offset_2.flatten()))
     ]).T / 1_000  # Divide by 1000 for mm to metres
-    assert np.allclose(loaded_data.coords['position'].values, expected_pixel_positions)
+    assert np.allclose(loaded_data.coords['base_position'].values,
+                       expected_pixel_positions)
```
SimonHeybrock, line 636: If there is no time-dependent transformation, shouldn't we expect/use `*_position`? If everything is `*_base_position`, downstream code cannot handle it. `base_position` may be there in addition, but generally I'd assume we want to work with the actual positions?

src/scippneutron/file_loading/_positions.py
```diff
@@ -44,46 +42,70 @@ def load_positions_of_components(groups: List[Group],
...
+    _add_coord_to_loaded_data(attr_name=f"{name}_base_position",
+                              data=data,
+                              value=position,
+                              unit=units,
+                              dtype=sc.DType.vector3)
+
+    if transformation is None:
+        _add_coord_to_loaded_data(attr_name=f"{name}_position",
+                                  data=data,
+                                  value=position,
```
SimonHeybrock, line 69: Is this making a copy? Would we want `position` and `base_position` to refer to the same variable instead?

src/scippneutron/file_loading/_positions.py
```diff
@@ -44,46 +42,70 @@ def load_positions_of_components(groups: List[Group],
...
+                                  unit=units,
+                                  dtype=sc.DType.vector3)
+    else:
+        if isinstance(transformation, sc.Variable):
+            _add_coord_to_loaded_data(
+                attr_name=f"{name}_position",
+                data=data,
+                value=(transformation * sc.vector(value=position, unit=units)).values,
```
SimonHeybrock, line 77: This is checking if it is not time-dependent?

src/scippneutron/file_loading/_positions.py
```diff
@@ -44,46 +42,70 @@ def load_positions_of_components(groups: List[Group],
...
+                unit=units,
+                dtype=sc.DType.vector3)
+
+        _add_coord_to_loaded_data(attr_name=f"{name}_transform",
+                                  data=data,
+                                  value=transformation,
+                                  unit=None)
```
SimonHeybrock, line 84: So now `transformation` can be a `DataArray`?

src/scippneutron/file_loading/_positions.py
```diff
@@ -44,46 +42,70 @@ def load_positions_of_components(groups: List[Group],
...
+
+        _add_position_to_data(name, data, transformation, position, units)
+
+
+def _add_position_to_data(name: str, data: sc.Variable,
+                          transformation: Optional[Union[sc.Variable, sc.DataArray]],
+                          position: np.ndarray, units: sc.Unit):
+
+    _add_coord_to_loaded_data(attr_name=f"{name}_base_position",
+                              data=data,
```
SimonHeybrock, line 61: If we store `position` below, would we rather store `base_position` as an attribute?

src/scippneutron/file_loading/_detector_data.py
SimonHeybrock: After having looked at the code in `_positions.py` below, this seems inconsistent? Not storing `position` if there is event data? Can we use the same helper function as in `_positions.py` to add the coords?

src/scippneutron/file_loading/_positions.py
Tom-Willemsen: Yes, it would only be a `Variable` if it's a single-valued (scalar) transform. If it is multi-valued (i.e. time dependent) it would be a `DataArray`. I'll clarify this in a comment.

src/scippneutron/file_loading/_positions.py
Tom-Willemsen: Yes, although it will get wrapped in a scalar `Variable` when used as a coord (because otherwise the `dims` interfere with the DataArray's dims). If there's a better way to do this let me know?

src/scippneutron/file_loading/_detector_data.py
Tom-Willemsen: > I presume this is wrapped in a scalar since it may have an extra time dimension?

Yes, that's correct. If there's a better way to do this let me know?


> After having looked at the code in `_positions.py` below, this seems inconsistent? Not storing `position` if there is event data? Can we use the same helper function as in `_positions.py` to add the coords?

Yes, this is an oversight, will fix this as you suggest.

src/scippneutron/file_loading/_positions.py
SimonHeybrock: I think it is just confusing that `_add_coord_to_loaded_data` wraps things in a scalar, when `unit=None`? Maybe we can avoid this non-obvious mechanism?

Tom-Willemsen: OK - I will see if I can find a way to refactor this to make it more obvious.

src/scippneutron/file_loading/_positions.py
Tom-Willemsen: Would this mean that as a user I sometimes need to do `data.coords['base_position']` and other times `data.attrs['base_position']` depending on whether the data contains transforms? If so I think I prefer keeping it consistent and always storing in `coords`.

Out of interest what would be the advantage of storing it in `attrs` as opposed to `coords`? Do `coords` have some additional computational cost compared to `attrs`?

SimonHeybrock: I agree with your concerns, but note that:
- The can use `da.meta['base_position']` if they do not care whether it is a coord or an attr.
- If there is `da.coords['position']` it is very unlikely that they care about `base_position`. As far as I understand `base_position` may typically be the detector panel placed at the origin, i.e., not very useful?
- Coordinates are compared in all binary operations, i.e., there is a small extra cost.
- If for some reason two data arrays have the same `position` but different `base_position` they cannot be added (subtracted, multiplied, ...), unless `base_position` is made an attribute.

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -404,9 +404,21 @@ def _bin_events(data: DetectorData):
...
-                da.attrs['position_transformations'] = sc.scalar(
+
+            if data.pixel_position_transforms is None:
+                da.coords['position'] = data.pixel_positions
+            else:
+                da.coords['position_transformations'] = sc.scalar(
                     value=data.pixel_position_transforms)
+
+                print(data.pixel_position_transforms)
+                print(data.pixel_positions)
```
SimonHeybrock, line 415: Leftover prints.

src/scippneutron/file_loading/_positions.py
```diff
@@ -42,55 +42,48 @@ def load_positions_of_components(groups: List[Group],
...
     else:
         if isinstance(transformation, sc.Variable):
-            _add_coord_to_loaded_data(
-                attr_name=f"{name}_position",
-                data=data,
-                value=(transformation * sc.vector(value=position, unit=units)).values,
-                unit=units,
-                dtype=sc.DType.vector3)
+            # The transform will be a (scalar) variable if it is not time-dependent.
+            # This means that we can calculate the actual position easily.
```
SimonHeybrock, line 70: Would it make sense to do this also if `transformation.sizes['time'] == 1`?

src/scippneutron/file_loading/_positions.py
```diff
@@ -120,25 +113,15 @@ def _get_base_pos_and_transforms_of_component(
...
-        if dtype is not None:
-            if dtype == sc.DType.vector3:
-                data[attr_name] = sc.vector(value=value, unit=unit)
-            else:
-                data[attr_name] = sc.scalar(value, dtype=dtype, unit=unit)
-        else:
-            data[attr_name] = sc.scalar(value, unit=unit)
+        data[attr_name] = coord
     except KeyError:
         pass
```
SimonHeybrock, line 127: When would such a `KeyError` occur?

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -404,9 +404,21 @@ def _bin_events(data: DetectorData):
...
+
+                if data.pixel_position_transforms.sizes == data.pixel_positions.sizes:
```
SimonHeybrock, line 417: Are you sure this is the correct condition? I suppose the transformation would typically be the same for all pixels, so they sizes rarely match since the application of the transform below would involve a broadcast.

src/scippneutron/file_loading/_positions.py
SimonHeybrock: Also, considering that 90%+ of files will not have time-dependent transformations, it would be odd to have the "same" coord twice, under a different name?

src/scippneutron/file_loading/_common.py
```diff
@@ -193,3 +193,30 @@ def check_1d():
...
+        attrs = data.attrs
+    else:
+        coords = data
+        attrs = data
+
+    if transforms is None:
+        coords[position_name] = positions
+        attrs[base_position_name] = positions
+    else:
+        coords[transform_name] = sc.scalar(value=transforms)
```
SimonHeybrock, line 215: Following the logic of what `transform_coords` would do: Computing `position = transforms * base_position` would turn both `transforms` and `base_position` into attributes, so I think we should do the same, if we store `position`?

SimonHeybrock (approved):

PR #269, opened 2022-02-07
Support loading NXdetector with multi-dim detector_number field

SimonHeybrock:


tests/nxdetector_test.py
```diff
@@ -93,6 +94,29 @@ def test_loads_event_data_mapped_to_detector_numbers_based_on_their_event_id(
             sc.array(dims=['detector_number'], dtype='int64', values=[2, 3, 1, 0]))
 
 
+def test_loads_event_data_with_2d_detector_numbers(nexus_group: Tuple[Callable,
```
nvaytet, line 97: I am guessing it would work but have you tried with a 3d detector?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -109,12 +109,12 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
                                              dtype=id_min.dtype)
             else:
                 detector_numbers = self.detector_number[select]
```
nvaytet, line 111: Can `detector_numbers` just be a single number in some cases? And if so, does it still behave well?

nvaytet (commented):
looks good, just some small questions

tests/nxdetector_test.py
SimonHeybrock: No I have not, but since nothing is hard-coded about 2d...

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: I suppose the question is more general: Can our code handle scalar (0d) HDF5 datasets?

SimonHeybrock:
I'll merge this since I have many follow-up changes in another branch.

PR #270, opened 2022-02-08
Load pixel offset, string dataset, and more

SimonHeybrock:
- Load `x_pixel_offset` and friends as `pixel_offset` in `NXdetector`.
- Fix a number of inconsistencies in `NXdetector`.
- Support loading datasets containing strings. I am very doubtful that it works in all cases since there appear to be multiple options for strings in HDF5.
- Refactor so `read_direct` is used in more cases (should be able to yield better performance).
- Avoid expensive `np.unique` when detector numbers are missing in `load_nexus`.
- Avoid expensive loop in `from_mantid` for empty `EventWorkspace`. This can significantly speed up loading metadata via Mantid, e.g., via `scn.load(filename, mantid_args={'MetaDataOnly':True})`.

### DIFF
--- a/src/scippneutron/file_loading/_hdf5_nexus.py
+++ b/src/scippneutron/file_loading/_hdf5_nexus.py
@@ -153,19 +153,22 @@ class LoadFromHdf5:
 
         if dtype is None:
             dtype = _ensure_supported_int_type(dataset.dtype.type)
-        if index == tuple():
-            variable = sc.empty(dims=dimensions,
-                                shape=dataset.shape,
-                                dtype=dtype,
-                                unit=self.get_unit(dataset))
-            if variable.values.flags["C_CONTIGUOUS"] and variable.values.size > 0:
-                dataset.read_direct(variable.values)
-            else:
-                variable.values = dataset
-            return variable
-        return sc.array(dims=dimensions,
-                        unit=self.get_unit(dataset),
-                        values=dataset[index].astype(dtype))
+
+        shape = list(dataset.shape)
+        if isinstance(index, slice):
+            index = (index, )
+        for i, ind in enumerate(index):
+            shape[i] = len(range(*ind.indices(shape[i])))
+
+        variable = sc.empty(dims=dimensions,
+                            shape=shape,
+                            dtype=dtype,
+                            unit=self.get_unit(dataset))
+        if variable.values.flags["C_CONTIGUOUS"] and variable.values.size > 0:
+            dataset.read_direct(variable.values, source_sel=index)
+        else:
+            variable.values = dataset[index]
+        return variable
 
     def load_dataset_from_group_as_numpy_array(self,
                                                group: h5py.Group,
--- a/src/scippneutron/file_loading/_hdf5_nexus.py
+++ b/src/scippneutron/file_loading/_hdf5_nexus.py
@@ -160,6 +160,9 @@ class LoadFromHdf5:
                             dimensions: Optional[List[str]] = [],
                             dtype: Optional[Any] = None,
                             index=tuple()) -> sc.Variable:
+        """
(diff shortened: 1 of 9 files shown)


src/scippneutron/file_loading/_hdf5_nexus.py
```diff
@@ -150,22 +150,40 @@ def load_dataset(self,
...
+        return self.load_dataset_direct(dataset,
+                                        dimensions=dimensions,
+                                        dtype=dtype,
+                                        index=index)
+
+    def load_dataset_direct(self,
+                            dataset: h5py.Dataset,
+                            dimensions: Optional[List[str]] = [],
+                            dtype: Optional[Any] = None,
+                            index=tuple()) -> sc.Variable:
```
nvaytet, line 162: This would also benefit from a docstring, at least to say what it's doing and in what cases it shoudl be used.

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -79,42 +80,68 @@ def _nxbase(self) -> NXdata:
...
+                offset.fields.y = self[comp][index].to(unit=x.unit, copy=False)
+        return offset.rename_dims(dict(zip(offset.dims, self.dims)))
 
     def _getitem(self, select: ScippIndex) -> sc.DataArray:
         # Note that ._detector_data._load_detector provides a different loading
         # facility for NXdetector but handles only loading of detector_number,
         # as needed for event data loading
+        coords = {}
+        coords['detector_number'] = self.detector_number(select)
+        coords['pixel_offset'] = self.pixel_offset(select)
```
nvaytet, line 111:
```suggestion
        coords = {'detector_number': self.detector_number(select),
                  'pixel_offset': self.pixel_offset(select)}
```

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -79,42 +80,68 @@ def _nxbase(self) -> NXdata:
...
-            if self.detector_number is None:
+            if coords['detector_number'] is None:
+                if select not in (Ellipsis, tuple()):
+                    raise NexusStructureError(
+                        "Cannot load slice of NXdetector since it contains event data "
+                        "but no 'detector_number' field, i.e., the shape is unknown. "
+                        "Use ellipsis or an empty tuple to load the full detector.")
                 # Ideally we would prefer to use np.unique, but a quick experiment shows
                 # that this can easily be 100x slower, so it is not an option. In
                 # practice most files have contiguous event_id values within a bank
```
nvaytet, line 126: Should we be checking whether the `event_id` values are indeed contiguous?

src/scippneutron/mantid.py
```diff
@@ -622,20 +622,24 @@ def convert_EventWorkspace_to_data_array(ws,
...
-        if load_pulse_times:
-            pulse_times['event',
-                        current:current + size].values = sp.getPulseTimesAsNumpy()
-        if _contains_weighted_events(sp):
-            weights['event', current:current + size].values = sp.getWeights()
-            weights['event', current:current + size].variances = sp.getWeightErrors()
-        begins.values[i] = current
-        ends.values[i] = current + size
-        current += size
+    if n_event > 0:  # Skip expensive loop if there are no events
```
nvaytet, line 625: :+1:

nvaytet (commented):
I don't have much to contribute here, although I have to say I am not so familiar with this part of the codebase...

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: They are not. There may be pixels without counts, and there will be no such `event_id` in the file. If you want to do this you are back to using `unique`, I think?

nvaytet (approved):

PR #275, opened 2022-02-10
NXtransformations bugs

### LINKED ISSUE #272: NXtransformations should be interpreted as active transformations
Currently `load_nexus` interprets transformations as *passive*. The origin of this is https://github.com/ess-dmsc/nexus-constructor/issues/829 (same author). But this has since been reverted, see https://github.com/ess-dmsc/nexus-constructor/pull/870 and https://jira.esss.lu.se/browse/ECDC-2718. This seems to align with what can be read between the lines in https://manual.nexusformat.org/classes/base_classes/NXtransformations.html, which never explicitly states that transformations are active, but the way transformations are applied to a sample vector are consistent with that.

### LINKED ISSUE #273: Transform applied to offset in NXtransformations loading?
Consider:
https://github.com/scipp/scippneutron/blob/2e39bea24387729c84c3a370e9d85771ad9879b1/src/scippneutron/file_loading/_transformations.py#L217

https://manual.nexusformat.org/classes/base_classes/NXtransformations.html says:

>  `o` is given by the offset attribute, `t` is given by the `vector` attribute multiplied by the field value

For me this should therefore give a translation defined as `value * vector + offset`, i.e., apart from the sign, the parenthesis are wrong.

Related: #272, which affects the sign in the equation.

### LINKED ISSUE #274: NXtransformations offset ignored when loading rotation
See https://github.com/scipp/scippneutron/blob/2e39bea24387729c84c3a370e9d85771ad9879b1/src/scippneutron/file_loading/_transformations.py#L259-L276.

SimonHeybrock:
- Fixes #272
- Fixes #273 
- Fixes #274

Requesting two reviews, since spatial transformations are so easy to get wrong...

### DIFF
--- a/src/scippneutron/file_loading/_transformations.py
+++ b/src/scippneutron/file_loading/_transformations.py
@@ -33,8 +33,7 @@ def _rotation_matrix_from_axis_and_angle(axis: np.ndarray,
         A dataset of rotation matrices.
     """
     rotvec = sc.vector(value=axis)
-    # We multiply by -1*angle to get a "passive transform"
-    rotvecs = rotvec * sc.scalar(-1.0) * angles.astype(sc.DType.float64, copy=False)
+    rotvecs = rotvec * angles.astype(sc.DType.float64, copy=False)
     matrices = sc.spatial.rotations_from_rotvecs(dims=angles.dims,
                                                  values=rotvecs.values,
                                                  unit=sc.units.rad)
@@ -75,7 +74,7 @@ def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> sc.Dat
 
     :param group: The HDF5 group of the component, containing depends_on
     :param nexus: wrap data access to hdf file or objects from json
-    :return: 4x4 passive transformation matrix as a data array
+    :return: 4x4 active transformation matrix as a data array
     """
     transformations = []
     try:
@@ -179,8 +178,11 @@ def _append_transformation(transform: Union[h5py.Dataset, GroupObject],
         try:
             offset = nexus.get_attribute_as_numpy_array(transform,
                                                         "offset").astype(float)
+            offset_unit = nexus.get_string_attribute(transform, "offset_units")
+            offset = sc.vector(value=offset, unit=offset_unit)
         except MissingAttribute:
-            offset = np.array([0., 0., 0.], dtype=float)
+            offset = sc.vector(value=np.array([0., 0., 0.], dtype=float),
+                               unit=sc.units.m)
 
         transform_type = nexus.get_string_attribute(transform, "transformation_type")
         if transform_type == 'translation':
@@ -209,7 +211,7 @@ def _normalize(vector: np.ndarray, transform_name: str) -> np.ndarray:
     return vector / norm
 
 
-def _append_translation(offset: np.ndarray, transform: GroupObject,
+def _append_translation(offset: sc.Variable, transform: GroupObject,
                         transformations: List[np.ndarray],
                         direction_unit_vector: np.ndarray, group_name: str,
                         nexus: LoadFromNexus):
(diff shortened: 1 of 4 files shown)


src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,8 +171,11 @@ def _append_transformation(transform: Union[h5py.Dataset, GroupObject],
         try:
             offset = nexus.get_attribute_as_numpy_array(transform,
                                                         "offset").astype(float)
+            offset_unit = nexus.get_string_attribute(transform, "offset_units")
+            offset = sc.vector(value=offset, unit=offset_unit)
         except MissingAttribute:
```
Tom-Willemsen, line 176: If `offset_units` isn't present will this fall back to the `except MissingAttribute:` clause and an assume a zero offset vector? Both `offset` and `offset_units` are optional in the NX standard but as far as I can tell it doesn't state you have to have units if you have values? If we have values but no units maybe we should warn/throw and not load this transform?

Tom-Willemsen (commented):
I think the comments on lines 1219 and 1285 of `test_load_nexus.py` relating to passive transforms need updating to reflect that we're now using active transforms (can't comment directly on those lines in github...)

Also, apologies as at least one of these bugs was probably introduced by me during the recent refactoring. I think I misread `o` as `0` in the rotations section of NXtransformations...

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: Good catch! Raising `TransformationError` now.

SimonHeybrock:
> Also, apologies as at least one of these bugs was probably introduced by me during the recent refactoring. I think I misread `o` as `0` in the rotations section of NXtransformations...

No worries, I reviewed it so we have to share the blame :D I remove the corresponding lines from the release notes, since apparently it was not released yet.

src/scippneutron/file_loading/_transformations.py
```diff
@@ -172,8 +171,16 @@ def _append_transformation(transform: Union[h5py.Dataset, GroupObject],
...
             offset = nexus.get_attribute_as_numpy_array(transform,
                                                         "offset").astype(float)
+            try:
+                offset_unit = nexus.get_string_attribute(transform, "offset_units")
+            except MissingAttribute:
+                raise TransformationError(
+                    f"Found offset={offset} but no corresponding 'offset_units' "
+                    f"attribute at {nexus.get_name(transform)}")
+            offset = sc.vector(value=offset, unit=offset_unit)
         except MissingAttribute:
```
jl-wynen, line 181: Nested `except` with the same exception type seem finicky. Can you move this one to line 174?

tests/test_load_nexus.py
```diff
@@ -958,19 +958,64 @@ def test_loads_component_position_from_single_transformation(
...
+@pytest.mark.parametrize("component_class,component_name", [(Sample, "sample"),
+                                                            (Source, "source")])
+@pytest.mark.parametrize("transform_type,value,value_units,expected_position",
+                         ((TransformationType.ROTATION, 180, "deg", [-1, -2, 3]),
+                          (TransformationType.TRANSLATION, 230, "cm", [1, 2, 5.3])))
+def test_raises_if_offset_but_not_offset_units_found(
+        component_class: Union[Type[Source], Type[Sample]], component_name: str,
+        transform_type: TransformationType, value: float, value_units: str,
+        expected_position: List[float], load_function: Callable):
+    builder = NexusBuilder()
```
jl-wynen, line 1007: Are we at a point where we can set up integration tests and substitute this for EC/DCs NeXus builder? This would have caught the active / passive mixup.

We should still keep our builder to facilitate cheap unit tests, though.

SimonHeybrock: I don't know enough about their builder to answer that.

jl-wynen: I was more concerned with our side of it. If we are able to load sufficiently complete files or if there is something missing.

SimonHeybrock: `load_nexus` is designed in a way that it just skips things it does not understand, so we should be able to use real files.

src/scippneutron/file_loading/_transformations.py
SimonHeybrock: Well, then we need to introduce an extra flag to tells us whether `offset` was found? Is that less finicky?

jl-wynen (approved):

PR #276, opened 2022-02-10
Fix for loading relative transforms

Tom-Willemsen:
Fixes loading chained transforms from certain files.

I think this case was previously broken but hidden by a misleading warning, and then became completely broken during one of the recent merges.

Relevant part of NX standard is:

> If a relative path is given, it is relative to the group enclosing the depends_on specification.

from https://manual.nexusformat.org/classes/base_classes/NXtransformations.html

### DIFF
--- a/src/scippneutron/file_loading/_transformations.py
+++ b/src/scippneutron/file_loading/_transformations.py
@@ -124,12 +124,18 @@ def _get_transformations(transform_path: str, transformations: List[np.ndarray],
     if transform_path != '.':
         try:
             transform = nexus.get_object_by_path(group.file, transform_path)
-        except KeyError:
+        except MissingDataset:
             raise TransformationError(
                 f"Non-existent depends_on path '{transform_path}' found "
                 f"in transformations chain for {group_name}")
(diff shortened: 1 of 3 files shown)


src/scippneutron/file_loading/_transformations.py
```diff
@@ -124,12 +124,18 @@ def _get_transformations(transform_path: str, transformations: List[np.ndarray],
...
             raise TransformationError(
                 f"Non-existent depends_on path '{transform_path}' found "
                 f"in transformations chain for {group_name}")
         next_depends_on = _append_transformation(transform, transformations, group_name,
                                                  nexus)
+
+        if not next_depends_on == "." and not next_depends_on.startswith("/"):
+            # Path is relative - convert it to an absolute path relative to the parent
+            # of the transform it was loaded from.
+            next_depends_on = f"/{transform.parent.name}/{next_depends_on}"
```
SimonHeybrock, line 137: Can you add a unit test?

SimonHeybrock (approved):

PR #277, opened 2022-02-14
Add API for selecting events when loading NXdetector

SimonHeybrock:
`NXdetector` may wrap an `NXevent_data`. In that case the detector defines a mapping of events into pixels. The dimensions of the event data are `['pulse']`, i.e., data is binned by pulse. `NXdetector` maps this into pixel-based grouping. In principle we would then have both, i.e., `dims=['detector_number', 'pulse']` but in practice this is completely infeasible due to the massive bin count (each axis could have a length of 1 million). Therefore `NXdetector` *drops* the "pulse" binning and has `dims=['detector_number']`.

This unavoidable limitation prevents selecting a subset of events when loading an `NXdetector` (otherwise we could have used `det['pulse', :1000]`, but 'pulse' is not a dimension of `det`). The alternative would be to work directly with `NXevent_data`, performing binning into pixels by hand, but in practice this is too cumbersome and error prone.

Therefore this PR introduced a mechanism for selecting a range of events (pulses) for loading with `NXdetector`:

```python
event_data = det.select_events['pulse', :1000][:]
```
would load events of the first 1000 pulses, mapped into pixels.

### DIFF
--- a/src/scippneutron/file_loading/nxdetector.py
+++ b/src/scippneutron/file_loading/nxdetector.py
@@ -1,9 +1,27 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+from __future__ import annotations
+from copy import copy
 from typing import List, Union
 import scipp as sc
 from .nxobject import NX_class, NXobject, Field, ScippIndex, NexusStructureError
 from .nxdata import NXdata
 from ._detector_data import NXevent_data
 from ._common import to_plain_index
+
+
+class EventSelector:
+    """A proxy object for creating and NXdetector based on a selection of events.
+    """
+    def __init__(self, detector):
+        self._detector = detector
+
+    def __getitem__(self, select) -> NXdetector:
+        """Return an NXdetector based on a selection (slice) of events."""
+        det = copy(self._detector)
+        det._event_select = select
+        return det
 
 
 class NXdetector(NXobject):
@@ -13,6 +31,10 @@ class NXdetector(NXobject):
     is used to map event do detector pixels. Otherwise this returns event data in the
     same format as NXevent_data.
     """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._event_select = tuple()
+
     @property
     def shape(self) -> List[int]:
         if self._is_events:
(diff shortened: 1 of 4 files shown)


jl-wynen:
Wouldn't it be more natural if the API were
```.py
event_data = det['pulse', :1000][:]
```
?
Or do we not consider the events as part of the dims of `NXdetector`?

SimonHeybrock:
> Wouldn't it be more natural if the API were
> 
> ```python
> event_data = det['pulse', :1000][:]
> ```
> 
> ? Or do we not consider the events as part of the dims of `NXdetector`?

It would be more *convenient*, but `det.dims == ['detector_number']`, so it would be very odd to support slicing a dimension that is not a dimension of the object.

jl-wynen:
> It would be more convenient, but det.dims == ['detector_number'], so it would be very odd to support slicing a dimension that is not a dimension of the object.

Good point

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -6,13 +11,30 @@
 from ._common import to_plain_index
 
 
+class EventSelector:
+    """A proxy object for creating and NXdetector based on a selection of events.
```
jl-wynen, line 15:
```suggestion
    """A proxy object for creating an NXdetector based on a selection of events.
```

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -6,13 +11,30 @@
...
+    """
+    def __init__(self, detector):
+        self._detector = detector
+
+    def __getitem__(self, select) -> NXdetector:
```
jl-wynen, line 20:
```suggestion
    def __getitem__(self, select: ScippIndex) -> NXdetector:
```
? Or is the type different here?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -76,6 +98,20 @@ def _nxbase(self) -> NXdata:
...
+        if self._is_events:
+            return self._nxbase
+
+    @property
+    def select_events(self) -> EventSelector:
+        """
+        Return a proxy object for selecting a slice of the underlying NXevent_data
+        group, while keeping wrapping the NXdetector.
+        """
+        return EventSelector(self)
```
jl-wynen, line 113: Should this raise an exception if the detector contains histogrammed data?

jl-wynen (approved):

PR #278, opened 2022-02-14
Test time axis interpolation and trim start time

SimonHeybrock:
The current implementation of loading NXtransformations uses extrapolation to fill in values before a log of transformation, i.e., while usually the *previous* log value is used (since everything after a motor move will see the effect of the translation/rotation), the *next* log value is used if the log for a transformation starts after that of another transformation. This risk some hard to spot bugs. Overall, we have the following options:

1. Keep current behavior, assuming that data acquisition logs transformation at the run start, or in small steps, i.e., error is likely small.
2. Use identity, which also risks bugs, maybe more visibly
3. Use NaN. This is inconvenient for the user.
4. Trim the time-dependence, i.e., start the transformation at the time when all transforms in the chain have a valid value.

Since 1.) is *usually* a good guess, but risks big hidden problems in some cases, we choose 4.): If good assumptions for 1.) apply, then the result is nearly identical, but it avoids the issue of big hidden bugs.

Also adding a test that exercises time interpolation.

### DIFF
--- a/src/scippneutron/file_loading/_transformations.py
+++ b/src/scippneutron/file_loading/_transformations.py
@@ -101,7 +101,11 @@ def get_full_transformation_matrix(group: Group, nexus: LoadFromNexus) -> sc.Dat
                 transform, xnew) * _interpolate_transform(total_transform, xnew)
         else:
             total_transform = transform * total_transform
-
+    if isinstance(total_transform, sc.DataArray):
+        time_dependent = [t for t in transformations if isinstance(t, sc.DataArray)]
+        times = [da.coords['time'][0] for da in time_dependent]
+        latest_log_start = sc.reduce(times).max()
+        return total_transform['time', latest_log_start:].copy()
     return total_transform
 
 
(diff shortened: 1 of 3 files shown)


jl-wynen:
> the next log value is used if the log for a transformation starts after that of another transformation

I don't understand what you are saying. When there are multiple transformations, does it use the value of the later one even for times before that transformation?

> 4. Trim the time-dependence, i.e., start the transformation at the time when all transforms in the chain have a valid value.

Again, I don't understand what you want to say. This sounds like you want to make transformations partially non-time-dependent. How would that work?

Tom-Willemsen (approved):

jl-wynen:
Ok, I understand what the code does from the unit test. But how would this be used? Users are going to get transforms that are invalid for part of their data. Is this the point? Do you want users to patch the transforms manually if need be? This only works if `load_nexus` does not apply any transformations itself. Otherwise, it would be impossible to load data.

SimonHeybrock:
> Ok, I understand what the code does from the unit test. But how would this be used? Users are going to get transforms that are invalid for part of their data. Is this the point?

Yes, kind of. If we do not know where the detector bank was at a given time, we should *not* claim that we do and provide some "default" position.

> Do you want users to patch the transforms manually if need be? This only works if `load_nexus` does not apply any transformations itself. Otherwise, it would be impossible to load data.

I doubt that there generally is the need to patch anything. Typically all transformations should be recorded from the start of the run. This in mainly do indicate that, e.g., events before a certain time need to be dropped (and it is unlikely that the users knows a detector position for such a time).

jl-wynen (approved):

PR #281, opened 2022-02-16
Add NXobject.items

SimonHeybrock:
I don't know if this is a good way of implementing this? Better suggestions?

### DIFF
--- a/src/scippneutron/file_loading/nxobject.py
+++ b/src/scippneutron/file_loading/nxobject.py
@@ -5,7 +5,7 @@ import numpy as np
 import scipp as sc
 from enum import Enum, auto
 import functools
-from typing import List, Union, NoReturn, Any, Dict
+from typing import List, Union, NoReturn, Any, Dict, Tuple
 
 from ._nexus import LoadFromNexus
 from ._hdf5_nexus import LoadFromHdf5
@@ -155,6 +155,9 @@ class NXobject:
             self._make(v) if self._loader.is_group(v) else Field(v, self._loader)
             for v in self._loader.values(self._group)
         ]
+
+    def items(self) -> List[Tuple[str, Union[Field, '__class__']]]:
+        return list(zip(self.keys(), self.values()))
 
     @functools.lru_cache()
     def by_nx_class(self) -> Dict[NX_class, List['__class__']]:
--- a/tests/nexus_test.py
+++ b/tests/nexus_test.py
@@ -66,6 +66,20 @@ def test_nxobject_root(nexus_group: Tuple[Callable, LoadFromNexus]):
         assert set(root.keys()) == {'entry', 'monitor'}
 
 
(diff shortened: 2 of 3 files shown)


src/scippneutron/file_loading/nxobject.py
```diff
@@ -156,6 +156,9 @@ def values(self) -> List[Union[Field, '__class__']]:
             for v in self._loader.values(self._group)
         ]
 
+    def items(self) -> List[Tuple[str, Union[Field, '__class__']]]:
+        return list(zip(self.keys(), self.values()))
```
jl-wynen, line 160: How about?
```suggestion
    def items(self) -> Iterable[Tuple[str, Union[Field, '__class__']]]:
        yield from zip(self.keys(), self.values())
```

tests/nexus_test.py
```diff
@@ -66,6 +66,20 @@ def test_nxobject_root(nexus_group: Tuple[Callable, LoadFromNexus]):
         assert set(root.keys()) == {'entry', 'monitor'}
 
 
+def test_nxobject_items(nexus_group: Tuple[Callable, LoadFromNexus]):
+    resource, loader = nexus_group
+    with resource(builder_with_events_monitor_and_log())() as f:
+        root = nexus.NXroot(f, loader)
+        items = root.items()
```
jl-wynen, line 73:
```suggestion
        items = list(root.items())
```
If you accept my other suggestion.

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Not sure? Python `dict` returns a view, not a generator?

jl-wynen: Yes, but then you would have to implement a custom class.

I looked at `keys` and `values`. They also return lists. So for consistency, `items` could do the same. I don't think the lists get very long, do they?

SimonHeybrock: A couple hundred items maybe.

jl-wynen: Then I am not sure. What about adding `items` to the two implementations of `LoadFromNexus`?

SimonHeybrock: I would want to avoid touching those classes wherever possible, since I am not sure we want to keep them around in this form and shape, or at all.

SimonHeybrock: If nothing is *conceptually* wrong with returning a `list` then I would say any other concerns are premature optimization.

jl-wynen: I'd say it is conceptually wrong when comparing with the standard Python types because they are all lazy. But it fits into the rest of this class, so it is fine.

jl-wynen (approved):

PR #283, opened 2022-02-16
Update dependency to scipp 0.12

jl-wynen:
Only merge once scipp 0.12 is released!


SimonHeybrock (approved):
I think we can merge now, since CI in the base branch fails anyway.

SimonHeybrock:
Maybe that was wrong, since now we can't build locally any more...

PR #288, opened 2022-02-23
Prepare 0.5 release

jl-wynen:


docs/about/release-notes.rst
```diff
@@ -3,25 +3,20 @@
...
+------
 
 Features
 ~~~~~~~~
 
-Breaking changes
-~~~~~~~~~~~~~~~~
+* Added Utilities for unwrapping frames `#242 <https://github.com/scipp/scippneutron/pull/242>`_.
+* Added First draft of low-level utilities for loading NeXus files `#249 <https://github.com/scipp/scippneutron/pull/249>`_.
+* Multi-valued transforms are now loaded by :func:`scippneutron.load_nexus` `#267 <https://github.com/scipp/scippneutron/pull/267>`_.
```
Tom-Willemsen, line 14: Might be worth clarifying where/how multi-valued transforms get loaded from (which is `NXlog` groups in the nexus file)?

jl-wynen: Like this?
```.rst
* Multi-valued transforms are now loaded from ``NXlog``s in NeXus files by :func:`scippneutron.load_nexus` `#267 <https://github.com/scipp/scippneutron/pull/267>`_.
```

We need proper documentation of how we load files, what structure we assume, and how it is encoded in the output. But that is a separate issue for when we converge on a solution...

docs/about/release-notes.rst
```diff
@@ -3,25 +3,20 @@
 Release Notes
 =============
 
-v0.5.0 (unreleased)
---------------------
+v0.5.0
+------
```
nvaytet, line 7:
```suggestion
v0.5.0 (February 2022)
----------------------
```

docs/about/release-notes.rst
```diff
@@ -3,25 +3,20 @@
 Release Notes
```
nvaytet, line 3: Simon usually adds the `v0.6.0 (unreleased)` section before the release is made, but I think it's better to leave it out until after the release, as it looks cleaner in the docs pages for a specific release.

jl-wynen: My thoughts exactly

Tom-Willemsen (approved):

nvaytet (approved):

PR #289, opened 2022-02-23
Try mambabuild for releases

nvaytet:


nvaytet:
No speedup for OSX and windows, so stay with `conda build`.

PR #293, opened 2022-03-04
Load all NXdata axes

### LINKED ISSUE #292: Extend NXdata loading to interpret and load datasets referenced by `AXISNAME_indices` attributes
This will be required for loading, e.g., NuRF data.

SimonHeybrock:
Fixes #292, and attempts to load axes even if those attributes are not present, as recommended in https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-axisname-indices-attribute.

### DIFF
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -92,8 +92,7 @@ class NXdata(NXobject):
             # since legacy files do not set this attribute.
             indices = self.attrs.get(f'{name}_indices')
             if indices is None:
-                if self.attrs.get('axes', self._axes_default) is not None:
-                    axes = self.attrs.get('axes', self._axes_default)
+                if (axes := self.attrs.get('axes', self._axes_default)) is not None:
                     # Unlike self.dims we *drop* entries that are '.'
                     axes = [a for a in axes if a != '.']
                 elif 'axes' in self._signal.attrs:
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -117,7 +117,7 @@ class NXdata(NXobject):
                     if len(shape) != 0:
                         raise NexusStructureError("Could not determine axis indices")
             else:
-                dims = np.array(da.dims)[indices]
+                dims = [np.array(da.dims)[indices]]
             index = to_plain_index(dims, select, ignore_missing=True)
             da.coords[name] = self._loader.load_dataset(self._group,
                                                         name,
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -87,7 +87,8 @@ class NXdata(NXobject):
 
         for name in items:
             # Newly written files should always contain indices attributes, but the
-            # standard recommends that readers should also make "best effort" guess.
+            # standard recommends that readers should also make "best effort" guess
+            # since legacy files do not set this attribute.
             indices = self.attrs.get(f'{name}_indices')
             if indices is None:
                 if self.attrs.get('axes', self._axes_default) is not None:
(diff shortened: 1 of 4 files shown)


src/scippneutron/file_loading/nxdata.py
```diff
@@ -79,16 +80,50 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
...
+        skip += list(self.attrs.get('auxiliary_signals', []))
+        items = [k for k in self.keys() if isinstance(self[k], Field)]
+        items = [k for k in items if k not in skip]
+
+        for name in items:
+            # Newly written files should always contain indices attributes, but the
+            # standard recommends that readers should also make "best effort" guess
+            # since legacy files do not set this attribute.
+            indices = self.attrs.get(f'{name}_indices')
+            if indices is None:
```
jl-wynen, line 94: Do we want to start using this or does it hurt readability?
```suggestion
            if (indices := self.attrs.get(f'{name}_indices')) is None:
```

src/scippneutron/file_loading/nxdata.py
```diff
@@ -79,16 +80,50 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
...
+                if self.attrs.get('axes', self._axes_default) is not None:
+                    axes = self.attrs.get('axes', self._axes_default)
```
jl-wynen, line 96: Here, we can definitely use it, I think
```suggestion
                if (axes := self.attrs.get('axes', self._axes_default)) is not None:
```

src/scippneutron/file_loading/nxdata.py
```diff
@@ -79,16 +80,50 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
...
+                    shape = list(self[name].shape)
+                    dims = []
+                    for dim, l in da.sizes.items():
+                        if len(shape) == 0:
+                            break
+                        if shape[0] == l:
+                            dims += [dim]
+                            shape.pop(0)
+                    if len(shape) != 0:
+                        raise NexusStructureError("Could not determine axis indices")
```
jl-wynen, line 120:
```suggestion
                        raise NexusStructureError(f"Could not determine axis indices for {self}")
```

tests/nxdata_test.py
```diff
@@ -0,0 +1,137 @@
...
+    da.coords['xx2'] = da.data['yy', 1]
+    da.coords['yy'] = da.data['xx', 0]
+    builder.add_data(Data(name='data1', data=da))
+    with resource(builder)() as f:
+        data = nexus.NXroot(f, loader)['entry/data1']
+        loaded = data[...]
+        assert sc.identical(loaded, da)
+
+
+def test_guessed_dim_for_2d_coord_not_matching_axis_name(
```
jl-wynen, line 58: What about data with `shape=[2,2]`? If I read it correctly, the code would currently load this data but does not check if the indices are correct. I think it should raise an exception in this case.

src/scippneutron/file_loading/nxdata.py
```diff
@@ -79,16 +80,50 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
...
+            else:
+                dims = np.array(da.dims)[np.array(indices).flatten()]
+            index = to_plain_index(dims, select, ignore_missing=True)
+            da.coords[name] = self._loader.load_dataset(self._group,
+                                                        name,
+                                                        dimensions=dims,
+                                                        index=index)
+
```
jl-wynen, line 128: Pretty long and deeply indented function. Can you split it?

src/scippneutron/file_loading/nxdata.py
SimonHeybrock: I think it does hurt readability a bit. But probably it would just take time getting used to it.

src/scippneutron/file_loading/nxdata.py
SimonHeybrock: 👍

src/scippneutron/file_loading/nxdata.py
SimonHeybrock: Done.

tests/nxdata_test.py
SimonHeybrock: Changed, but I am not sure it is the right way. Wouldn't it make sense to assume that it is not transposed?

jl-wynen (approved):

PR #294, opened 2022-03-07
Load monitor positions

nvaytet:


src/scippneutron/file_loading/_positions.py
```diff
@@ -15,26 +16,39 @@ class PositionError(Exception):
     pass
 
 
+def _make_prefixed_names(name, prefix):
+    if prefix is None:
+        prefix = f"{name}_"
+    return prefix + "transform", prefix + "position", prefix + "base_position"
```
SimonHeybrock, line 22: Is this so you can support moving monitors? Is that really required?

nvaytet: I just copied the way it was done for other components, as you suggested. I didn't really think about moving components.

src/scippneutron/file_loading/_monitor_data.py
```diff
@@ -69,6 +71,12 @@ def load_monitor_data(monitor_groups: List[Group], nexus: LoadFromNexus) -> Dict
             else:
                 monitor = nxmonitor[()]
             monitor_name = group.name.split("/")[-1]
+            load_positions_of_components(groups=[group],
+                                         data=monitor,
+                                         name=monitor_name,
+                                         nx_class=nx_monitor,
+                                         nexus=nexus,
+                                         name_prefix="")
```
SimonHeybrock, line 79: I think I should add support for NXtransformations in the new API, so this bit can be moved in `NXmonitor` above.

src/scippneutron/file_loading/_positions.py
```diff
@@ -44,32 +58,35 @@ def load_positions_of_components(groups: List[Group],
...
 
 def _get_base_pos_and_transforms_of_component(
-        group: Group,
-        name: str,
-        nx_class: str,
-        nexus: LoadFromNexus,
-        default_position: Optional[np.ndarray] = None
-) -> Tuple[sc.Variable, sc.DataArray]:
+        group: Group, name: str, nx_class: str, nexus: LoadFromNexus,
+        default_position: ArrayLike) -> Tuple[sc.Variable, sc.DataArray]:
```
SimonHeybrock, line 89: 👍 on getting rid of the hidden default of [0,0,0]

tests/test_load_nexus.py
```diff
@@ -2026,6 +2029,40 @@ def test_load_monitors_with_event_mode_data(load_function: Callable):
                  dtype=sc.DType.float64))
 
 
+def test_load_monitor_with_transformation(load_function: Callable):
```
SimonHeybrock, line 2032: You said you had trouble with the unit test...? This looks ok at first sight.

nvaytet: The test passes when loading from nexus, fails when loading from json

SimonHeybrock (approved):

PR #296, opened 2022-03-07
Change `nexus.Field` dim labels to use parent class dims

SimonHeybrock:
Previously accessing a field of a Nexus class used default dim labels such as `['dim_0', 'dim_1']`.

- Infer actual labels from containing Nexus class, where possible.
- Use slicing with labels also for fields. Previously this used plain numpy syntax, despite having (default) field labels, which was inconsistent.

Generally (with exceptions), this ensures that properties such as data or coords of a loaded group:

```python
group[some_index].data
group[some_index].coords[name]
```

are identical to loading the corresponding fields:

```python
group[data][some_index]
group[name][some_index]
```

(modulo restricting `some_index` to only index dims that the field contains).

### DIFF
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -162,7 +162,7 @@ def _to_canonical_select(dims: List[str], select: ScippIndex) -> ScippIndex:
     return select
 
 
-def to_plain_index(dims: List[str], select: ScippIndex, ignore_missing: bool = False):
+def to_plain_index(dims: List[str], select: ScippIndex):
     """
     Given a valid "scipp" index 'select', return an equivalent plain numpy-style index.
     """
@@ -171,7 +171,7 @@ def to_plain_index(dims: List[str], select: ScippIndex, ignore_missing: bool = F
         raise ValueError("Cannot process index {select}.")
     index = [slice(None)] * len(dims)
     for key, sel in select.items():
-        if not ignore_missing and key not in dims:
+        if key not in dims:
             raise ValueError(
                 f"'{key}' used for indexing not found in dataset dims {dims}.")
         index[dims.index(key)] = sel
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -159,6 +159,8 @@ def _to_canonical_select(dims: List[str], select: ScippIndex) -> ScippIndex:
     elif isinstance(select, int) or isinstance(select, slice):
         check_1d()
         return {dims[0]: select}
+    if not isinstance(select, dict):
+        raise ValueError("Cannot process index {select}.")
     return select
 
 
@@ -167,8 +169,6 @@ def to_plain_index(dims: List[str], select: ScippIndex):
     Given a valid "scipp" index 'select', return an equivalent plain numpy-style index.
     """
     select = _to_canonical_select(dims, select)
-    if not isinstance(select, dict):
-        raise ValueError("Cannot process index {select}.")
     index = [slice(None)] * len(dims)
     for key, sel in select.items():
         if key not in dims:
--- a/src/scippneutron/file_loading/_common.py
+++ b/src/scippneutron/file_loading/_common.py
@@ -156,7 +156,7 @@ def to_plain_index(dims: List[str], select: ScippIndex, ignore_missing: bool = F
     if isinstance(select, tuple):
         check_1d()
         if len(select) != 1:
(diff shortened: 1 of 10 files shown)


src/scippneutron/file_loading/_common.py
```diff
@@ -136,43 +136,65 @@ def convert_time_to_datetime64(
...
         if len(select) != 1:
-            raise ValueError(f"Dataset a single dimension {dims}, "
+            raise ValueError(f"Dataset has single dimension {dims}, "
                              "but multiple indices {select} were specified.")
-        return select[0]
+        return {dims[0]: select[0]}
     elif isinstance(select, int) or isinstance(select, slice):
         check_1d()
+        return {dims[0]: select}
+    return select
```
jl-wynen, line 162: Is this fall-through case used for anything? `to_plain_index` checks for it using `if not isinstance(select, dict):`. `to_child_select` does not check the type but uses `select` like a dict.

It looks to me like `_to_canonical_select` should always return a `dict`.

src/scippneutron/file_loading/_common.py
```diff
@@ -136,43 +136,65 @@ def convert_time_to_datetime64(
...
+
+
+def to_plain_index(dims: List[str], select: ScippIndex):
+    """
+    Given a valid "scipp" index 'select', return an equivalent plain numpy-style index.
+    """
+    select = _to_canonical_select(dims, select)
+    if not isinstance(select, dict):
+        raise ValueError("Cannot process index {select}.")
```
jl-wynen, line 171:
```suggestion
        raise ValueError(f"Cannot process index {select}.")
```

src/scippneutron/file_loading/_common.py
```diff
@@ -136,43 +136,65 @@ def convert_time_to_datetime64(
...
                              "but multiple indices {select} were specified.")
-        return select[0]
+        return {dims[0]: select[0]}
     elif isinstance(select, int) or isinstance(select, slice):
         check_1d()
+        return {dims[0]: select}
+    return select
+
+
+def to_plain_index(dims: List[str], select: ScippIndex):
```
jl-wynen, line 165:
```suggestion
def to_plain_index(dims: List[str], select: ScippIndex) -> tuple:
```

tests/nexus_test.py
```diff
@@ -201,6 +201,22 @@ def test_field_properties(nexus_group: Tuple[Callable, LoadFromNexus]):
         assert field.unit == sc.Unit('ns')
 
 
+def test_field_dim_labels(nexus_group: Tuple[Callable, LoadFromNexus]):
```
jl-wynen, line 204: What is this test checking exactly? Are all these dim labels discovered via the parents? Are there any cases where it falls back to `dim_xxx`?

tests/nxdata_test.py
```diff
@@ -161,3 +190,20 @@ def test_auxiliary_signal_is_not_loaded_as_coord(nexus_group: Tuple[Callable,
...
+    da.coords['xx'] = da.data['yy', 0]
+    da.coords['xx2'] = da.data['yy', 1]
+    da.coords['yy'] = da.data['xx', 0]
+    builder.add_data(Data(name='data1', data=da))
+    with resource(builder)() as f:
+        data = nexus.NXroot(f, loader)['entry/data1']
+        assert sc.identical(data['xx', :2].data, data['signal1']['xx', :2])
+        assert sc.identical(data['xx', :2].coords['xx'], data['xx']['xx', :2])
+        assert sc.identical(data['xx', :2].coords['xx2'], data['xx2']['xx', :2])
+        assert sc.identical(data['xx', :2].coords['yy'], data['yy'][:])
```
jl-wynen, line 209: This looks like it tests slicing and loading of signal + coordinates. But the name claims it is about dims. What am I missing?

SimonHeybrock: You are right that it also compares the data, but it is testing that the dim labels after loading the whole class match those of loading individual fields.

tests/nexus_test.py
SimonHeybrock: Yes, defined via parent. No fallbacks happening here. I think I probably need to make a non-standard-compliant file to test that...

SimonHeybrock: I added a test (where dim-deduction fails) in `nxdata_test.py`.

I also changed the behavior of the loader to continue (and warn) in this case, skipping the addition of the conflicting coords.

jl-wynen (approved):

PR #298, opened 2022-03-09
Fix data streaming resource leak

Tom-Willemsen:
Fixes a memory/thread leak in data streaming code when exiting a notebook while a streaming task in in progress. The code can still leak if the cell containing the streaming code is re-run but this seems much harder to fix and I have not come up with a good solution for this.

### DIFF
--- a/docs/developer/data-stream.rst
+++ b/docs/developer/data-stream.rst
@@ -290,17 +290,20 @@ Try using ``scippneutron.data_stream``, for example
 
         import asyncio
         import scippneutron as scn
-        from scippneutron.data_streaming.data_stream import StartTime, StopTime
+        from scippneutron.data_streaming.data_stream import StartTime
 
         async def my_stream_func():
             detector_ids = sc.Variable(dims=["detector_id"],
                                        values=np.arange(32*288).astype(np.int32))
-            async for data in scn.data_stream('localhost:9092', run_info_topic="AMOR_runInfo",
-                                              start_time=StartTime.START_OF_RUN, stop_time=StopTime.END_OF_RUN,
+            async for data in scn.data_stream('localhost:9092',
+                                              run_info_topic="AMOR_runInfo",
+                                              start_time=StartTime.START_OF_RUN,
                                               interval=2. * sc.units.s):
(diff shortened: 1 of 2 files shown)


docs/developer/data-stream.rst
```diff
@@ -290,17 +290,20 @@ Try using ``scippneutron.data_stream``, for example
...
-                                              start_time=StartTime.START_OF_RUN, stop_time=StopTime.END_OF_RUN,
+            async for data in scn.data_stream('localhost:9092',
+                                              run_info_topic="AMOR_runInfo",
+                                              start_time=StartTime.START_OF_RUN,
                                               interval=2. * sc.units.s):
                 events = sc.bin(data, groups=[detector_ids])
                 counts = events.bins.sum()
-                plot_data.values = plot_data.values + sc.fold(sc.flatten(counts, to="counts"), dim='counts', sizes={'y': 288, 'x': 32}).values
+                if "tof" in counts.dims:
+                    counts = counts["tof", 0].copy()
```
Tom-Willemsen, line 305: This is a hack because we sometimes get the events binned into a (single) ToF bin and sometimes do not. I haven't yet figured out what the cause of this difference is. This is not the right long-term fix.

src/scippneutron/data_streaming/data_stream.py
```diff
@@ -225,12 +225,17 @@ async def _data_stream(
     # https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
     # pytorch docs mention Queue problem:
     # https://pytorch.org/docs/stable/notes/multiprocessing.html
+    #
+    # Note also that daemonising this Process is important so that resources are
+    # properly freed when restarting the notebook kernel (or shutting down the
+    # notebook entirely).
```
jl-wynen, line 231: Does this affect usage in a Python script instead of notebook?

Tom-Willemsen: The resource leak would also have happened from a python script if killing the script via `SIGKILL` or similar. But I don't believe the change has any *negative* impact on running outside a notebook, if that's what you mean.

nvaytet (approved):

PR #299, opened 2022-03-10
Refactor to remove old NXlog loading code

SimonHeybrock:
First step for moving beyond the old Nexus loading code. Whether the added code in `_load_data` will stay like this, or will be refactored later to use a build is not clear yet. Existing tests pass, so I have to assume that this implementation is equivalent.

Similar steps for other Nexus classes will follow, but some classes are still missing or incomplete. I will try to do this step by step.

### DIFF
--- a/src/scippneutron/file_loading/load_nexus.py
+++ b/src/scippneutron/file_loading/load_nexus.py
@@ -2,15 +2,13 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 # @author Matthew Jones
 import json
-
 import scipp as sc
 
-import scipp
+from ..nexus import NXroot, NX_class
 
-from ._common import Group, MissingDataset
+from ._common import Group, MissingDataset, BadSource
 from ._detector_data import load_detector_data
 from ._monitor_data import load_monitor_data
-from .nxlog import load_logs
 from ._hdf5_nexus import LoadFromHdf5
 from ._json_nexus import LoadFromJson, get_streams_info, StreamInfo
 from ._nexus import LoadFromNexus, ScippData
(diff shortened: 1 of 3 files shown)


src/scippneutron/file_loading/load_nexus.py
```diff
@@ -190,11 +188,22 @@ def add_metadata(metadata: Dict[str, sc.Variable]):
             else:
                 loaded_data[key] = value
 
+    # Note: Currently this wastefully walks the tree in the file a second time.
+    root = NXroot(nexus_file, nexus)
+    classes = root.by_nx_class()
```
jl-wynen, line 193: Would it help performance if we could select a class here instead of creating a dict of all and then extracting only one? It would still have to walk the entire tree but maybe it would need to load fewer entries?

src/scippneutron/file_loading/load_nexus.py
```diff
@@ -190,11 +188,22 @@ def add_metadata(metadata: Dict[str, sc.Variable]):
...
     if groups[nx_entry]:
         add_metadata(_load_title(groups[nx_entry][0], nexus))
         add_metadata(_load_start_and_end_time(groups[nx_entry][0], nexus))
-    if groups[nx_log]:
-        add_metadata(load_logs(groups[nx_log], nexus))
+
+    logs = {}
+    for name, log in classes.get(NX_class.NXlog, {}).items():
+        try:
+            logs[name] = sc.scalar(log[()])
```
jl-wynen, line 202: Are logs always data arrays or datasets? Or can they also be variables (i.e.) we should not wrap them in a scalar.

tests/test_load_nexus.py
```diff
@@ -498,18 +498,13 @@ def test_loads_data_from_multiple_logs_with_same_name(load_function: Callable):
...
-                           values_2)
-    elif np.allclose(loaded_data[name].data.values.values, values_2):
-        # Then the other log should be
-        assert np.allclose(loaded_data[f"entry_{name}"].data.values.values, values_1)
-    else:
-        assert False
+    assert len(loaded_data) == 2
+    for key, da in loaded_data.items():
+        assert key.endswith('test_log')
+        if 'detector_0' in key:
```
jl-wynen, line 504: It would be nice to avoid loops and conditionals in a test. Are there guarantees about the order the logs are loaded in?

src/scippneutron/file_loading/load_nexus.py
SimonHeybrock: That would probably be worse, since we eventually *do* want to process all or most classes.

src/scippneutron/file_loading/load_nexus.py
SimonHeybrock: Even if they would be variables they may have conflicting dims.

tests/test_load_nexus.py
SimonHeybrock: No, as you can see from the old test, it even had the *naming* depend on the order of how things were loaded: It tested both cases for different naming schemes. By version has consistent naming (but uses longer names, including the full group path).

jl-wynen (approved):

PR #301, opened 2022-03-11
Use NXmonitor in load_nexus

SimonHeybrock:
Next step after #299 to remove old Nexus loader code. There are some temporary duplications of bin handling and position loading that should disappear in the future, e.g., when we have implemented `class NXtransformation` (which I may tackle in the next step).

Note that I changed the behavior to not store the transformation and base position. Only final position is stored. Since monitors typically do not move this should be sufficient.


tests/test_load_nexus.py
```diff
@@ -2003,13 +2003,6 @@ def test_load_monitors_with_event_mode_data(load_function: Callable):
     mon_1_events = load_function(builder)["monitor1"].data.values
     mon_2_events = load_function(builder)["monitor2"].data.values
 
-    assert sc.identical(
```
jl-wynen, line 2006: Why did you remove those tests?

tests/test_load_nexus.py
```diff
@@ -2049,13 +2042,6 @@ def test_load_monitor_with_transformation(load_function: Callable):
 
     assert sc.identical(loaded.coords["position"],
                         sc.vector(value=[1., 2., 9.5], unit="m"))
-    assert sc.identical(loaded.attrs["base_position"],
-                        sc.vector(value=[0, 0, 0], unit="m"))
-    assert sc.identical(
-        loaded.attrs["transform"].value,
-        sc.spatial.affine_transform(value=[[1., 0., 0., 1.], [0., 1., 0., 2.],
-                                           [0., 0., 1., 9.5], [0., 0., 0., 1.]],
-                                    unit='m'))
```
jl-wynen, line 2058: Didn't we say that we always load and output the base position and transform?

tests/test_load_nexus.py
SimonHeybrock: I was not 100% sure about this either, but I think monitors having a detector ID is an tradition adopted from Mantid. `NXmonitor` does not really appear to define this.

tests/test_load_nexus.py
SimonHeybrock: I think that may become necessary for time-dependent transformations. In general I think we may actually want to avoid it, but it is complicated. Best to discuss face-to-face.

tests/test_load_nexus.py
jl-wynen: Ok. The IDs that the old code defined were kind of bad anyway because could clash with IDs of the actual detector.

tests/test_load_nexus.py
SimonHeybrock: I'd suggest to leave this as it is for now. We will later be able to build custom load functions that produce different results (such as keeping the transforms or base positions). Bear with me.

jl-wynen: Ok. This discussion affects components outside of this PR anyway.

jl-wynen (approved):

PR #302, opened 2022-03-11
Load positions and UB matrices using new API

SimonHeybrock:
- Do not store `base_position` and `transformations` any more. Instead, we preserve these within the attributes, e.g., `da.attrs['sample']` will have an attribute giving the transformation (named `depends_on`).
- Add NXsample and NXsource (very basic)
- Some hacks for backwards-compatibility are added directly in `load_nexus.py`. I hope we can remove some of these later on.
- Remove a whole bunch of code.


src/scippneutron/file_loading/load_nexus.py
```diff
@@ -207,21 +198,36 @@ def add_metadata(metadata: Dict[str, sc.Variable]):
...
                     warn(f"Skipped loading {group.name} due to:\n{e}")
         add_metadata(items)
+        return loaded_groups
 
     load_and_add_metadata(classes.get(NX_class.NXlog, {}))
     load_and_add_metadata(classes.get(NX_class.NXmonitor, {}), _monitor_to_canonical)
+    for name, tag in {'sample': NX_class.NXsample, 'source': NX_class.NXsource}.items():
+        comps = classes.get(tag, {})
+        comps = load_and_add_metadata(comps)
+        attrs = loaded_data if isinstance(loaded_data,
```
nvaytet, line 217: Not sure I follow what is going on here. If `loaded_data` is a `Dataset`, both `attrs` and `coords` are set to the Dataset itself. Can you explain a little?

src/scippneutron/file_loading/nxsource.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+import scipp as sc
+from ._common import to_plain_index
+from .nxobject import NXobject, ScippIndex
+
+
+class NXsource(NXobject):
```
nvaytet, line 9: This looks very similar to `NXsample` above.
On one hand, I want to say the the duplicate code could be made common, to reduce the number of places that need to be touched in future changes.
On the other hand, there is not too much code, and trying to remove code duplication often makes code harder to read (and it partly also got us in the situation we are in with the nexus loader).
So I am not sure...

tests/test_load_nexus.py
```diff
@@ -1113,17 +1096,9 @@ def test_loads_component_position_with_multiple_multi_valued_log_transformations
...
-
-    expected_transformed_positions = np.array([[0, 0, 6], [0, 0, 60]])
-
-    assert np.allclose((loaded_data[f"{component_name}_transform"].value *
-                        loaded_data[f"{component_name}_base_position"]).values,
-                       expected_transformed_positions)
+    comp = loaded_data[component_name].value
     assert sc.identical(
-        loaded_data[f"{component_name}_transform"].value.coords["time"],
+        comp.coords['depends_on'].value.coords["time"],
```
nvaytet, line 1101: :+1:

src/scippneutron/file_loading/load_nexus.py
```diff
@@ -207,21 +198,36 @@ def add_metadata(metadata: Dict[str, sc.Variable]):
 
     def load_and_add_metadata(groups, process=lambda x: x):
         items = {}
+        loaded_groups = []
         for name, group in groups.items():
             try:
                 items[name] = sc.scalar(process(group[()]))
-            except (RuntimeError, KeyError, BadSource, SkipSource) as e:
+                loaded_groups.append(name)
+            except Exception as e:
```
jl-wynen, line 206: We need a better mechanism. This one can hide bugs in our code. (The previous `RuntimeError` wasn't much better.)

src/scippneutron/file_loading/load_nexus.py
SimonHeybrock: This is what `load_nexus` does: There may be no NXdetector or NXevent_data in the file that can be loaded, so it cannot return a `DataArray`. This it falls back to making a `Dataset`.

This is unchanged.

src/scippneutron/file_loading/nxsource.py
SimonHeybrock: We will see later, when these classes are extended to support more, whether there is common code to be extracted or not.

src/scippneutron/file_loading/load_nexus.py
SimonHeybrock: Agree. I think a generic `load_nexus` can probably not do much better, that is why we want to add a builder for creating more specific loaders that can avoid the need to continue loading despite all kinds of errors.

SimonHeybrock:
More comments on this?

jl-wynen (approved):

PR #304, opened 2022-03-14
Add NXdisk_chopper and refactor to new API

SimonHeybrock:
... and another step, removing old code.

### DIFF
--- a/src/scippneutron/file_loading/nxdisk_chopper.py
+++ b/src/scippneutron/file_loading/nxdisk_chopper.py
@@ -1,0 +1,26 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+import scipp as sc
+from ._common import to_plain_index
+from .nxobject import NXobject, ScippIndex
+
+
+class NXdisk_chopper(NXobject):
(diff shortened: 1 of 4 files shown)


src/scippneutron/file_loading/nxdisk_chopper.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @property
+    def dims(self):
+        return []
+
+    def _getitem(self, select: ScippIndex) -> sc.Dataset:
+        ds = sc.Dataset()
+        for name in ['distance', 'rotation_speed']:
+            if (field := self.get(name)) is not None:
+                ds[name] = field[select]
```
jl-wynen, line 21: This conditional loading construct is quite similar to what you did for NXsample and NXsource. Would it make sense to have a more general mechanism (maybe in NXobject) that conditionally loads field if they exist in the file? This might help with customisation. A user might pass a list of expected fields to the constructor of the NX* object.
Obviously, we still need more specialised implementations for cases where the data needs to be transformed after loading or it needs to trigger loading additional data.

SimonHeybrock: Probably, but later. This is just refactoring to remove the old API.

jl-wynen (approved):

PR #305, opened 2022-03-14
Refactor detector load

SimonHeybrock:
This is the largest change of the series refactoring steps. The old code to load detector data (including event data) is removed and replaced by the new API. `load_events` now directly contains some mapping code to preserve the old "standard" layout. This part will most likely be refactored/removed in the future, once we have a new mechanism.

There are a couple of things to note here:

- There were a couple of minor behavior changes. For example, `bin_by_pixel` was removed. I believe we never made this "public" and no one used it, so this should be ok. The same goes (I think) for the other changes.
- I removed or disabled certain tests. Some seem to be related to streaming behavior which I do not know, but the approaches seemed peculiar. I would prefer to move to a better mechanism, rather than making, e.g., our `class NXdetector` hacky (in on example: support loading neither data nor event data).
- Fixed bug of loading Y and Z pixel offset in `class NXdetector`
- Checking types of `event_id` and `detector_number` fields to be integer. Since Nexus does generally specify the allowed types for a field, I think we should extend `class Field` with a mechanism to do this check automatically. I will address this separately.

We should test this manually for files we know that `load_nexus` could previously load. I believe not all of such files are used in CI, and I am not certain how good the test coverage is. If you know of such files, either test yourself, or point me to the file.


src/scippneutron/file_loading/load_nexus.py
```diff
@@ -69,16 +72,14 @@ def _load_start_and_end_time(entry_group: Group, nexus: LoadFromNexus) -> Dict:
...
-               bin_by_pixel: bool = True) -> Optional[ScippData]:
+               quiet=True) -> Optional[ScippData]:
     """
     Load a NeXus file and return required information.
 
     :param data_file: path of NeXus file containing data to load
     :param root: path of group in file, only load data from the subtree of
       this group
     :param quiet: if False prints some details of what is being loaded
-    :param bin_by_pixel: if True, bins the loaded detector data by pixel. If False, bins
```
jl-wynen, line 81: Also remove the line below

src/scippneutron/file_loading/load_nexus.py
```diff
@@ -152,15 +147,83 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
...
+        # Add single tof bin
+        loaded_data = sc.DataArray(loaded_data.data.fold(dim='detector_id',
+                                                         sizes={
+                                                             'detector_id': -1,
+                                                             'tof': 1
+                                                         }),
+                                   coords={k: v
+                                           for k, v in loaded_data.coords.items()},
+                                   attrs={k: v
+                                          for k, v in loaded_data.attrs.items()})
```
jl-wynen, line 222:
```suggestion
                                   coords=dict(loaded_data.coords.items()),
                                   attrs=dict(loaded_data.attrs.items()))
```

tests/test_load_nexus.py
```diff
@@ -631,13 +631,13 @@ def test_loads_pixel_positions_with_event_data(load_function: Callable):
...
-    ]).T / 1_000  # Divide by 1000 for mm to metres
+    ]).T
     assert np.allclose(loaded_data.coords['position'].values, expected_pixel_positions)
-    assert loaded_data.meta[
-        'position'].unit == sc.units.m, \
-        "Expected positions to be converted to metres"
+    assert loaded_data.meta['position'].unit == 'mm'
 
 
+@pytest.mark.skip(reason="It is unclear of NXdetector with either data or event_data "
```
jl-wynen, line 639: ?
```suggestion
@pytest.mark.skip(reason="It is unclear if NXdetector with neither data or event_data "
```

tests/test_load_nexus.py
```diff
@@ -681,175 +681,6 @@ def test_loads_pixel_positions_without_event_data(load_function: Callable):
...
-        np.concatenate((y_pixel_offset_1, y_pixel_offset_2.flatten())),
-        np.concatenate((z_pixel_offset_1, z_pixel_offset_2.flatten()))
-    ]).T / 1_000  # Divide by 1000 for mm to metres
-    assert np.allclose(loaded_data.coords['position'].values, expected_pixel_positions)
-    assert loaded_data.meta[
-               'base_position'].unit == sc.units.m, \
-        "Expected positions to be converted to metres"
-
-
-def test_loads_event_data_when_missing_from_some_detectors(load_function: Callable):
```
jl-wynen, line 736: Do all or no detectors have to have event data now?

tests/test_load_nexus.py
```diff
@@ -2005,66 +1836,9 @@ def test_load_monitor_with_transformation(load_function: Callable):
                         sc.vector(value=[1., 2., 9.5], unit="m"))
 
 
-def test_load_raw_detector_data_from_nexus_file(load_function: Callable):
```
jl-wynen, line 2008: Why did you remove this test?

jl-wynen (commented):
Works with some MCStas data that was converted to NeXus by Mantid.

tests/test_load_nexus.py
SimonHeybrock: Basically the old code simply loaded the `NXdetector` without `NXevent_data` as if it had event lists with 0 counts. See this snippet below:

```python
   # The event data from detector_1 has been loaded
    counts_on_detectors = loaded_data.bins.sum()
    expected_counts = np.array([[0], [2], [1], [2], [0], [0], [0], [0]])
```

I don't think this is something we should do, so this was removed.

tests/test_load_nexus.py
SimonHeybrock: Because the `bin_by_pixel` option was removed.

jl-wynen: Ok. Then I think the test name was misleading. But that does not matter anymore

SimonHeybrock: I think "raw" was probably meant to imply "as it is in the file"

src/scippneutron/file_loading/_detector_data.py
```diff
@@ -2,30 +2,17 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
```
nvaytet, line 2: I don't have much to add. Tested with the amor `sample.nxs` and `reference.nxs` and it worked fine.

SimonHeybrock:
Do not merge yet. I found this fails on PG3 files.

nvaytet (approved):

nvaytet:
strange, I thought the docs (e.g. instrument view) were using the PG3 files, so how did the builds pass?

SimonHeybrock:
Those are using `load`, i.e., Mantid, not `load_nexus`.

PR #307, opened 2022-03-16
Refactor NXdetector

SimonHeybrock:
NXdetector suffers from ambiguities in definition of dimension labels and absence/presence of signal datasets and/or event data. The current implementation had gotten quite convoluted. This is an attempt to improve this.

By using more functionality from NXdata this thereby extends support of automatically loading additional datasets with metadata.

I am adding support for interpreting `pixel_id` as alternative name for `detector_number`. This is used at SNS, and I am unhappy about hard-coding this. I am not sure we want to keep this, but it is useful (and helped to uncover a number of issues with the old implementation) until we made up our minds about a potentially better solution.

The diff if `nxdetector.py` is probably not very enlightening, I recommend looking at the full file.

### DIFF
--- a/src/scippneutron/file_loading/load_nexus.py
+++ b/src/scippneutron/file_loading/load_nexus.py
@@ -184,7 +184,10 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
             det = det.flatten(to='detector_id')
             det.bins.coords['tof'] = det.bins.coords.pop('event_time_offset')
             det.bins.coords['pulse_time'] = det.bins.coords.pop('event_time_zero')
-            det.coords['detector_id'] = det.coords.pop('detector_number')
+            if 'detector_number' in det.coords:
+                det.coords['detector_id'] = det.coords.pop('detector_number')
+            else:
+                det.coords['detector_id'] = det.coords.pop('pixel_id')
             if 'pixel_offset' in det.coords:
                 add_position_and_transforms_to_data(
                     data=det,
--- a/src/scippneutron/file_loading/load_nexus.py
+++ b/src/scippneutron/file_loading/load_nexus.py
@@ -121,6 +121,24 @@ def _monitor_to_canonical(monitor):
     return da
 
 
+def _zip_pixel_offset(da: sc.DataArray) -> sc.DataArray:
+    """Remove [xyz]_pixel_offset and replace them by single 'pixel_offset' coord.
+
+    Returns unchanged data array if x_pixel_offset is not found.
+    """
+    if 'x_pixel_offset' not in da.coords:
+        return da
+    x = da.coords.pop('x_pixel_offset')
+    offset = sc.zeros(dims=da.dims, shape=da.shape, unit=x.unit, dtype=sc.DType.vector3)
+    offset.fields.x = x.to(dtype='float64', copy=False)
+    if (y := da.coords.pop('y_pixel_offset', None)) is not None:
+        offset.fields.y = y.to(dtype='float64', unit=x.unit, copy=False)
+    if (z := da.coords.pop('z_pixel_offset', None)) is not None:
+        offset.fields.z = z.to(dtype='float64', unit=x.unit, copy=False)
+    da.coords['pixel_offset'] = offset.rename_dims(dict(zip(offset.dims, da.dims)))
+    return da
+
+
 def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
                nexus: LoadFromNexus, quiet: bool) \
         -> Optional[ScippData]:
@@ -162,6 +180,7 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
     for name, group in detectors.items():
         try:
             det = group[()]
+            det = _zip_pixel_offset(det)
             det = det.flatten(to='detector_id')
             det.bins.coords['tof'] = det.bins.coords.pop('event_time_offset')
             det.bins.coords['pulse_time'] = det.bins.coords.pop('event_time_zero')
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 # @author Simon Heybrock
+from __future__ import annotations
 from typing import List, Union
 from warnings import warn
 import scipp as sc
@@ -16,16 +17,21 @@ class NXdata(NXobject):
                  group: Group,
                  loader: LoadFromNexus = LoadFromHdf5(),
                  signal_name_default: str = None,
-                 signal=None,
+                 signal_override: Union[Field, EventField] = None,
                  axes: List[str] = None,
                  skip: List[str] = None):
         """
-        :param signal: Default signal name used, if no `signal` attribute found in file.
+        :param signal_name_default: Default signal name used, if no `signal`
+            attribute found in file.
+        :param signal_override Signal field-like to use instead of trying to read
+            signal from the file. This is used when there is no signal or to provide
+            a signal computed from NXevent_data
         :param axes: Default axes used, if no `axes` attribute found in file.
+        :param skip: Names of fields to skip when loading coords.
         """
         super().__init__(group, loader)
         self._signal_name_default = signal_name_default
-        self._signal_override = signal
+        self._signal_override = signal_override
         self._axes_default = axes
         self._skip = skip if skip is not None else []
 
@@ -71,7 +77,7 @@ class NXdata(NXobject):
             return f'{self._signal_name_default}_errors'
 
     @property
-    def _signal(self) -> Field:
+    def _signal(self) -> Union[Field, EventField]:
         if self._signal_override is not None:
             return self._signal_override
         return self[self._signal_name]
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -18,7 +18,7 @@ class NXdata(NXobject):
             group: Group,
             loader: LoadFromNexus = LoadFromHdf5(),
             signal_name_default: str = None,
-            signal_override: Union[Field, EventField] = None,  # noqa: F821
+            signal_override: Union[Field, _EventField] = None,  # noqa: F821
             axes: List[str] = None,
             skip: List[str] = None):
         """
@@ -78,7 +78,7 @@ class NXdata(NXobject):
             return f'{self._signal_name_default}_errors'
 
     @property
-    def _signal(self) -> Union[Field, EventField]:  # noqa: F821
+    def _signal(self) -> Union[Field, _EventField]:  # noqa: F821
         if self._signal_override is not None:
             return self._signal_override
         return self[self._signal_name]
--- a/src/scippneutron/file_loading/nxdata.py
+++ b/src/scippneutron/file_loading/nxdata.py
@@ -31,12 +31,17 @@ class NXdata(NXobject):
     def shape(self) -> List[int]:
         return self._signal.shape
 
-    @property
-    def dims(self) -> List[str]:
+    def _get_group_dims(self) -> Union[None, List[str]]:
         # Apparently it is not possible to define dim labels unless there are
         # corresponding coords. Special case of '.' entries means "no coord".
         if (axes := self.attrs.get('axes', self._axes_default)) is not None:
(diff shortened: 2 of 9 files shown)


src/scippneutron/file_loading/nxdetector.py
```diff
@@ -24,6 +23,78 @@ def __getitem__(self, select: ScippIndex) -> NXdetector:
...
+        if self._detector_number is None:
+            if select not in (Ellipsis, tuple()) and select != slice(None):
+                raise NexusStructureError(
+                    "Cannot load slice of NXdetector since it contains event data "
+                    "but no 'detector_number' field, i.e., the shape is unknown. "
+                    "Use ellipsis or an empty tuple to load the full detector.")
+            # Ideally we would prefer to use np.unique, but a quick experiment shows
+            # that this can easily be 100x slower, so it is not an option. In
+            # practice most files have contiguous event_id values within a bank
+            # (NXevent_data).
```
jl-wynen, line 73: How long is that in absolute time?
What is the risk of loading wrong detector numbers? Maybe we should warn.

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -24,6 +23,78 @@ def __getitem__(self, select: ScippIndex) -> NXdetector:
...
+        event_id = detector_number.flatten(to='event_id')
+        event_data.bins.coords['event_time_zero'] = sc.bins_like(
+            event_data, fill_value=event_data.coords['event_time_zero'])
+        # After loading raw NXevent_data it is guaranteed that the event table
+        # is contiguous and that there is no masking. We can therefore use the
+        # more efficient approach of binning from scratch instead of erasing the
+        # 'pulse' binning defined by NXevent_data.
+        event_data = sc.bin(event_data.bins.constituents['data'], groups=[event_id])
+        if self._detector_number is None:
+            event_data.coords['detector_number'] = event_data.coords.pop('event_id')
```
jl-wynen, line 92: Why does this store a detector number coord in this case but not in the other?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -87,26 +138,39 @@ def _is_events(self) -> bool:
...
+                      signal_name_default='data' if 'data' in self else None,
+                      signal_override=signal,
+                      skip=self._nxevent_data_fields)
 
     @property
     def events(self) -> Union[None, NXevent_data]:
         """Return the underlying NXevent_data group, None if not event data."""
-        if self._is_events:
-            return self._nxbase
+        if not self._is_events:
```
jl-wynen, line 168: I got confused when looking at `events` because I did not remember that `_is_events` checks that the group structure is correct. Maybe that logic would be better placed here and `_is_events` uses `self.events is not None`.

src/scippneutron/file_loading/load_nexus.py
```diff
@@ -162,10 +180,14 @@ def _load_data(nexus_file: Union[h5py.File, Dict], root: Optional[str],
     for name, group in detectors.items():
         try:
             det = group[()]
+            det = _zip_pixel_offset(det)
             det = det.flatten(to='detector_id')
             det.bins.coords['tof'] = det.bins.coords.pop('event_time_offset')
             det.bins.coords['pulse_time'] = det.bins.coords.pop('event_time_zero')
-            det.coords['detector_id'] = det.coords.pop('detector_number')
+            if 'detector_number' in det.coords:
+                det.coords['detector_id'] = det.coords.pop('detector_number')
```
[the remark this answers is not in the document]
SimonHeybrock: No, in general there is no such dim. The `detector_number` coord can be, e.g., 2-D. We deal with this below using `flatten`, which also gets rid of the original dim names.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -71,7 +71,12 @@ def __init__(self,
                  dims=None):
         self._dataset = dataset
         self._loader = loader
-        self._dims = [f'dim_{i}' for i in range(self.ndim)] if dims is None else dims
+        if dims is not None:
+            self._dims = dims
+        elif (axes := self.attrs.get('axes')) is not None:
+            self._dims = axes.split(',')
```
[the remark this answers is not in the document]
SimonHeybrock: I have not seen anything mentioned about this. Note that newer files use a list of strings, this branch is for legacy files.

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -24,6 +23,78 @@ def __getitem__(self, select: ScippIndex) -> NXdetector:
         return det
 
 
+class EventField:
```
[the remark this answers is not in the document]
SimonHeybrock: Not sure. It only makes sense in the context of NXdetector, in fact it is an implementation detail?

src/scippneutron/file_loading/nxdetector.py
```diff
@@ -24,6 +23,78 @@ def __getitem__(self, select: ScippIndex) -> NXdetector:
...
+        if self._detector_number is None:
+            return ['detector_number']
+        return self._detector_number.dims
+
+    @property
+    def shape(self):
+        if self._detector_number is None:
+            raise NexusStructureError(
+                "Cannot get shape of NXdetector since no 'detector_number' "
+                "field found but detector contains event data.")
```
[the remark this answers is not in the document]
SimonHeybrock: This is not possible without loading potentially GBytes of event data: When there is no detector number field it is determined automatically by looking at the range of values in `event_id`.

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: Very long.

There is no "risk" here, apart from risk of inserting empty bins. Or do you see another risk?

To be honest, I was tempted to remove this handling entirely, i.e., reject all NXdetector with event data that does not define IDs.

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: NXdata loads the `detector_number` field if it is present. If it is auto-generated we store it ourselves.

src/scippneutron/file_loading/nxobject.py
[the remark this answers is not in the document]
jl-wynen: Ok

src/scippneutron/file_loading/nxdetector.py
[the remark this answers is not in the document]
jl-wynen: Right, I thought it got returned by public methods of NXdetector.
So rename to `_EventField`?

src/scippneutron/file_loading/nxdetector.py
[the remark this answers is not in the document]
jl-wynen: Ok, makes sense, I think,

src/scippneutron/file_loading/nxdetector.py
jl-wynen: > To be honest, I was tempted to remove this handling entirely, i.e., reject all NXdetector with event data that does not define IDs.

Are there important legacy files that need this?

> Or do you see another risk?

No, this should be fine.

src/scippneutron/file_loading/nxdetector.py
jl-wynen: This depends on the parameters of `EventField` to be consistent. This code here checks `self._detector_number` but relies on `self._nxevent_data` to contain a detector number. So should the condition instead be `if 'detector_number' not in event_data.coords`?

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: > Are there important legacy files that need this?

I am not entirely sure. We *used* to need it for SNS files that use `pixel_id` instead of `detector_number`. But now we are also supporting the former, so at least this case has gone away.

Let us keep it for now but consider for removal later?

jl-wynen: Ok

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: Good idea. Managed to remove `_is_events`, and also changed some inconsistent behavior.

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: I do not understand your comment. `event_data.coords` will never contain `'detector_number'` since `NXevent_data` does not use such a field.

jl-wynen: Then I don't understand yours. If `event_data.coords` never contains detector_number, the two branches produce results with different coords.

SimonHeybrock: I think this is handled by the `fold` a couple lines below (and the `flatten` above): The `event_id` dimension becomes whatever dims the `detector_number` has. So if it is auto-generated then that basically just renamed `event_id` to `detector_number`. If it is in the file (and potentially multi-dim) we are using its dims, but the underlying NXdata will add the coord.

src/scippneutron/file_loading/nxdata.py
```diff
@@ -18,7 +18,7 @@ def __init__(
             group: Group,
             loader: LoadFromNexus = LoadFromHdf5(),
             signal_name_default: str = None,
-            signal_override: Union[Field, EventField] = None,  # noqa: F821
+            signal_override: Union[Field, _EventField] = None,  # noqa: F821
```
jl-wynen, line 21: I think this will still produce an error in type checkers because `_EventField` is never defined in this module.
Solutions would be (if `_EventField` should remain an implementation detail) to either have a common base class of `Field` and `_EventField` or, since we dropped py3.7, to use a [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol)

But I am fine with postponing this until we extract the scippnexus. We will probably have to do cleanup refactors anyway.

tests/nxdetector_test.py
```diff
@@ -26,8 +26,8 @@ def test_raises_if_no_data_found(nexus_group: Tuple[Callable, LoadFromNexus]):
             detector[...]
 
 
-def test_raises_if_data_and_event_data_found(nexus_group: Tuple[Callable,
-                                                                LoadFromNexus]):
+def test_loads_events_when_data_and_events_found(nexus_group: Tuple[Callable,
+                                                                    LoadFromNexus]):
```
jl-wynen, line 30: Should this case warn?
There should at least be clear documentation how it is handled.

SimonHeybrock: I was considering a warning, but it appears to be common to define the `signal` field as something that can be used to quickly plot/preview the NXdetector (you could, e.g., imagine SciCat doing this). So adding a warning would produce a lot of noise.

src/scippneutron/file_loading/nxdetector.py
jl-wynen: Now I am really confused. Does the output of this function contain a coord 'detector_number' in all cases? It does if `self._detector_number is None`.
If you say that it also contains it in the `else` case, then the condition should be `'detector_number' not in event_data.coords`, I think. If it does not, we are returning inconsistent results.

SimonHeybrock: It contains a `detector_number` coord *only* if it was auto-generated.

tests/nxdetector_test.py
jl-wynen: Ok

src/scippneutron/file_loading/nxdetector.py
jl-wynen: Why? Is it added somewhere else?

SimonHeybrock: By NXdata.

jl-wynen: Ok. I am not happy with this solution because it means that `NXdata` and `EventField` depend on each other to set up the coord correctly. But since it passes the test, I am fine with leaving it as it is.

jl-wynen (approved):

src/scippneutron/file_loading/nxdetector.py
SimonHeybrock: Indeed, I did not like that `NXdata` needs to distinguish `_signal` returning a `Variable` vs. `DataArray`. I had considering caching the auto-generated detector-number in `_EventField`, returning only the data variable here, and adding the det nums later by hand (after NXdata did its thing).

SimonHeybrock: This entire problem will disappear if we decide to remove the automatic-detector-number mechanism, so let us keep this for now.

PR #310, opened 2022-03-18
Begin wrapping JSON Nexus

SimonHeybrock:
This is the first actual step towards removal of `LoadFromHdf5` and `LoadFromJson`. We introduce wrappers for attributes and datasets. This is not fully complete, and the next main step will be a wrapper for groups. This PR illustrates the approach.

### DIFF
--- a/src/scippneutron/file_loading/_json_nexus.py
+++ b/src/scippneutron/file_loading/_json_nexus.py
@@ -254,13 +254,6 @@ class LoadFromJson:
         except KeyError:
             return dataset[_nexus_dataset]["dtype"]
 
-    @staticmethod
-    def get_shape(dataset: Dict) -> List:
-        """
-        The shape of the dataset
-        """
-        return np.asarray(dataset[_nexus_values]).shape
-
     def get_object_by_path(self, group: Dict, path_str: str) -> Dict:
         for node in filter(None, path_str.split("/")):
             group = self._get_child_from_group(group, node)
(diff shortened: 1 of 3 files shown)


src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -310,6 +273,53 @@ def contains_stream(group: Dict):
...
+        except MissingAttribute:
+            return False
+        return True
+
+    def __getitem__(self, name):
+        return _get_attribute_value(self._node, name)
+
+
+class JSONDataset:
+    def __init__(self, node: dict, loader: LoadFromJson):
```
SimonHeybrock, line 292: Currently this still keeps a reference to `LoadFromJson`, but this should disappear later when we absorb functionality fully into this class.

src/scippneutron/file_loading/nxobject.py
```diff
@@ -71,7 +69,8 @@ def __init__(self,
                  dataset: Dataset,
                  loader: LoadFromNexus = LoadFromHdf5(),
                  dims=None):
-        self._dataset = dataset
+        self._dataset = dataset if isinstance(loader, LoadFromHdf5) else JSONDataset(
+            dataset, loader)
```
SimonHeybrock, line 73: `Field` still needs the loader, but it should also disappear here later on in the refactoring.

src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -310,6 +273,53 @@ def contains_stream(group: Dict):
...
+        else:
+            return self._node.get(_nexus_path, '/')
+
+    @property
+    def file(self):
+        return self._node.file
+
+    @property
+    def shape(self):
+        return np.asarray(self._node[_nexus_values]).shape
```
jl-wynen, line 320: Is this really the best solution? Looks overkill to load the values just to get the shape.

jl-wynen (approved):
A good step forward.

src/scippneutron/file_loading/_json_nexus.py
SimonHeybrock: I don't think it actually loads anything: In the JSON case the entire "file" structure is just a nested dict in memory. Note that this is unchanged.

PR #312, opened 2022-03-21
Squeeze scalar attributes and fix f-string

SimonHeybrock:


nvaytet (approved):
Tested locally. Works :+1:

PR #313, opened 2022-03-21
Add "JSONGroup" and remove dependence on loader from new Nexus API

SimonHeybrock:
- Add JSONGroup (named _JSONGroup until the small legacy of the same name is removed).
- Remove need to track the "loader" in NXobject, Field, and friends.

This is the last but one big step in removing the older LoadFromHdf5 and LoadFromJson classes. There should now be a fair amount of code that can be moved and removed. Next, _JSONGroup and JSONDataset will be refactored to avoid the cumbersome reliance on LoadFromJson. I chose to not include this refactoring step (which will result in code removal) here, since I felt that the diff would be much harder to decipher. This implies that a number of details in this PR are temporary.

### DIFF
--- a/src/scippneutron/file_loading/_json_nexus.py
+++ b/src/scippneutron/file_loading/_json_nexus.py
@@ -316,17 +316,6 @@ class JSONNode:
         else:
             return JSONDataset(item, self._loader, parent=parent)
 
-    def __getitem__(self, name: str) -> Union[JSONDataset, _JSONGroup]:
-        if name.startswith('/') and name.count('/') == 1:
-            parent = self.file
-        elif '/' in name:
-            parent = self['/'.join(name.split('/')[:-1])]
-        else:
-            parent = self
-        if (item := self._loader.get_child_from_group(self._node, name)) is not None:
-            return self._as_group_or_dataset(item, parent)
-        raise KeyError(f"Unable to open object (object '{name}' doesn't exist)")
-
     @property
     def file(self):
         return self._file
@@ -353,8 +342,6 @@ class JSONDataset(JSONNode):
         return np.asarray(self._node[_nexus_values]).shape
 
     def __getitem__(self, index):
-        if isinstance(index, str):
-            return super().__getitem__(index)
         return np.asarray(self._node[_nexus_values])[index]
 
     def read_direct(self, buf, source_sel):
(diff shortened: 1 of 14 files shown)


src/scippneutron/file_loading/nxobject.py
```diff
@@ -6,11 +6,12 @@
...
 import functools
 from typing import List, Union, NoReturn, Any, Dict, Tuple
+import numpy as np
 import scipp as sc
 import h5py
 
-from ._nexus import LoadFromNexus
-from ._hdf5_nexus import LoadFromHdf5, _cset_to_encoding, _ensure_str
+from ._hdf5_nexus import _cset_to_encoding, _ensure_str
+from ._hdf5_nexus import _ensure_supported_int_type, _warn_latin1_decode
```
jl-wynen, line 14: Importing a lot of protected symbols. Are you planning to factor their use out of this file?

src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -287,38 +287,107 @@ def __contains__(self, name):
...
 
-class JSONDataset:
-    def __init__(self, node: dict, loader: LoadFromJson):
+class JSONNode:
+    def __init__(self, node: dict, loader: LoadFromJson, *, parent=None):
+        self._file = parent.file if parent is not None else self
+        self._parent = self if parent is None else parent
         self._node = node
         self._loader = loader
+        print(self.name, parent is None, self._file.name, self._parent.name)
```
jl-wynen, line 300:
```suggestion
```

tests/nexus_test.py
```diff
@@ -61,15 +70,15 @@ def builder_with_events_monitor_and_log():
 def test_nxobject_root(nexus_group: Tuple[Callable, LoadFromNexus]):
     resource, loader = nexus_group
```
jl-wynen, line 71: Looks like `loader` can be removed from all those tests

src/scippneutron/file_loading/nxobject.py
SimonHeybrock: Yes, this will all move.

tests/nexus_test.py
SimonHeybrock: Working on it in follow-up.

jl-wynen (approved):

PR #314, opened 2022-03-22
Refactor new JSON wrappers and remove old code

SimonHeybrock:
There is more cleanup to do, moving more code around, but this is the major step removing the old `LoadFromJson` and `LoadFromHdf5`.

### DIFF
--- a/src/scippneutron/file_loading/_json_nexus.py
+++ b/src/scippneutron/file_loading/_json_nexus.py
@@ -91,8 +91,10 @@ def contains_stream(group: Dict) -> bool:
     """
     Return True if the group contains a stream object
     """
+    if not isinstance(group, _JSONGroup):
+        return False
     try:
-        for child in group[_nexus_children]:
+        for child in group._node[_nexus_children]:
             try:
                 if child["type"] == _nexus_stream:
                     return True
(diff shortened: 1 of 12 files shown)


src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -55,6 +54,14 @@ def _get_attribute_value(element: Dict,
     raise MissingAttribute
 
 
+class _Node(dict):
```
jl-wynen, line 57: What is the reason for having this in addition to `JSONNode`?

tests/nexus_helpers.py
```diff
@@ -240,14 +240,31 @@ def add_soft_link(file_root: h5py.File, new_path: str, target_path: str):
...
+    children = obj["children"]
+    for child in children:
+        if child.get('name') == name:
+            return child
+    return None
+
+
+def _get_object_by_path(file_root, path):
+    if not path.startswith('/'):
+        return _get_child(file_root, path)
```
jl-wynen, line 253: Why does this case no split the path? Seems to me like it should use the same loop as the other case.

src/scippneutron/file_loading/_json_nexus.py
SimonHeybrock: Some of the old code around stream handling still depends on this. `JSONNode` only supports the part of the json tree that correspond to Nexus, and ignores everything else.

tests/nexus_helpers.py
SimonHeybrock: Simply because it was not required for the tests.

jl-wynen (approved):

PR #319, opened 2022-03-29
Refactor more Nexus tests

SimonHeybrock:
- Refactor (I think all) remaining tests of new Nexus API.
- Support writing scipp.DType.string
- Avoid timezone warning
- Support `NXobject.__setitem__`, equivalent to `h5py` behavior.
  - I was wondering if this might enable a convenient syntax for creating classes? I am not 100% happy with `group.create_class('mylog', NX_class.NXlog)`, but I cannot see how, right now?

Next step will be to remove remaining dependencies, and move new API into a submodule/subfolder, so it can be extracted into a package later.

### DIFF
--- a/src/scippneutron/file_loading/_json_nexus.py
+++ b/src/scippneutron/file_loading/_json_nexus.py
@@ -45,7 +45,8 @@ numpy_to_filewriter_type = {
     np.uint16: "uint16",
     np.uint32: "uint32",
     np.uint64: "uint64",
-    np.str_: "string"
+    np.str_: "string",
+    np.object_: "string"
 }
 
 
--- a/tests/nexus_test.py
+++ b/tests/nexus_test.py
@@ -1,15 +1,11 @@
-from .nexus_helpers import (
-    NexusBuilder,
-    EventData,
-    Log,
-    Monitor,
-)
+from .nexus_helpers import NexusBuilder
 from .test_load_nexus import UTF8_TEST_STRINGS
+import h5py
 import numpy as np
 import pytest
-from typing import Callable
 import scipp as sc
 from scippneutron import nexus
+from scippneutron.nexus import NX_class
 
 
 def open_nexus(builder: NexusBuilder):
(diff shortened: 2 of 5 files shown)


src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -45,7 +45,8 @@
     np.uint16: "uint16",
     np.uint32: "uint32",
     np.uint64: "uint64",
-    np.str_: "string"
+    np.str_: "string",
+    np.object_: "string"
```
jl-wynen, line 49: How does this work? Does it call `str` on every element and store the result?

tests/nexus_test.py
```diff
@@ -21,295 +17,239 @@ def open_json(builder: NexusBuilder):
...
+    name, entry = items[0]
+    assert name == 'entry'
+    entry.create_class('monitor', NX_class.NXmonitor)
+    entry.create_class('log', NX_class.NXlog)
+    for k, v in entry.items():
+        if k == 'log':
+            assert v.nx_class == nexus.NX_class.NXlog
+        else:
+            assert k == 'monitor'
+            assert v.nx_class == nexus.NX_class.NXmonitor
```
jl-wynen, line 48: Is there a way to avoid the loop + if? E.g. something like `set(entry.items()) == ...`?

src/scippneutron/file_loading/_json_nexus.py
SimonHeybrock: I probably only happens to work if the objects are actually strings. I suspect it is broken for other objects. I hope we can deal with that when we get to it.

tests/nexus_test.py
SimonHeybrock: Maybe? I always find this kind of iterables hard to test.

jl-wynen: Maybe
```.py
assert set(entry) == {'log', 'monitor'}
assert entry['log'].nx_class == nexus.NX_class.NXlog
assert entry['monitor'].nx_class == nexus.NX_class.NXmonitor
```
?

I'd like to keep with the advice of not using loops and conditionals in tests. It took me some detailed reading to figure out that this test actually requires there to be only two items...

SimonHeybrock: But that is not testing `items()`, is it?

jl-wynen: Ah, right!
```.py
assert {k: v.nx_class for k, v in entry.items()} == {'log': nexus.NX_class.NXlog, 'monitor': nexus.NX_class.NXmonitor}
```
Though that is still a loop.

Pick whichever you prefer!

jl-wynen (approved):

tests/nexus_test.py
SimonHeybrock: See https://github.com/scipp/scippnexus/pull/1

PR #320, opened 2022-03-29
Extract Nexus API to scippnexus

SimonHeybrock:
This creates a scippnexus "package". Note that it does not split the packing yet. I think we will do that when we split this off into its own repo.


src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -3,9 +3,9 @@
 # @author Matthew Jones
 from __future__ import annotations
 from typing import Tuple, Dict, List, Any, Union
-import numpy as np
-from ._common import Group, MissingAttribute
 from dataclasses import dataclass
+import numpy as np
+from scippnexus._common import Group, MissingAttribute
```
jl-wynen, line 8: Should move those into a public module.

src/scippnexus/nxevent_data.py
```diff
@@ -132,12 +131,6 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
         return sc.DataArray(data=binned, coords={'event_time_zero': event_time_zero})
 
     def _check_for_missing_fields(self):
-        if contains_stream(self._group):
-            # Do not warn about missing datasets if the group contains
-            # a stream, as this will provide the missing data
-            raise SkipSource("Data source is missing datasets"
-                             "but contains a stream source for the data")
```
jl-wynen, line 139: Do we no longer need this?

src/scippnexus/nxobject.py
```diff
@@ -72,8 +76,7 @@ def __contains__(self, name: str) -> bool:
     def __getitem__(self, name: str) -> Any:
         attr = self._attrs[name]
         # Is this check for string attributes sufficient? Is there a better way?
-        is_json = isinstance(self._attrs, JSONAttributeManager)
-        if isinstance(attr, (str, bytes)) and not is_json:
```
jl-wynen, line 76: What happens in the case of json?

src/scippnexus/nxobject.py
```diff
@@ -138,8 +141,7 @@ def attrs(self) -> Attrs:
     @property
     def dtype(self) -> str:
         dtype = self._dataset.dtype
-        if str(dtype).startswith('str') or (not isinstance(self._dataset, JSONDataset)
-                                            and h5py.check_string_dtype(dtype)):
+        if str(dtype).startswith('str') or h5py.check_string_dtype(dtype):
```
jl-wynen, line 144: Hard coded dependency on h5py. Can this be refactored to decouple it from the loader?

src/scippnexus/nxevent_data.py
SimonHeybrock: I think we do, but I do not understand the workaround, or how this works. I don't know how we can resolve this on our own though, probably need input from ECDC.

src/scippnexus/nxobject.py
SimonHeybrock: As far as I can tell (from the tests) it is always utf8 (and this is how I implemented the workaround), so nothing will be done.

src/scippnexus/nxobject.py
SimonHeybrock: Are you suggesting to re-implement `check_string_dtype`?

src/scippneutron/file_loading/_json_nexus.py
SimonHeybrock: Added `scippnexus.typing`.

src/scippnexus/nxobject.py
SimonHeybrock: What problem would we solve though?

PR #322, opened 2022-04-04
Remove scippnexus folder and use scippnexus package instead

SimonHeybrock:
After code has been extracted previously into folder, removal/replacement turns out to be trivial, as intended.


conda/meta.yaml
```diff
@@ -32,6 +32,7 @@ requirements:
     - ninja
     - python
     - scipp 0.13.*
+    - scippnexus
```
nvaytet, line 35: I think you need to add `scippnexus` to the environment files in the docs from the installation pages: `docs/environments/scippneutron.yml`.
There is also a `latest` version in there. I guess we will remove that once we stop making dev packages as in scipp.

SimonHeybrock: Why? Doesn't conda make this a dependency of scippneutron?

nvaytet: hmmm that's a good point. Forget what I said...

nvaytet (approved):

PR #326, opened 2022-04-13
Various fixes for recent scippnexus changes

SimonHeybrock:
Some behavior changes. Not 100% happy with everything, but since it mainly affects the cases in which the file contains (partially) invalid fields or attributes I am not too worried about this at this point.

Depends on scipp/scippnexus#20 and a release that will be made after that.

### DIFF
--- a/src/scippneutron/file_loading/_json_nexus.py
+++ b/src/scippneutron/file_loading/_json_nexus.py
@@ -234,6 +234,14 @@ class JSONAttributeManager:
         self._node['attributes'].append(attr)
         return self[name]
 
+    def __iter__(self):
+        if (attrs := self._node.get('attributes')) is not None:
+            if isinstance(attrs, dict):
+                yield from attrs
+            else:
+                for item in attrs:
+                    yield item[_nexus_name]
+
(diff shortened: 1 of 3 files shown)


src/scippneutron/file_loading/_json_nexus.py
```diff
@@ -234,6 +234,14 @@ def __setitem__(self, name, value):
         self._node['attributes'].append(attr)
         return self[name]
 
+    def __iter__(self):
+        if (attrs := self._node.get('attributes')) is not None:
+            if isinstance(attrs, dict):
+                yield from attrs
+            else:
+                for item in attrs:
+                    yield item[_nexus_name]
```
jl-wynen, line 243: What type is `attrs` in this case? I am confused why in the dict case, it yields the keys, but here it looks like it yields values.

SimonHeybrock: Indeed, it is confusing. In the JSON `item[_nexus_name]` is the *name* of the attribute.

jl-wynen: Ok

PR #358, opened 2022-07-27
Use provision-with-micromamba action

jl-wynen:
Changed to miniconda in #352 because mamba failed to download mantid. But miniconda is super slow. So trying the new version of mamba.

### DIFF
--- a/.buildconfig/ci-linux.yml
+++ b/.buildconfig/ci-linux.yml
@@ -34,5 +34,5 @@ dependencies:
   - requests
   - sphinx
   - sphinx-autodoc-typehints>=1.17.0
-  - sphinx-book-theme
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
Fine for CI, but are we sure we want to do this for our package releases? Is it 100% guaranteed that the packages will work for miniconda users?

jl-wynen:
> Fine for CI, but are we sure we want to do this for our package releases? Is it 100% guaranteed that the packages will work for miniconda users?

It only uses mamba for solving the environment. The packages are still built with conda-build. So there should be no difference. Except that mamba might succeed to solve during build but conda fails during installation. But AFAIK the dependency specs are identical. They depend on `meta.yaml` and not the actual build env.

PR #361, opened 2022-08-17
Refactor coord transform implementation to make kernels public

jl-wynen:
There should be no functional changes, code has only been moved around and documented. And there are new tests.

Except for
- graphs are in different submodules, so they need to be imported differently.
- `two-theta` uses a new implementation because the old one had a floating point error of ~1e-6 for small angles.
- graphs are now always copied. Previously, it was possible to modify global objects through the returned graphs in some cases.
- bugfix in `dspacing_from_wavelength`. `c` was constructed incorrectly.


docs/about/release-notes.rst
```diff
@@ -3,6 +3,26 @@
...
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+* `scippneutron.tof.conversions` has been split into `scippneutron.conversion.graph.beamline` and `scippneutron.conversion.graph.tof` `361 <https://github.com/scipp/scipp/pull/361>`_.
+
+Features
+~~~~~~~~
+
+* Kernels for cordinate transformations are now public in `scippneutron.conversion` `361 <https://github.com/scipp/scipp/pull/361>`_.
+
```
SimonHeybrock, line 18: You also mentioned 2 bugfixes in the PR?

docs/user-guide/coordinate-transformations.ipynb
```diff
@@ -349,9 +356,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from scippneutron.tof.conversions import beamline, elastic_Q\n",
+    "from scippneutron.conversion import graph\n",
     "\n",
-    "q_with_gravity = {**beamline(scatter=True), **elastic_Q(\"tof\")}\n",
+    "q_with_gravity = {**graph.beamline.beamline(scatter=True),\n",
```
SimonHeybrock, line 361: `graph.beamline.beamline` seems odd.Also, what kind of beamline? At a reactor a beamline is very different, e.g., there is no L1 or source positions.

src/scippneutron/_utils/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+Internal utilities; do not use outside scippneutron!
+"""
+
+from scipp.typing import VariableLike
+import scipp as sc
+
+
+def elem_unit(var: VariableLike) -> sc.Unit:
+    return var.bins.unit if var.bins is not None else var.unit
```
SimonHeybrock, line 13: I keep running into this... I wonder if there is a better solution (in scipp). Should the elem unit be exposed in the `unit` property of the parent? But that would not work for a dataset as a buffer.

src/scippneutron/conversion/beamline.py
```diff
@@ -0,0 +1,235 @@
...
+   :align: center
+
+The 'source position' defines the point where :math:`t=0` (or vice versa).
+Neutrons are scattered at the 'sample position' and detected
+at 'position' at :math:`t=\\mathsf{tof}`.
+
+In addition, ``total_straight_beam_length_no_scatter`` can be used if the beam
+does not scatter off of a sample.
+This is useful for instance for monitors.
+In this case, ``Ltotal`` is the distance from source to detector.
```
SimonHeybrock, line 23: All this stuff is TOF specific.

src/scippneutron/conversion/beamline.py
```diff
@@ -0,0 +1,235 @@
...
+
+In experiments with curved guides or additional reflections in the beam,
+it is often possible to use virtual positions in the input, such that a straight
+beamline with those positions is equivalent to a curved geometry with real positions.
```
SimonHeybrock, line 27: This is what is done in Mantid, due to the lack of better options. In scipp we might not want to recommend this, rather use a graph nodes that computes or provides the actual L1.

SimonHeybrock (commented):
Not a full review, some initial comments and questions:

src/scippneutron/_utils/__init__.py
jl-wynen: A simple solution would be a property that does what this function here does and raises for dataset buffers. But I think the actual unit of the variable needs to be accessible.

docs/user-guide/coordinate-transformations.ipynb
jl-wynen: `graph.beamline.tof` and `graph.beamline.reactor`?

SimonHeybrock (approved):

PR #365, opened 2022-08-26
Update logo

SimonHeybrock:


jl-wynen (changes requested):
Same as with scipp, the font is wrong on my machine.

jl-wynen (approved):

PR #378, opened 2022-11-03
Use plopp for instrument view if it is being used for other plots

nvaytet:


src/scippneutron/instrument_view.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 import scipp as sc
+from scipp.plotting.objects import PlotDict
```
SimonHeybrock, line 7: Two possible or future problems here:

- Does importing scippneutron without scipp's plotting dependencies still work?
- This will break once `scipp.plotting` is removed. Can we write it in a future-proof manner?

src/scippneutron/instrument_view.py
```diff
@@ -242,23 +243,30 @@ def instrument_view(scipp_obj=None,
...
+                      projection="3d",
+                      positions=positions,
+                      pixel_size=pixel_size,
+                      **kwargs)
+        scene = fig.view.figure.scene
 
     # Add additional components from the beamline
-    if components and not isinstance(plt, PlotDict):
-        scene = plt.view.figure.scene
+    if components and not isinstance(fig, PlotDict):
```
SimonHeybrock, line 269: Not sure I understand in which cases `fig` is *not* a `PlotDict`?

nvaytet: It's when you call instrument view on a dataset. To be honest, it's never used and I think we should just not support it, thus getting rid of the other problems in your first comment.

SimonHeybrock: 👍

tests/instrument_view_test.py
```diff
@@ -14,12 +14,6 @@ def test_neutron_instrument_view_3d():
     scn.instrument_view(d["a"])
 
 
-def test_neutron_instrument_view_with_dataset():
```
SimonHeybrock, line 17: So this is a breaking change? Add it to the release notes?

SimonHeybrock (approved):

PR #380, opened 2022-11-14
Fixes for scippnexus

SimonHeybrock:
Fixes for 0.4 (will need 0.4.1, see https://github.com/scipp/scippnexus/pull/76).

### DIFF
--- a/src/scippneutron/file_loading/load_nexus.py
+++ b/src/scippneutron/file_loading/load_nexus.py
@@ -168,7 +168,7 @@ def _zip_pixel_offset(da: sc.DataArray) -> sc.DataArray:
 
 
 def _by_nx_class(group) -> Dict[str, Dict[str, 'NXobject']]:
-    from scippnexus.nxobject import _nx_class_registry, _make
+    from scippnexus.nxobject import _nx_class_registry
     classes = {name: [] for name in _nx_class_registry()}
 
     def _match_nx_class(_, node):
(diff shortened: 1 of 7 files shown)


src/scippneutron/file_loading/load_nexus.py
```diff
@@ -186,7 +186,7 @@ def _match_nx_class(_, node):
         names = [group.name.split('/')[-1] for group in groups]
         if len(names) != len(set(names)):  # fall back to full path if duplicate
             names = [group.name for group in groups]
-        out[nx_class] = {n: _make(g) for n, g in zip(names, groups)}
+        out[nx_class] = {n: group._make(g) for n, g in zip(names, groups)}
```
jl-wynen, line 189: Is it really necessary to use private members? (`_make_`, `nx_class_registry`)

SimonHeybrock: Depends on whether `load_nexus` will survive. If we plan to support it long term then we should indeed to move to a public API.

jl-wynen: Ok. that is a discussion for another time.

SimonHeybrock:
Builds seem to be working now. Do we want to make a release after merging?

jl-wynen (approved):

jl-wynen:
> Do we want to make a release after merging?

Sure

PR #381, opened 2022-11-29
Try using `scipp.elemwise_func`

SimonHeybrock:
Considering some options for using `scipp.elemwise_func`. Please see comments in code.

### DIFF
--- a/src/scippneutron/conversion/tof.py
+++ b/src/scippneutron/conversion/tof.py
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
     c = _energy_constant(elem_unit(incident_energy), tof, L2)
     dtype = _common_dtype(incident_energy, tof)
     scale = (c * L2**2).astype(dtype, copy=False)
+
+    import numba
+    sig = numba.double(numba.double, numba.double, numba.double, numba.double)
+
+    # This is a way to avoid repeating the part that is needed for the unit computation,
+    # but unfortunately we need to manually use numba.cfunc for the nested call.
+    # I cannot think of a way to do this in sc.elemwise_func automatically.
+    @numba.cfunc(sig)
+    def dE(tof, t0, scale, incident_energy):
+        return incident_energy - scale / (tof - t0)**2
+
+    def dE_or_nan(tof, t0, scale, incident_energy):
+        delta_tof = tof - t0
+        return np.NAN if delta_tof <= 0.0 else dE(tof, t0, scale, incident_energy)
+
+    # Converting everything to float64...
+    # could make a different branch for float32?
+    func = sc.elemwise_func(dE_or_nan, unit_func=dE, auto_convert_dtypes=True)
+    return func(tof, t0, scale, incident_energy)
+
+    # Should we keep this to continue support if the user does not have numba?
+    # Or should numba become a mandatory dependency of scippneutron?
     delta_tof = tof - t0
     return sc.where(delta_tof <= sc.scalar(0, unit=elem_unit(delta_tof)),
                     sc.scalar(np.nan, dtype=dtype, unit=elem_unit(incident_energy)),


SimonHeybrock:
@jl-wynen Do you have any preferences on how to proceed?

src/scippneutron/conversion/tof.py
```diff
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
     c = _energy_constant(elem_unit(incident_energy), tof, L2)
     dtype = _common_dtype(incident_energy, tof)
     scale = (c * L2**2).astype(dtype, copy=False)
+
+    import numba
+    sig = numba.double(numba.double, numba.double, numba.double, numba.double)
+
+    # This is a way to avoid repeating the part that is needed for the unit computation,
+    # but unfortunately we need to manually use numba.cfunc for the nested call.
+    # I cannot think of a way to do this in sc.elemwise_func automatically.
```
jl-wynen, line 188: Why does `dE` need to be compiled at all? I though we just call the unit function as a Python function.

Also what repetition does this avoid?

src/scippneutron/conversion/tof.py
```diff
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
     c = _energy_constant(elem_unit(incident_energy), tof, L2)
     dtype = _common_dtype(incident_energy, tof)
     scale = (c * L2**2).astype(dtype, copy=False)
```
jl-wynen, line 181: Those unit conversions are not used in the new implementation. Do we need to pre or post process the inputs / output?

src/scippneutron/conversion/tof.py
```diff
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
...
+    sig = numba.double(numba.double, numba.double, numba.double, numba.double)
+
+    # This is a way to avoid repeating the part that is needed for the unit computation,
+    # but unfortunately we need to manually use numba.cfunc for the nested call.
+    # I cannot think of a way to do this in sc.elemwise_func automatically.
+    @numba.cfunc(sig)
+    def dE(tof, t0, scale, incident_energy):
+        return incident_energy - scale / (tof - t0)**2
+
+    def dE_or_nan(tof, t0, scale, incident_energy):
```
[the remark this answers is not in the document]
SimonHeybrock: Good point!

src/scippneutron/conversion/tof.py
```diff
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
...
+        delta_tof = tof - t0
+        return np.NAN if delta_tof <= 0.0 else dE(tof, t0, scale, incident_energy)
+
+    # Converting everything to float64...
+    # could make a different branch for float32?
```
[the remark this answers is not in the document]
SimonHeybrock: Yes, and branch here to compile two difference numba.cfuncs

src/scippneutron/conversion/tof.py
SimonHeybrock: Which conversion is not used?

src/scippneutron/conversion/tof.py
```diff
@@ -179,6 +179,28 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
...
+    func = sc.elemwise_func(dE_or_nan, unit_func=dE, auto_convert_dtypes=True)
+    return func(tof, t0, scale, incident_energy)
+
+    # Should we keep this to continue support if the user does not have numba?
+    # Or should numba become a mandatory dependency of scippneutron?
```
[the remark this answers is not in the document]
SimonHeybrock: One way would be to use `tox`, with two different environments (with and without numba)?

src/scippneutron/conversion/tof.py
jl-wynen: To `incident_energy.unit`.

src/scippneutron/conversion/tof.py
jl-wynen: Ok, didn't notice that. I tried to find a way to do this without using `cfunc` but didn't find any.

src/scippneutron/conversion/tof.py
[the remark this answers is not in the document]
jl-wynen: Is there a practical need for single precision floats in this case?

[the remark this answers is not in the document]
SimonHeybrock: Could be, inelastic data can be huge.

src/scippneutron/conversion/tof.py
SimonHeybrock: Where? What is not used?

jl-wynen:
Looks reasonable. Except, we maybe shouldn't use `else` with `try` and group all kernels in there but use something like `if numba_availale: ... else: ...`

SimonHeybrock:
> Looks reasonable. Except, we maybe shouldn't use `else` with `try` 

Why?

jl-wynen:
> Why?

- To group the implementation of the numba kernel with the outer python function. When there are multiple of each, it gets harder to read if the numba kernels are all at the top of the file and the others scattered throughout.
- Minor: To avoid `try-else` (I don't know the precise semantics)

SimonHeybrock:
> * Minor: To avoid `try-else` (I don't know the precise semantics)

I think it is actually recommended? It avoids putting more than necessary into the `try:`-block, thus avoiding catching more things than intended.

jl-wynen:
But I was suggesting
```python
try:
  import numba
except ImportError:
  numba = None

[...]

if numba is not None:
  def kernel():
    ...
else:
  kernel = None
```

PR #384, opened 2022-12-09
Remove C++ components

jl-wynen:
Please check that you can make a local installation!


.github/workflows/release.yml
```diff
@@ -71,36 +105,44 @@ jobs:
           name: html
           path: html/
 
-  upload_packages:
-    name: Deploy packages
-    needs: [docs]
+  upload_pypi:
```
SimonHeybrock, line 108: You are keeping the docs step before the package uploads?

docs/about/release-notes.rst
```diff
@@ -34,6 +34,31 @@ Release Notes
...
+
+Features
+~~~~~~~~
+
+* ScippNeutron is now available on PyPI `#384 <https://github.com/scipp/scipp/pull/384>`_.
+
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+* Removed C++ components and CMake package configuration. This only affects users that linked against the C++ part of ScippNeutron`#384 <https://github.com/scipp/scipp/pull/384>`_.
```
SimonHeybrock, line 48:
```suggestion
* Removed C++ components and CMake package configuration. This only affects users that linked against the C++ part of ScippNeutron `#384 <https://github.com/scipp/scipp/pull/384>`_.
```

pyproject.toml
```diff
@@ -1,5 +1,54 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+]
+requires-python = ">=3.8"
+dynamic = ["version"]
+dependencies = [
+    "h5py",
+    "numpy>=1.20",
+    "pooch",
```
SimonHeybrock, line 32: Wondering if we really need to require `pooch`?

tox.ini
```diff
@@ -1,17 +1,44 @@
...
-skip_install = true
+deps = -r requirements/test.txt
+setenv =
+  JUPYTER_PLATFORM_DIRS = 1
+commands = pytest
+
+[testenv:docs]
+description = invoke sphinx-build to build the HTML docs
+basepython = python3.8
+deps = {posargs:}
```
SimonHeybrock, line 14: No harm in keeping this, but the posargs where used to set the version in the docs workflow (to install via pip), which I think you are not using for ScippNeutron?

SimonHeybrock (commented):
Nice, must have been satisfying to delete so many things. :)

.github/workflows/release.yml
jl-wynen: This PR already changes a lot about the build process. I want to keep it as simple as possible and update other aspects separately.

pyproject.toml
jl-wynen: We can split off a few into an extra.

tox.ini
jl-wynen: But won't we do that in the future?

SimonHeybrock: See above, not sure what your plan is regarding the Mantid dependency.

.github/workflows/release.yml
jl-wynen: I am planning to change it. But concerning Mantid, we can't use tox unless we allow it to access external packages. At this point we might get into trouble with conflicting conda and pip packages.

SimonHeybrock (approved):

PR #385, opened 2022-12-13
Update front page and install instructions

SimonHeybrock:


docs/index.rst
```diff
@@ -9,8 +9,9 @@ News
 
 - [|SCIPPNEUTRON_RELEASE_MONTH|] scippneutron-|SCIPPNEUTRON_VERSION| `has been released <about/release-notes.rst>`_.
   Check out the `What's new <about/whats-new.rst>`_ notebook for an overview of recent highlights and major changes.
-- [January 2022] `scipp-0.11 <https://scipp.github.io/about/release-notes.html#v0-11-0-january-2022>`_ has been released.
-- Just like scipp, scippneutron has switched from GPLv3 to the more permissive BSD-3 license which fits better into the Python eco system.
+- [December 2022] ScippNeutron is now also available on PyPI.
+  See the updated ``pip`` `installation instructions <getting-started/installation.rst#Pip>`_.
+- Just like Scipp, ScippNeutron has switched from GPLv3 to the more permissive BSD-3 license which fits better into the Python eco system.
```
nvaytet, line 14: Unrelated to these changes but should there be a date on this item? So that people can easily see when the switch was made? Or alternatively, is it still needed? We switched quite a while ago, not sure it still qualifies as 'News'

docs/index.rst
```diff
@@ -9,8 +9,9 @@ News
...
 
 Where can I get help?
 ---------------------
```
nvaytet, line 17: I think the section below also needs updating.
There is no 'Question' label in the issues, and we should also point users to the Discussions, as we did in Scipp.

docs/getting-started/installation.rst
```diff
@@ -50,12 +50,6 @@ To create a new conda environment with scippneutron:
...
    $ conda create -n env_with_scipp -c conda-forge -c scipp -c ess-dmsc scippneutron
 
-For a more up-to-date version, the `scipp/label/dev` channel can be used instead:
-
-.. code-block:: sh
-
-   $ conda install -c conda-forge -c scipp/label/dev -c ess-dmsc scippneutron
-
 .. note::
    Instaling ``scippneutron`` on Windows requires ``Microsoft Visual Studio 2019 C++ Runtime`` (and versions above) installed.
```
jl-wynen, line 54: Is this still correct? What is the runtime needed for?

SimonHeybrock: I don't know, not sure we have a current Windows user.

SimonHeybrock:
I think the comments are unrelated, should we merge this so we can make a release?

jl-wynen (approved):
Go ahead

PR #387, opened 2022-12-19
Configure dependabot to update scipp

SimonHeybrock:
Copied from scippnexus.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,10 @@ updates:
       # Optional: Official actions have moving tags like v1;
       # if you use those, you don't need updates.
       - dependency-name: "actions/*"
+  - package-ecosystem: "pip"
+    directory: "/requirements"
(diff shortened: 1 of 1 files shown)


jl-wynen (approved):
Not sure I see the point. We typically know to update the scippneutron setup when scipp gets released. Dependabot would only do this when we take longer than a week to update.

SimonHeybrock:
> Not sure I see the point. We typically know to update the scippneutron setup when scipp gets released. Dependabot would only do this when we take longer than a week to update.

1. We will no longer need to do this manually, since scippneutron should typically be compatible with new releases.
2. We may forget, and not notice incompatibilities. As dependencies will no longer be pinned, as user installing scippneutron would thus get an incompatible scipp.
3. This is running daily, i.e., you find out by the next day if it is incompatible.
4. It is automating manual labor, I don't understand your complaint?

PR #391, opened 2022-12-21
Fix instrument_view when plopp is enabled

SimonHeybrock:


src/scippneutron/instrument_view.py
```diff
@@ -242,18 +242,14 @@ def instrument_view(scipp_obj,
...
-        pp = None
-
     positions_var = scipp_obj.meta[positions]
     if pixel_size is None:
         pos_array = positions_var.values
         if len(pos_array) > 1:
             pixel_size = np.linalg.norm(pos_array[1] - pos_array[0])
 
-    if (pp is not None) and (sc.plot is pp.plot):
+    if sc.config['plot'].get('use_plopp'):
```
nvaytet, line 251: Can you write a test so we make sure we are aware in the future if this breaks again?

SimonHeybrock: As we will remove this in the near future, I think it may not be worth the effort of coming up with a mechanism to write a config file, without breaking the developers local config?

jl-wynen (approved):

PR #400, opened 2023-01-24
Update test for more flexible scipp dtype support

SimonHeybrock:
Apparently the int32/int64 mismatch is fine for the current scipp/main. Updated test so the warning is still triggered.


jl-wynen:
I need more context. Why did you change this? What mismatch are you talking about?

SimonHeybrock:
Previously `sc.group` raised when groups were given as int32 but labels were int64.

jl-wynen (approved):

PR #407, opened 2023-01-31
Add streaming dependencies in buildconfig

jl-wynen:


SimonHeybrock (approved):
I wonder if we disabled this at some point because it was flaky in the past?

Issue #402, opened 2023-01-25
Implement frame unwrapping (and pulse skipping) into an actual utility


nvaytet:
Details on the algorithm to implement can be found [here](https://scipp.github.io/scippneutron/reference/frame-unwrapping.html).

- Ask the IDSs if they need tools to create time-distance diagrams for debugging (e.g. check that the frames look ok when doing pulse-skipping).

- We may also need to check for drift, analysing many pulses over a long period of time.

- Previous issue and more details in #236

SimonHeybrock:
Subtasks:

- [ ] Define required input parameter and check back with IDSs if those make sense.
- [ ] Implement transformation, see #236.
- [ ] With IDSs, figure out how those parameters should be obtained.
- [ ] Implement functionality to compute those parameters. This might involve, e.g., parsing and processing chopper parameters.

SimonHeybrock:
This is mostly addressed by #472 and #468.

PR #408, opened 2023-02-01
Frame unwrapping with pulse-skipping

### LINKED ISSUE #236: Transformations from `event_time_zero` and `event_time_offset` to `pulse_time` and `tof`

SimonHeybrock:
Part of #402.
Fixes #236.

@arm61 please have a look if this matches what you will need for Estia.

Apart from the input parameters found in the NeXus files directly (`event_time_zero` and `event_time_offset`), the frame unwrapping with pulse skipping requires:

- `pulse_period` (I suppose this is fixed at ESS)
- `pulse_stride` (configured when selecting pulse-skipping)
- `frame_offset` (could be obtain from some monitors, or based on opening of resolution choppers, if present)
- `lambda_min` (from chopper params, may be configured by user in experiment-control software?)
- `first_pulse_time` (I think this can be obtain from the phasing information of the choppers)

There will be follow-up work in which we can implement functionality that may be required to parse or compute these input parameters, based on additional meta data that should be available from the NeXus files.

Here is a PDF version of the modified notebook:

[frame-unwrapping.pdf](https://github.com/scipp/scippneutron/files/10557783/frame-unwrapping.pdf)

To do (after initial review):

- [x] More thorough testing.

### DIFF
--- a/docs/reference/frame-unwrapping.ipynb
+++ b/docs/reference/frame-unwrapping.ipynb
@@ -248,9 +248,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "wrapped = da.transform_coords(event_time_offset=lambda time: time % pulse_period)\n",
-    "wrapped = wrapped.transform_coords(\n",
-    "    event_time_zero=lambda time: pulse_period * (time // pulse_period)\n",
+    "wrapped = da.transform_coords(\n",
+    "    event_time_offset=lambda time: time % pulse_period,\n",
+    "    event_time_zero=lambda time: pulse_period * (time // pulse_period),\n",
     ")\n",
     "wrapped.hist(Ltotal=100, event_time_offset=100).plot()"
    ]
@@ -458,7 +458,7 @@
     "    raw,\n",
     "    pulse_period=pulse_period,\n",
     "    pulse_stride=2,\n",
-    "    lambda_min=0.8*wavmin,\n",
+    "    lambda_min=0.8 * wavmin,\n",
     "    frame_offset=0.0 * sc.Unit('ms'),\n",
     "    first_pulse_time=0 * pulse_period,\n",
     ").hist(Ltotal=100, tof=200).plot()"
--- a/docs/reference/frame-unwrapping.ipynb
+++ b/docs/reference/frame-unwrapping.ipynb
@@ -32,6 +32,7 @@
    "outputs": [],
    "source": [
     "from frameunwrapping import default_frame_diagram\n",
+    "\n",
     "default_frame_diagram().show()"
    ]
   },
@@ -87,6 +88,15 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "e9e31362-3464-4ccb-8499-7e7e2796ee0b",
+   "metadata": {},
+   "source": [
+    "`scippneutron.tof.frames` provides utilities for performing such computations.\n",
+    "The transformation graph looks as follows:"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
    "id": "66353e3c-ef2c-409f-a965-3478b035fdb3",
@@ -94,8 +104,18 @@
    "outputs": [],
    "source": [
     "from scippneutron.tof import frames\n",
+    "\n",
     "import scipp as sc\n",
+    "\n",
     "sc.show_graph(frames.to_tof(), simplified=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "994575cf-a896-43f9-947f-ba16cd007841",
+   "metadata": {},
+   "source": [
+    "Note how `time_offset` and `frame_period` are aliases for `event_time_offset` and `pulse_period`, respectively."
    ]
   },
   {
@@ -103,7 +123,10 @@
    "id": "4b532f66-9119-4e65-8827-1ed2f4dcfbd0",
    "metadata": {},
    "source": [
-    "## Pulse-skipping"
+    "## Pulse-skipping\n",
+    "\n",
+    "Choppers may be used to skip pulses, for the purpose of a simultaneous study of a wider wavelength range.\n",
+    "Conceptually this looks as follows:"
    ]
   },
   {
@@ -114,20 +137,17 @@
    "outputs": [],
    "source": [
     "from frameunwrapping import frame_skipping_diagram\n",
+    "\n",
     "frame_skipping_diagram().show()"
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "65f6f6ae-8025-4d53-b59f-34ef4f20dafe",
+   "cell_type": "markdown",
+   "id": "578444ae-edfa-4070-a00f-40afa2cdd92e",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "import scipp as sc\n",
-    "from scippneutron.tof import frames\n",
-    "graph = frames.to_tof()\n",
-    "sc.show_graph(graph)"
+    "The transformation graph that was given above for the non-pulse-skipping mode is then extended as shown below.\n",
+    "The difference is that `frame_period` and `time_offset` are now computed from additional input parameters:"
    ]
   },
   {
@@ -137,7 +157,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sc.show_graph(frames.to_tof(pulse_skipping=True), simplified=False)"
+    "sc.show_graph(frames.to_tof(pulse_skipping=True), simplified=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "02baefcd-75d6-4833-bd30-31b37cd1cfef",
+   "metadata": {},
+   "source": [
+    "For illustration, we create same fake event data:"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


docs/reference/frame-unwrapping.ipynb
```diff
@@ -86,6 +87,15 @@
     "Note that there are other valid definitions of $t_\\text{pivot}$, differing only in how neutrons in the intermediate \"background\" region (region between two pulses) are mapped to frames."
```
jl-wynen, line 87: Would it make sense to also compute an upper pivot time for the largest wavelength? Then we could mask the background region.

jl-wynen, line 87: Would it make sense to also compute an upper pivot time for the largest wavelength? Then we could mask the background region.

docs/reference/frame-unwrapping.ipynb
```diff
@@ -86,6 +87,15 @@
...
    ]
```
jl-wynen, line 88: I have got to ask again, are you sure that only two pulses can contribute to a frame? With WFM, there can be more, no? Or is this unrelated to this frame unwrapping code?

docs/reference/frame-unwrapping.ipynb
```diff
@@ -94,16 +104,29 @@
...
+  },
   {
    "cell_type": "markdown",
    "id": "4b532f66-9119-4e65-8827-1ed2f4dcfbd0",
    "metadata": {},
    "source": [
-    "## Pulse-skipping"
+    "## Pulse-skipping\n",
+    "\n",
+    "Choppers may be used to skip pulses, for the purpose of a simultaneous study of a wider wavelength range.\n",
```
jl-wynen, line 128:
```suggestion
    "Choppers may be used to skip pulses, for the purpose of a study of a wider wavelength range.\n",
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+   "source": [
+    "sc.show_graph(frames.to_tof(pulse_skipping=True), simplified=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "02baefcd-75d6-4833-bd30-31b37cd1cfef",
+   "metadata": {},
+   "source": [
+    "For illustration, we create same fake event data:"
```
jl-wynen, line 168:
```suggestion
    "For illustration, we create some fake event data:"
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+    "    table = sc.DataArray(sc.ones(dims=dims, shape=[N]), coords={'Ltotal': L, 'tof': t})\n",
+    "    table = table.transform_coords('wavelength', graph=graph.tof.elastic(\"tof\"))\n",
+    "    da = table.bin(wavelength=1).bins['wavelength', wavmin:wavmax]\n",
+    "    return da.transform_coords(time=lambda tof: tof + frame_number * frame_period)\n",
+    "\n",
+    "\n",
+    "da = sc.concat([fake_frame(i) for i in range(-2, 8)], 'dummy').bins.concat('dummy')\n",
+    "\n",
+    "da = da.bins['time', 0.0 * sc.Unit('ms') : 8 * frame_period]\n",
+    "del da.attrs['time']  # avoid problems below with non-monotonous coord transform"
```
jl-wynen, line 208:
```suggestion
    "del da.attrs['time']  # avoid problems below with non-monotonic coord transform"
```

I don't someone will understand the problem from this short sentence. Can you explain it a bit more concretely? Maybe in a footnote.

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ee2e08b2-6e25-49c7-a2d4-07cf90b04496",
+   "metadata": {},
+   "source": [
+    "Our fake data illustrates what we might observe when placing many neutron monitors every along our beamline.\n",
```
jl-wynen, line 216:
```suggestion
    "Our fake data illustrates what we might observe when placing many neutron monitors along our beamline.\n",
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "fc5892a1-fc0f-4108-8c7d-e995d345bc35",
+   "metadata": {},
+   "source": [
+    "We refer to each of the emitted bundles of neutrons as *frame*.\n",
+    "\n",
+    "In practice, NeXus files record event data using `NXevent_data`, and in particular the `event_time_zero` and `event_time_offset` fields.\n",
+    "`event_time_zero` gives the time of the source pulse.\n",
+    "When pulse-skipping is used, the neutrons from each *frame* may nevertheless assigned to *pulses* in the NeXus files.\n",
```
jl-wynen, line 239:
```suggestion
    "When pulse-skipping is used, the neutrons from each *frame* may nevertheless be assigned to *pulses* in the NeXus files.\n",
```

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+    "That is, while we have a frame-period of $2 \\cdot 71~\\text{ms}$ (in this example), `event_time_zero` is in steps of $71~\\text{ms}$.\n",
+    "The `event_time_offset` that we would obtain from a NeXus files thus records the modulus of the `time`:"
```
jl-wynen, line 241:
```suggestion
    "The `event_time_offset` that we would obtain from a NeXus file thus records the modulus of the `time`:"
```

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: Yes that might make sense, but I didn't consider it essential for now, especially since it can easily be done later.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: I think splitting things for WFM would be done after this. Why would you think there are more with WFM?

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: This is how people refer to it. You can also study a wider wavelength range by simply making to runs with different wavelength bands and then join the data.

docs/reference/frame-unwrapping.ipynb
SimonHeybrock: I don't think it is relevant?

docs/reference/frame-unwrapping.ipynb
```diff
@@ -114,17 +137,343 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "cab93a0a-88ea-41a1-8bf8-87e732395c8b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wrapped = da.transform_coords(event_time_offset=lambda time: time % pulse_period)\n",
+    "wrapped = wrapped.transform_coords(\n",
```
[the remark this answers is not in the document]
SimonHeybrock: Hmm, probably.

src/scippneutron/tof/frames.py
```diff
@@ -53,35 +92,109 @@ def to_tof():
...
+    """
+    Unwrap raw timestamps from ``NXevent_data`` into time-of-flight.
+
+    The input data must provide the following coordinates:
+
+    - ``event_time_offset``, as read from ``NXevent_data``
+    - ``event_time_zero``, as read from ``NXevent_data``
+      (only for ``pulse_stride > 1``)
+    - ``Ltotal`` or coordinates that allow for computation of thereof. By default
+      ``Ltotal`` may be defined including the full distance between source and sample.
```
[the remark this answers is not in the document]
SimonHeybrock: No, Ltotal also includes the sample-detector distance.

docs/reference/frame-unwrapping.ipynb
jl-wynen: Depends on how the frame period is defined. If it is the (proton) source period, then there are multiple neutron pulses in every frame.

src/scippneutron/tof/frames.py
```diff
@@ -4,37 +4,75 @@
...
+def _tof_from_wavelength(*, wavelength: sc.Variable,
+                         Ltotal: sc.Variable) -> sc.Variable:
     scale = (m_n / h).to(unit=sc.units.us / elem_unit(Ltotal) / elem_unit(wavelength))
     return as_float_type(Ltotal * scale, wavelength) * wavelength
 
 
-def _tof_to_time_offset(*, tof, frame_length, frame_offset):
-    unit = tof.unit
-    frame_length = frame_length.to(unit=unit)
+def _tof_to_time_offset(*, tof: sc.Variable, frame_period: sc.Variable,
```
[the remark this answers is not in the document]
SimonHeybrock: What was the reason for the reverse order?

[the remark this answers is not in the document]
jl-wynen: I don't think there was any. It looks like `x = f(y)`. And it makes it easier to say `x_from_y_and_z`. But I came up with those two right now.
Either order works for me, I am more concerned about a consistent order.

jl-wynen (approved):

src/scippneutron/tof/frames.py
```diff
@@ -53,35 +95,103 @@ def to_tof():
...
+        'pulse_period', 'pulse_stride', 'first_pulse_time', 'frame_offset', 'lambda_min'
+    ]
+    for x in coords:
+        if x in da.meta:
+            raise ValueError(f"Input data has '{x}' coord or attr, but values should "
+                             "be given only as a function parameter.")
+    da.attrs['pulse_period'] = pulse_period
+    da.attrs['frame_offset'] = frame_offset
+    da.attrs['lambda_min'] = lambda_min
+    if pulse_stride != 1:
```
nvaytet, line 192: Do we want to guard against non-integer values supplied for `pulse_stride`?

tests/tof/frames_test.py
```diff
@@ -137,98 +150,75 @@ def test_make_frames_reproduces_true_pulses(tof_min, frame_offset):
...
-                                                  time_zero=300.0 * sc.Unit('ms')) == 1
-        assert time_zero_to_detection_frame_index(**params,
-                                                  time_zero=301.0 * sc.Unit('ms')) == 1
-        assert time_zero_to_detection_frame_index(**params,
-                                                  time_zero=400.0 * sc.Unit('ms')) == 0
+    # Should reproduce reference 'tof' within rounding errors
+    expected = reference.bins.coords['tof']
+    assert sc.allclose(da.bins.coords['tof'],
+                       expected,
+                       atol=sc.scalar(1e-9, unit=expected.bins.unit),
```
nvaytet, line 223: Any particular reason why tolerance here is 1.0e-9 and it is 1.0e-12 in a test further up?

nvaytet (commented):
I think the 'bad cases' in the notebook is really interesting to have :+1:

tests/tof/frames_test.py
SimonHeybrock: I am not really sure, but the differences where larger here. I don't know if it is just the wider range of params I tested?

src/scippneutron/tof/frames.py
SimonHeybrock: Hmm, I think this highlights a limitation of `transform_coords`: While I could add a check here, someone could still use `to_tof` to get the underlying graph. I could add a check in every graph node that uses `pulse_stride`, but that is quite cumbersome.

So to really solve this, we would either:

- Support checking of dtypes in `transform_coords`.
- Support non-variable inputs in `transform_coords`. That is, the `pulse_stride` would not be passed via the input's coords, but as a direct param when calling `transform_coords`. This could then also do validation.

The second option seems worthwhile exploring, since currently we always have to add temporary dummy coords to inputs to use `transform_coords`, which is not very elegant. @jl-wynen Do you have any thoughts on this?

jl-wynen: Only that I recently ran into something similar in the reflectometry workflow.
I think this can be done quite easily.

But this introduces a number of edge cases. E.g. what if an explicit parameter has the same name as a coord. Or what if the parameter introduces a new dimension or has a shape mismatch with an existing dim.

SimonHeybrock: I suppose we need to define an order of priority, or always raise an exception.

Shape mismatches are not a new problem, in principle: Nodes could also introduce this, so I think it will/should just fail.

SimonHeybrock:
I will merge this as I have not gotten further feedback. It may require iteration in the future anyway, if it turns out to be lacking.

Issue #411, opened 2023-02-09
Transformations to `(Qx, Qy, Qz)` and `hkl`


SimonHeybrock:
- [x] We have a transformation to `Q`, the magnitude of $\vec Q$. For single-crystal diffraction a transform to $\vec Q$ will be required.
- [x] A similar case is conversion to `hkl`. This requires knowledge of the crystal orientation matrix. See for example https://docs.mantidproject.org/v6.1.0/concepts/Lattice.html.

jl-wynen:
Waiting for a release of Scipp with the inversions of transforms.

nvaytet:
To get the SANS `Qx`, `Qy`, would we just extract the components of the $Q$ vector?

jl-wynen:
The current implementation produces `Qx`, `Qy`, `Qz` separately. So you can just take those as individual coords. (Provided the maths is correct for sans(?))

nvaytet:
we might want to do something specific for SANS anyway, because they probably also want to take gravity into account.

SimonHeybrock:
I am not sure this is what you need for SANS. I think it is not even clear if the common SANS Qx/Qy formalism makes sense for LOKI without changes, since it has some quite large scattering angles. Should check with Wojciech.

PR #412, opened 2023-02-09
Coordinate transformation to Q vector

jl-wynen:
Part of #411

### DIFF
--- a/src/scippneutron/conversion/tof.py
+++ b/src/scippneutron/conversion/tof.py
@@ -301,7 +301,7 @@ def _wavelength_Q_conversions(x: VariableLike, two_theta: VariableLike) -> Varia
 
 def Q_from_wavelength(*, wavelength: VariableLike,
                       two_theta: VariableLike) -> VariableLike:
-    """Compute the momentum transfer from wavelength.
+    """Compute the absolute value of the momentum transfer from wavelength.
 
     The result is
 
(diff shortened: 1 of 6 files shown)


src/scippneutron/conversion/tof.py
```diff
@@ -361,6 +361,47 @@ def wavelength_from_Q(*, Q: VariableLike, two_theta: VariableLike) -> VariableLi
...
+
+    Returns
+    -------
+    :
+        The momentum transfer vector :math:`\\vec{Q}`.
+        Has the inverse unit of ``wavelength``.
+    """
+    e_i = incident_beam / sc.norm(incident_beam)
+    e_f = scattered_beam / sc.norm(scattered_beam)
+    return 2 * np.pi / wavelength * (e_i - e_f)
```
SimonHeybrock, line 402: I am not sure a `vector3` dtype is convenient here. Typically, the user will want to bin into the 3 Q components in a subsequent step, so returning `Qx`, `Qy`, and `Qz` may be better?

tests/conversion/tof_conversions_test.py
```diff
@@ -291,6 +312,35 @@ def test_wavelength_from_Q_single_precision(two_theta_dtype):
...
+    Q_vec = tof_conv.Q_vec_from_wavelength(wavelength=wavelength,
+                                           incident_beam=incident_beam,
+                                           scattered_beam=scattered_beam)
+    assert sc.identical(Q_vec['pos', 0],
+                        2 * np.pi / wavelength * sc.vector([-1.0, 0.0, 1.0]))
+    assert sc.identical(Q_vec['pos', 1],
+                        2 * np.pi / wavelength * sc.vector([0.0, -1.0, 1.0]))
+    assert sc.allclose(
+        Q_vec['pos', 2], 2 * np.pi / wavelength *
+        sc.vector([0.0, -0.4472135954999579, 0.1055728090000841]))
```
SimonHeybrock, line 341: As a sanity check, you could compute `Q` and `Q_vec`, and check if the norm is consistent?

src/scippneutron/conversion/tof.py
jl-wynen: After talking to Celine, I decided to switch to returning individual components and drop `Q_vec` for now as  it seems less important. But it is easy to add back in if it turns out that it is useful, e.g. for computing the hkl vector.

tests/conversion/tof_conversions_test.py
```diff
@@ -314,24 +315,49 @@ def test_wavelength_from_Q_single_precision(two_theta_dtype):
...
+@settings(**global_settings)
+def test_Q_elements_from_wavelength_consistent_with_Q(incident_beam, scattered_beam,
+                                                      wavelength):
+    wavelength = wavelength.rename({wavelength.dim: 'wavelength'})
+    (Qx, Qy, Qz) = tof_conv.Q_elements_from_wavelength(wavelength=wavelength,
+                                                       incident_beam=incident_beam,
+                                                       scattered_beam=scattered_beam)
+    Q_vec = sc.vectors(dims=Qx.dims,
+                       values=np.dstack((Qx.values, Qy.values, Qz.values)),
+                       unit=Qx.unit)
```
SimonHeybrock, line 340: Use https://scipp.github.io/generated/functions/scipp.geometry.position.html?

(I think we should rename this function though, I don't like the name).

src/scippneutron/conversion/tof.py
```diff
@@ -393,13 +399,18 @@ def Q_vec_from_wavelength(*, wavelength: VariableLike, incident_beam: VariableLi
...
+        y-component of the momentum transfer :math:`\\vec{Q}`.
+    Qz: scipp.VariableLike
+        z-component of the momentum transfer :math:`\\vec{Q}`.
     """
     e_i = incident_beam / sc.norm(incident_beam)
     e_f = scattered_beam / sc.norm(scattered_beam)
-    return 2 * np.pi / wavelength * (e_i - e_f)
+    e = e_i - e_f
+    k = 2 * np.pi / wavelength
+    return k * e.fields.x, k * e.fields.y, k * e.fields.z
```
SimonHeybrock, line 413: It is unlikely that this will ever become a performance bottleneck, but it would be interesting to think about `scipp::variable::transform` that would return multiple `Variable`.

tests/conversion/tof_conversions_test.py
jl-wynen: > (I think we should rename this function though, I don't like the name).

I guess this is why I missed it

jl-wynen: Ah no, this is not the reason. I missed it because I looked at `scipp.spatial`. Why do we have both modules? (And `geometry` is not documented as a separate module.)

SimonHeybrock: This is just legacy, from before when we had `spatial`. I think we could have something named `spatial.zip` and `spatial.unzip`, maybe? Or `spatial.to_vector`?

SimonHeybrock (approved):

PR #415, opened 2023-02-13
WFM

SimonHeybrock:
WFM for elastic scattering. Fixes scipp/ess#165.

The main addition for users is `scippneutron.tof.wfm.stitch_elastic`.


src/scippneutron/tof/wfm.py
```diff
@@ -0,0 +1,93 @@
...
+    -------
+    :
+        Data with corrected 'tof' coordinate. Furthermore, the 'incident_beam'
+        coordinate is set based on `subframe_source_position` such that subsequent
+        calculations of flight-path-length-dependent quantities will be correct
+        when using the standard conversion graphs.
+    """
+    da = da.copy(deep=False)
+    da.coords['subframe_source_position'] = subframe_source_position
+    graph = beamline.beamline(scatter=True)
```
nvaytet, line 82: Should `scatter` be a parameter? For example, in imaging you can have either scatter (if you are doing Bragg-edge stuff) or not (if you are looking at transmission)?

src/scippneutron/tof/wfm.py
```diff
@@ -0,0 +1,93 @@
...
+    binned.bins.coords[dim][dim, ::2] -= subframe_offset
+    del binned.coords[dim]
+    return binned[dim, ::2].bins.concat(dim)
+
+
+def stitch(da: sc.DataArray, wavelength_min: sc.Variable, wavelength_max: sc.Variable,
+           subframe_source_position: sc.Variable,
+           subframe_offset: sc.Variable) -> sc.DataArray:
+    """
+    Stitch WFM subframes of unstitched input data.
```
nvaytet, line 50: Should mention it only works for event data?

nvaytet (commented):
I think it looks good but you should try it on the WFM notebooks in `ess` to make sure we get the same results as the old code.

src/scippneutron/tof/wfm.py
SimonHeybrock: I gave it a try, but it just complicates the code. I think we should make a separate wrapper for that case. Instead, I renamed `stitch` -> `stitch_elastic` (and you may not that apart from no-scattering this implies the third case, inelastic scattering).

src/scippneutron/tof/wfm.py
SimonHeybrock: Done.

tests/tof/wfm_test.py
```diff
@@ -0,0 +1,182 @@
...
+                                  subframe_offset=subframe_offset)
+    # The stitched result is does not contain wavelength, but coords are setup to
+    # yield correct result.
+    stitched_wavelength = stitched.transform_coords('wavelength',
+                                                    graph=graph.tof.kinematic("tof"))
+    hist = stitched_wavelength.hist(wavelength=expected.coords['wavelength'])
+    assert sc.identical(expected.data, hist.data)
+
+
+def test_stitch_incorrect_given_bad_subframe_source_position():
```
nvaytet, line 99: Not sure I see so much value in this test and the ones below. It doesn't hurt to have them, but it seems kind of obvious that if you slightly change something in the input, it will yield different results.

Unless we check excatly how the results are changed, e.g. if events end up in the wrong frame, or if their wavelengths are shifted by an amount that we can predict.

I understand that checking for these exact numbers is tedious, and potentially not really worth it.
Looking at the old tests in the `ess` repo, we had a nice thing where we were checking the contract the WFM promises on the wavelength resolution the stiching yields.

However, you probably need all the chopper settings to be able to add such a test, thus making it outside the scope of the present PR. I was just thinking that tests along those lines are more valuable?

SimonHeybrock: > but it seems kind of obvious that if you slightly change something in the input, it will yield different results.

Exactly, this ensures that the implementation actually does anything. Given the complexity of the input and output data that is actually not a given. While implementing these tests, I had several cases where some attribute leftover from test data creation led to the stitching code taking a "shortcut" --- it simply used the attribute with the known result (e.g., `Ltotal`) instead of recomputing it from the parameters.

docs/about/release-notes.rst
```diff
@@ -41,6 +41,7 @@ Features
 ~~~~~~~~
 
 * Coordinate transformation kernel for :math:`\vec{Q}` `412 <https://github.com/scipp/scipp/pull/412>`_.
+* Added :py:func:`scippneutron.tof.wfm.stitch_elastic` for stitching event data from WFM beamlines with elastic scattering `415 <https://github.com/scipp/scipp/pull/415>`_.
```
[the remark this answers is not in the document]
SimonHeybrock:
```suggestion
* Added :py:func:`scippneutron.tof.wfm.stitch_elastic` for stitching event data from WFM beamlines with elastic scattering `#415 <https://github.com/scipp/scipp/pull/415>`_.
```

nvaytet (approved):

SimonHeybrock:
@nvaytet Despite having tested this against the example you have in the `ess` docs, I am not convinced this (or rather the wavelength->time-cut) code is correct. Doesn't this assume that the opening time of the second chopper is the closing time of the first chopper (compare figure from cell 8: https://scipp.github.io/ess/techniques/wfm/introduction-to-wfm.html#A-closer-look)?

SimonHeybrock:
How should we proceed with this? The chopper-related bits certainly need more iterations. Should we merge it anyway, and get back to this when we have more information?

SimonHeybrock:
Based on discussion:

- Look into how this fits together with [frame-unwrapping](https://scipp.github.io/scippneutron/reference/frame-unwrapping.html). In particular we need info about start of first subframe to be able to unwrap frames.
- Consider moving forward with this, under the condition that we make it more clear that it assumes optically blind choppers.
- Iterate based on that.

PR #417, opened 2023-02-15
XYE filewriter

### LINKED ISSUE #413: GSAS-II file writer
Need to find out which parts of the format are required. At the minimum, we need to save files that can be read by easyDiffraction.

jl-wynen:
Fixes #413

easyDiffraction can load files written by `save_xye`.


docs/about/release-notes.rst
```diff
@@ -41,6 +41,7 @@ Features
 ~~~~~~~~
 
 * Coordinate transformation kernel for :math:`\vec{Q}` `412 <https://github.com/scipp/scipp/pull/412>`_.
+* Added :func:`scippneutron.io.xye.save_xye` and :func:`scippneutron.io.xye.load_xye` `417 <https://github.com/scipp/scipp/pull/417>`_.
```
SimonHeybrock, line 44: The `xye` module feels like overkill? Will there be more than 2 user-facing functions?

src/scippneutron/__init__.py
```diff
@@ -14,6 +14,7 @@
 from .core import convert
 from .mantid import from_mantid, to_mantid, load, fit
 from .instrument_view import instrument_view
-from .file_loading.load_nexus import load_nexus, load_nexus_json
+from .io.nexus.load_nexus import load_nexus, load_nexus_json
+from .io.xye import save_xye
```
SimonHeybrock, line 18: If this is imported at the top-level, why do release notes refer to nested modules?

Do we want to make it top-level, use use something like `scippneutron.io.save_xyz`? If the latter, we should of course consider the same for the other `load` functions.

src/scippneutron/io/xye.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    if da.variances is None:
+        raise sc.VariancesError(
+            'Cannot save data to XYE file because it has no variances.')
+    if da.ndim != 1:
+        raise sc.DimensionError(
+            'Cannot save data to XYE file because it is not one-dimensional. '
+            f'It has dimensions {da.dims}')
+    if len(da.coords) == 0:
+        raise ValueError('Cannot save data to XYE file because it has no coordinates.')
```
SimonHeybrock, line 71: Have you considered raising when there are masks? Dropping them is almost always wrong, so it might avoid hours/days of debugging for users.

src/scippneutron/io/xye.py
```diff
@@ -0,0 +1,142 @@
...
+    if coord is None:
+        if len(da.coords) > 1:
+            raise ValueError(
+                'Cannot deduce which coordinate to save because the data has more '
+                f'than one: {list(da.coords.keys())}')
+        coord = next(iter(da.coords))
```
SimonHeybrock, line 77: This does not feel very convenient. Why not use `da.coords[da.dim]` by default?

src/scippneutron/io/xye.py
```diff
@@ -0,0 +1,142 @@
...
+    if da.coords.is_edges(coord):
+        raise sc.CoordError('Cannot save data with bin-edges to XYE file.'
+                            'Compute bin-centers before calling save_xye.')
```
SimonHeybrock, line 80: Link to `midpoints`?

src/scippneutron/io/xye.py
```diff
@@ -0,0 +1,142 @@
...
+        header = _generate_xye_header(da, coord)
+    np.savetxt(fname, to_save, delimiter=' ', header=header)
+
+
+def load_xye(fname: Union[str, Path, io.TextIOBase],
+             *,
+             dim: str = 'dim_0',
+             coord: Optional[str] = None,
+             unit: Optional[sc.Unit] = None,
+             coord_unit: Optional[sc.Unit] = None) -> sc.DataArray:
```
SimonHeybrock, line 92: `None`, or better `default_unit` (so you get `dimensionless`)?

src/scippneutron/io/xye.py
```diff
@@ -0,0 +1,142 @@
...
+
+    See Also
+    --------
+    scippneutron.io.xye.save_xye:
+        Function to write XYE files.
+    """
+    coord = dim if coord is None else coord
+    loaded = np.loadtxt(fname, delimiter=' ', unpack=True)
+    if loaded.ndim == 1:
+        loaded = loaded[:, np.newaxis]
```
SimonHeybrock, line 127: I presume the header is not standardized, so we can make no attempt to read units and dims from there?

docs/about/release-notes.rst
jl-wynen: Probably not. But there may be many functions for different formats. I think it is cleaner to keep them in separate modules.
Note that it is possible to call them with `scn.io.save_xye`. But Sphinx doesn't want to generate link to that, only to the actual definition of the function.

src/scippneutron/__init__.py
jl-wynen: Ah, right. I did not really mean to add it in the top level, only in `io`. But note my comment above on Sphinx.

src/scippneutron/io/xye.py
jl-wynen: I considered several approaches:

- Raise when there are masks, attrs, other coords and add a flag to ignore this check.
- Log a message about other metadata.
- Do nothing.

I went with the last because I was unsure how useful this check would be. But I agree that masks may cause problems. I can add a check for those.

And maybe a log message of the form 'saving data or unit {unit} with coordiante {coord} to XYE file {fname}'?

src/scippneutron/io/xye.py
jl-wynen: This seemed risky. Since the file doesn't really include metadata, it is hard to tell what it actually contains.

src/scippneutron/io/xye.py
jl-wynen: :+1:

src/scippneutron/io/xye.py
jl-wynen: I picked None to indicate that the unit is unknown. If we load as dimensionless, then the data can be used in arithmetic. But by using `None`, there is a higher chance that it blows up and users are forced to fix it.

src/scippneutron/io/xye.py
jl-wynen: Correct. Mantid writes a similar header but places the units in separate lines.

docs/about/release-notes.rst
SimonHeybrock: We have the same problem in Scipp and elsewhere. I think we have to fix that (maybe by not using `automodule`, but listing contents by hand?).

src/scippneutron/io/xye.py
SimonHeybrock: I think ignoring other coords and attrs is not harmful, but ignoring masks definitely is?

src/scippneutron/io/xye.py
SimonHeybrock: But the dimension-coord should be the most sensible one in 99% of cases?

src/scippneutron/io/xye.py
SimonHeybrock: Hmm, so maybe it should be *mandatory* arguments?

jl-wynen: Done

SimonHeybrock (approved):

PR #419, opened 2023-02-20
Update for next scippnexus

SimonHeybrock:
Test will fail until scippnexus is released and dependencies updated. Locally I got all tests to pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "numpy>=1.20",
     "pooch",
     "scipp>=23.01.1",
-    "scippnexus>=32.3.0",
+    "scippnexus>=23.3.0",
     "scipy>=1.7.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "numpy>=1.20",
     "pooch",
     "scipp>=23.01.1",
-    "scippnexus>=0.4.1",
+    "scippnexus>=32.3.0",
     "scipy>=1.7.0",
 ]
 
--- a/src/scippneutron/io/nexus/_json_nexus.py
+++ b/src/scippneutron/io/nexus/_json_nexus.py
@@ -331,6 +331,9 @@ class JSONGroup(JSONNode):
         children = self._node[_nexus_children]
         return [child[_nexus_name] for child in children if not contains_stream(child)]
 
+    def items(self) -> List[Tuple[str, JSONNode]]:
+        return [(key, self[key]) for key in self.keys()]
+
     def _as_group_or_dataset(self, item, parent):
         if item['type'] == _nexus_group:
             return JSONGroup(item, parent=parent)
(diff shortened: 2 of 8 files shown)


src/scippneutron/io/nexus/_json_nexus.py
```diff
@@ -331,6 +331,9 @@ def keys(self) -> List[str]:
         children = self._node[_nexus_children]
         return [child[_nexus_name] for child in children if not contains_stream(child)]
 
+    def items(self) -> List[Tuple[str, JSONNode]]:
+        return [(key, self[key]) for key in self.keys()]
```
jl-wynen, line 335: Should this rather be a generator? (I don't know how long these lists get...)

src/scippneutron/io/nexus/load_nexus.py
```diff
@@ -132,7 +132,10 @@ def _depends_on_to_position(obj) -> Union[None, sc.Variable]:
                       (str, sc.DataArray)) or transform.dtype == sc.DType.DataArray:
             return None  # cannot compute position if bad transform or time-dependent
         else:
-            return transform * _origin(transform.unit)
+            if transform.dtype in (sc.DType.rotation3, sc.DType.linear_transform3):
```
jl-wynen, line 135: Why `linear_transform3`? This can have a unit, no?

tests/load_nxdetector_test.py
```diff
@@ -1,18 +1,10 @@
 import scipp as sc
-import scippnexus as nexus
-
-import scippneutron as scn
-
-
-def test_can_load_nxdetector_from_bigfake():
-    with nexus.File(scn.data.bigfake()) as f:
-        da = f['entry/instrument/detector_1'][...]
-        assert da.sizes == {'dim_0': 300, 'dim_1': 300}
```
jl-wynen, line 10: Does bigfake no longer work?

SimonHeybrock: It contains plain HDF5 groups. I think we should just ignore the file because it is basically broken.

src/scippneutron/io/nexus/load_nexus.py
SimonHeybrock: I thought that is a rotation unit?

src/scippneutron/io/nexus/_json_nexus.py
SimonHeybrock: Maybe we can worry about that after profiling? More important might be to change that in scippnexus.

src/scippneutron/io/nexus/load_nexus.py
jl-wynen: What is a 'rotation unit'? Linear transforms can contain a scaling transform which can have an arbitrary unit.

tests/load_nxdetector_test.py
jl-wynen: ok

src/scippneutron/io/nexus/load_nexus.py
SimonHeybrock: Right, I got confused with the unit of angle-axis.

SimonHeybrock:
```suggestion
            if transform.dtype == sc.DType.rotation3:
```

pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "numpy>=1.20",
     "pooch",
     "scipp>=23.01.1",
-    "scippnexus>=0.4.1",
+    "scippnexus>=32.3.0",
```
SimonHeybrock, line 34:
```suggestion
    "scippnexus>=23.3.0",
```

jl-wynen (approved):

PR #421, opened 2023-03-08
Update to sphinx 6 and sphinx-book-theme 1.0

jl-wynen:
See https://github.com/scipp/scipp/pull/3073

This is different here because scippneutron uses a conda env to build the docs and we don't (yet) pin all versions there.

When this is merged, I will manually upload the locally built docs.


SimonHeybrock:
What happened to the banner color?

image

SimonHeybrock:
The font sizes (and fonts in general?) also look odd... especially for the size in the toolbar?

SimonHeybrock:
Hmm, the banner color was a javascript problem.

SimonHeybrock:
Icons are not aligned? 
image

SimonHeybrock:
The search on the left sidepane is gone. The new magnifying class in the toolbar does not seem to work for me (finds no results).

jl-wynen:
> The font sizes (and fonts in general?) also look odd... especially for the size in the toolbar?

This is the default now. (I believe in pydata)

> Icons are not aligned?

Saw that, too. It doesn't happen on all pages. The book-theme docs don't have this problem. The only difference I could find between their generated HTML and our is that they have fontawesome 6 and we have 5 for some reason. But switching that in the browser did not help.

> The search on the left sidepane is gone. The new magnifying class in the toolbar does not seem to work for me (finds no results).

Odd, I have the same issue. Don't know what could have caused that.
This new method is from pydata. In their theme, it is possible to get the old search bar back. (But it only acts as a button and opens the same search bar in the centre of the screen as the magnifying glass.) Again, I don't think this is customisable in book-theme.

jl-wynen:
I'll revert to the old version for now

PR #424, opened 2023-03-16
Set unit=None in conversion from mantid.MaskWorkspace

SimonHeybrock:
This code predates the support for `unit=None` in Scipp and was missed when updating. Code for masks of workspaces seems to be correct, only this special case of the mask being values was affected.


SimonHeybrock:
@jl-wynen Do you feel like making a patch release for this?

PR #41, opened 2021-03-24
Release 0.1

SimonHeybrock:
Docs and other settings updated.

PR #427, opened 2023-03-20
Compute hkl

jl-wynen:
Fixes #41

Is this a useful API?

- Contrary to Q, this returns a vector. We could provide a function to unpack if needed.
- The argument names are pretty arbitrary. Are the current ones fitting?
- Are U and B usually kept separate or together?

The performance measurements were done using https://gist.github.com/jl-wynen/5b656104ec7949270d2143c0c6e97ac6

### DIFF
--- a/src/scippneutron/conversion/tof.py
+++ b/src/scippneutron/conversion/tof.py
@@ -29,13 +29,13 @@ def _common_dtype(a, b):
 
 
 def wavelength_from_tof(*, tof: VariableLike, Ltotal: VariableLike) -> VariableLike:
-    """Compute the wavelength from time-of-flight.
+    r"""Compute the wavelength from time-of-flight.
 
     The result is the de Broglie wavelength
 
     .. math::
 
-        \\lambda = \\frac{h t}{m_n L_\\mathsf{total}}
+        \lambda = \frac{h t}{m_n L_\mathsf{total}}
 
     Where :math:`m_n` is the neutron mass and :math:`h` the Planck constant.
 
@@ -49,7 +49,7 @@ def wavelength_from_tof(*, tof: VariableLike, Ltotal: VariableLike) -> VariableL
     Returns
     -------
     :
-        Wavelength :math:`\\lambda`.
+        Wavelength :math:`\lambda`.
         Has unit ångström.
     """
     c = sc.to_unit(const.h / const.m_n,
@@ -60,13 +60,13 @@ def wavelength_from_tof(*, tof: VariableLike, Ltotal: VariableLike) -> VariableL
 
 def dspacing_from_tof(*, tof: VariableLike, Ltotal: VariableLike,
                       two_theta: VariableLike) -> VariableLike:
-    """Compute the d-spacing from time-of-flight.
+    r"""Compute the d-spacing from time-of-flight.
 
     The result is the inter-planar lattice spacing
 
     .. math::
 
-        d = \\frac{h t}{m_n L_\\mathsf{total}\\; 2 \\sin \\theta}
+        d = \frac{h t}{m_n L_\mathsf{total}\; 2 \sin \theta}
 
     Where :math:`m_n` is the neutron mass and :math:`h` the Planck constant.
 
@@ -77,7 +77,7 @@ def dspacing_from_tof(*, tof: VariableLike, Ltotal: VariableLike,
     Ltotal:
         Total beam length.
     two_theta:
-        Scattering angle :math:`2 \\theta`.
+        Scattering angle :math:`2 \theta`.
 
     Returns
     -------
@@ -103,13 +103,13 @@ def _energy_constant(energy_unit: sc.Unit, tof: VariableLike, length: VariableLi
 
 
 def energy_from_tof(*, tof: VariableLike, Ltotal: VariableLike) -> VariableLike:
-    """Compute the neutron energy from time-of-flight.
+    r"""Compute the neutron energy from time-of-flight.
 
     The result is
 
     .. math::
 
-        E = \\frac{m_n L_\\mathsf{total}^2}{2 t^2}
+        E = \frac{m_n L_\mathsf{total}^2}{2 t^2}
 
     Where :math:`m_n` is the neutron mass.
 
@@ -138,19 +138,19 @@ def _energy_transfer_t0(energy, tof, length):
 def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
                                     L2: VariableLike,
                                     incident_energy: VariableLike) -> VariableLike:
-    """Compute the energy transfer in direct inelastic scattering.
+    r"""Compute the energy transfer in direct inelastic scattering.
 
     The result is
 
     .. math::
 
-        \\Delta E = E_i - \\frac{m_n L_2^2}{2 {(t - t_0)}^2}
+        \Delta E = E_i - \frac{m_n L_2^2}{2 {(t - t_0)}^2}
 
     With
 
     .. math::
 
-        t_0 = \\sqrt{m_n L_1^2 / (2 E_i)}
+        t_0 = \sqrt{m_n L_1^2 / (2 E_i)}
 
     and :math:`m_n` the neutron mass.
 
@@ -170,7 +170,7 @@ def energy_transfer_direct_from_tof(*, tof: VariableLike, L1: VariableLike,
     Returns
     -------
     :
-        Energy transfer :math:`\\Delta E`.
+        Energy transfer :math:`\Delta E`.
         Has the same unit as incident_energy.
 
     See Also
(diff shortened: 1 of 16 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+from typing import Any, List
+
+import pytest
+import scipp as sc
+
+pytest.register_assert_rewrite('scipp.testing.assertions')
+
+
+def pytest_assertrepr_compare(op: str, left: Any, right: Any) -> List[str]:
```
SimonHeybrock, line 12: Is this something we should move into `scipp.testing`?

src/scippneutron/conversion/tof.py
```diff
@@ -480,3 +474,111 @@ def dspacing_from_energy(*, energy: VariableLike,
...
+    :
+        ``Qx``, ``Qy``, ``Qz`` combined into a single variable of dtype ``vector3``.
+    """
+    if Qx.sizes != Qy.sizes or Qx.sizes != Qz.sizes:
+        raise sc.DimensionError("Qx, Qy, Qz must have the same sizes. "
+                                f"Got {Qx.sizes=}, {Qy.sizes=}, {Qz.sizes=}.")
+    return sc.geometry.position(Qx, Qy, Qz)
+
+
+def hkl_vec_from_Q_vec(*, Q_vec: Variable, u_matrix: Variable, b_matrix: Variable,
```
SimonHeybrock, line 502: Usually I think people have the UB matrix. Maybe you can create another node to optionally take U and B as inputs?

src/scippneutron/conversion/tof.py
```diff
@@ -480,3 +474,111 @@ def dspacing_from_energy(*, energy: VariableLike,
...
+
+    Returns
+    -------
+    :
+        ``Qx``, ``Qy``, ``Qz`` combined into a single variable of dtype ``vector3``.
+    """
+    if Qx.sizes != Qy.sizes or Qx.sizes != Qz.sizes:
+        raise sc.DimensionError("Qx, Qy, Qz must have the same sizes. "
+                                f"Got {Qx.sizes=}, {Qy.sizes=}, {Qz.sizes=}.")
+    return sc.geometry.position(Qx, Qy, Qz)
```
SimonHeybrock, line 499: This shows a shortcoming of our API: We really just want to attach some generic function to the graph, but this is not really possible nicely.

src/scippneutron/conversion/tof.py
```diff
@@ -480,3 +474,111 @@ def dspacing_from_energy(*, energy: VariableLike,
...
+    # others in particularly bad cases and equal in the majority of cases.
+    #
+    # Concerning performance, R, U, B are scalar variables or short array variables
+    # in typical use cases while Q is typically a long, potentially multi-dim array.
+    # So implementation 3 will likely perform the best.
+    # Since we don't have performance measurements or requirements at this point,
+    # the most accurate implementation was chosen.
+    return (sc.spatial.inv(b_matrix) * (sc.spatial.inv(u_matrix) *
+                                        (sc.spatial.inv(sample_rotation) * Q_vec)) /
+            (2 * np.pi))
```
SimonHeybrock, line 584: I think just as for Q, people will want to make 3D HKL plots (or 2 cuts), so have elements would be more useful. We should either move ahead with scipp/scipp#3046, or consider supporting strides for input coords in `bin` (when we can use the fields of the vector for binning).

src/scippneutron/conversion/tof.py
```diff
@@ -480,3 +474,111 @@ def dspacing_from_energy(*, energy: VariableLike,
...
+    # to have the best overall accuracy, being an order of magnitude better than the
+    # others in particularly bad cases and equal in the majority of cases.
+    #
+    # Concerning performance, R, U, B are scalar variables or short array variables
+    # in typical use cases while Q is typically a long, potentially multi-dim array.
+    # So implementation 3 will likely perform the best.
+    # Since we don't have performance measurements or requirements at this point,
+    # the most accurate implementation was chosen.
+    return (sc.spatial.inv(b_matrix) * (sc.spatial.inv(u_matrix) *
+                                        (sc.spatial.inv(sample_rotation) * Q_vec)) /
```
SimonHeybrock, line 583: In practice Q will be massive, so I think we have to pick the implementation that operates on it only once. Since all the matrices are double-precision I doubt that accuracy will become an issue (compared to the uncertainties on the inputs such as U, B, and R), or did you have any data suggesting otherwise?

tests/conversion/tof_conversions_test.py
```diff
@@ -414,3 +414,85 @@ def test_dspacing_from_energy_single_precision(two_theta_dtype):
...
+    b = 0.3
+    c = 1.1
+    alpha = 0.7
+    beta = 4.6
+    gamma = 1.4
+
+    return sc.spatial.linear_transform(
+        value=[[a, b * np.cos(gamma), c * np.cos(beta)],
+               [0, b * np.sin(gamma), -c * np.sin(beta) * np.cos(alpha)], [0, 0, 1 /
+                                                                           c]])
```
SimonHeybrock, line 470: Did you switch scippneutron to black yet?

tests/conversion/tof_conversions_test.py
```diff
@@ -414,3 +414,85 @@ def test_dspacing_from_energy_single_precision(two_theta_dtype):
...
+    sample_rotation_coeffs = np.array([1.2, 0.8, -1.2, -0.1])
+    sample_rotation_matrix = sc.spatial.rotation(value=sample_rotation_coeffs /
+                                                 np.linalg.norm(sample_rotation_coeffs))
+
+    hkl_vec = tof_conv.hkl_vec_from_Q_vec(Q_vec=Q_vec,
+                                          u_matrix=u_matrix,
+                                          b_matrix=b_matrix,
+                                          sample_rotation=sample_rotation_matrix)
+
+    assert hkl_vec.unit == Q_vec.unit
```
SimonHeybrock, line 492: Hmm, I thought HKL had no units?

tests/conftest.py
jl-wynen: We can't. This needs to be in the project's `conftest.py`. Or in a pytest plugin. We could look into bundling a plugin with Scipp that gets installed automatically when you install scipp.

src/scippneutron/conversion/tof.py
jl-wynen: Do you mean a generic function that combines things into a vector?

If the workflow system works out well, we can maybe reimplement coord transforms on top of it and support more types of operations.

src/scippneutron/conversion/tof.py
jl-wynen: I only have the data that I linked in my original comment. For particularly large values in `B`, the relative error of the resulting hkl is `10^-5` in this implementation and `10^-4` in the others. (The others are typically 1 order of magnitude worse except for cases where the error is very small to begin with.)

But I don't know what are realistic values for `B`.

tests/conversion/tof_conversions_test.py
SimonHeybrock: There are some existing tests in `test/load_nexus_test.py`. We made sure the correct units are set (they are not in the file) when implementing this. So this is not a real answer...

src/scippneutron/conversion/tof.py
SimonHeybrock: What do large values in B mean? Does this happen in practice?

jl-wynen: I used eq 3 in to build B matrices. The error was bad when `b1` or `b2` ~1000
https://www.psi.ch/sites/default/files/import/sinq/zebra/PracticalsEN/[1967-Busing-Levy-3-4-circle-Acta22.pdf](https://www.psi.ch/sites/default/files/import/sinq/zebra/PracticalsEN/1967-Busing-Levy-3-4-circle-Acta22.pdf)

jl-wynen: But I have no idea what realistic values are.

SimonHeybrock: Maybe you can speak to @g5t about this?

src/scippneutron/conversion/tof.py
```diff
@@ -577,8 +602,7 @@ def hkl_vec_from_Q_vec(*, Q_vec: Variable, u_matrix: Variable, b_matrix: Variabl
...
     # So implementation 3 will likely perform the best.
-    # Since we don't have performance measurements or requirements at this point,
-    # the most accurate implementation was chosen.
-    return (sc.spatial.inv(b_matrix) * (sc.spatial.inv(u_matrix) *
-                                        (sc.spatial.inv(sample_rotation) * Q_vec)) /
-            (2 * np.pi))
+    #
+    # This function uses implementation 3 as the performance gain
+    # is expected to be significant over 1 and 2.
+    return (sc.spatial.inv(sample_rotation * ub_matrix) * Q_vec) / (2 * np.pi)
```
SimonHeybrock, line 610: Wasting a lot by dividing by a constant last. Instead, divide before the broadcast to Q_vec.

jl-wynen: Transforms don't support this. We would have to unpack them into their values, divide, and repack into variables.

src/scippneutron/conversion/tof.py
jl-wynen: According to him, `~1/ångström`. With these values, the errors are below `1e-9` for implementation 1 and 3 and below `1e-8` for implementation 2. So we should be fine.

SimonHeybrock (approved):

PR #434, opened 2023-04-13
Scipp==23.03.2, h5py>3.2, mantid==6.6

nvaytet:
- Moved to `scipp==23.03.2` (deprecation of `scipp.geometry` in 23.03.1)
- Moved to `mantid==6.6` which allows us to remove the pin `h5py<3.2`
- Pinned all visible packages in the `.buildconfig/*.yml` files
- Applied `black` on notebooks

### DIFF
--- a/.github/workflows/pr_and_main.yml
+++ b/.github/workflows/pr_and_main.yml
@@ -70,7 +70,7 @@ jobs:
         # Linkcheck can be flaky. Avoid randomly breaking PR builds by running only on `main`
         if: ${{ contains(matrix.variant.os, 'ubuntu') && matrix.python-version == '3.8' && github.ref == 'refs/heads/main' }}
 
-      - uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


.buildconfig/ci-linux.yml
```diff
@@ -12,17 +13,17 @@ dependencies:
   - ipykernel
   - ipympl
   - ipywidgets
-  - mantid=6.5.*
+  - mantid=6.6.*
   - matplotlib
   - ninja
-  - plopp >= 22.12.1
+  - plopp
```
SimonHeybrock, line 19: Is there no minimum required for building the docs?

nvaytet: Could be, I can't remember. Ideally, I would like to pin all the versions in that file, as we have in [ess](https://github.com/scipp/ess/blob/main/environment.yml). Should I do that?

PR #436, opened 2023-04-18
Rename load to load_with_mantid

### LINKED ISSUE #221: Names of file loaders
- Should `scn.load` be renamed to be explicit about using mantid?
- Inconsistent keyword args

jl-wynen:
Fixes  #221


src/scippneutron/mantid.py
```diff
@@ -952,6 +953,89 @@ def load(
...
+
+    Examples
+    --------
+    >>> from scippneutron import load_with_mantid
+    >>> d = sc.Dataset()
+    >>> d["sample"] = load(filename='PG3_4844_event.nxs',
+    ...                    load_pulse_times=False,
+    ...                    mantid_args={'BankName': 'bank184',
+    ...                                 'LoadMonitors': True})  # doctest: +SKIP
+    """
```
SimonHeybrock, line 1017: Wasn't there a Sphinx directive for the version sth. was deprecated in?

jl-wynen: Yes. Added it with a guess for the next version number because the placeholder mechanism is not implemented in ScippNeutron.

SimonHeybrock (approved):

PR #438, opened 2023-05-09
Add deprecation warning to `load_nexus`

### LINKED ISSUE #453: ScippNeutron release compatible with ScippNexus 23.08

SimonHeybrock:
Also fixes #453.


src/scippneutron/io/nexus/load_nexus.py
```diff
@@ -119,6 +120,10 @@ def load_nexus(
     Usage example:
       data = sc.neutron.load_nexus('PG3_4844_event.nxs')
     """
+    warn(
+        "`load_nexus` is deprecated, please switch to using ScippNexus.",
```
nvaytet, line 124: Do we want to add a planned removal date?

jl-wynen: I had the same comment. But apparently I forgot to submit it :thinking:. So :+1:

SimonHeybrock: According to a recent survey (on Slack) no one is using it. What is the earliest we can get rid of it?

jl-wynen: Is there any benefit to removing it early? We are not developing ScippNeutron at the moment, so this doesn't cause any overhead.
We can stick with the usual 6 months period. Or maybe shorten it to 3 if you are sure that it won't cause trouble.

jl-wynen: Alternatively, we could keep the old scippnexus as a v1 submodule. But I see your point.
So why deprecate load_nexus at all?

SimonHeybrock: For example such that we will see if it used by `ess` code, and we get a transition period, instead of a hard break? But if we are confident that we can remove all uses ahead of time then we can of course skip the deprecation.

SimonHeybrock:
Added preparations for a 23.08 release, compatible with latest ScippNexus.

jl-wynen (approved):

PR #443, opened 2023-07-11
Update scipp

jl-wynen:
I tried to only update scipp in the requirements but `pip-compile-multi` failed to resolve the different environments consistently.

The tests will fail because this does not yet include updates in the mantid wrappers. Those will come in a separate PR because they include some behaviour changes around data groups.


src/scippneutron/tof/frames.py
```diff
@@ -209,12 +209,12 @@ def unwrap_frames(
                 f"Input data has '{x}' coord or attr, but values should "
                 "be given only as a function parameter."
             )
-    da.attrs['pulse_period'] = pulse_period
-    da.attrs['frame_offset'] = frame_offset
-    da.attrs['lambda_min'] = lambda_min
+    da.coords['pulse_period'] = pulse_period
+    da.coords['frame_offset'] = frame_offset
+    da.coords['lambda_min'] = lambda_min
```
SimonHeybrock, line 214: So this includes not just changes for 23.07, but for the future Scipp release which will deprecate `attrs`?

jl-wynen: `transform_coords` no longer reads from `attrs`. So at least some of those need to be coords now.

SimonHeybrock: Right!

SimonHeybrock (approved):

PR #461, opened 2023-10-20
Update for next scipp

jl-wynen:
This won't build yet because it needs `deprecated_attrs` with has not been released.


src/scippneutron/instrument_view.py
```diff
@@ -249,13 +240,35 @@ def instrument_view(
...
+    :
+        The 3D plot object
     """
     if not p3:
         raise _pythreejs_import_error
 
     import plopp as pp
 
-    positions_var = scipp_obj.meta[positions]
+    try:
```
nvaytet, line 268: Is the `try/except` here for backward compatibility, for older versions of Scipp who do not have `deprecated_meta`? If so, does that mean we don't need to set the minimum scipp requirement to 23.08?

However, there are then places below that make use of `deprecated_attrs` that are not in a `try/except` block.
So they would need the version pin. We should either pin min scipp version 23.08 and use `deprecated_attrs`, or not pin and use `try/except`. Right now it seems there is a mix of both?

Unless I just completely misunderstood the motivation for the `try/except`?

jl-wynen: This is more about forward compatibility since it uses `coords` in the `except`. So when we fully remove `attrs`, this code will continue to work.

I was unsure if I should make all uses of attrs and meta backwards compatible. But I decided against it because it would introduce even more noise. So I'd rather go with a lower version limit.
Concerning forward compat, I can update `frames.py` to fall back to `coords`. But this makes no sense for `load_nexus` as that will be removed before attrs.

nvaytet: > I can update frames.py to fall back to coords

:+1:

src/scippneutron/tof/frames.py
```diff
@@ -88,6 +88,14 @@ def frame_period_from_pulse_period(
     return pulse_period * pulse_stride
 
 
+def _meta_or_fallback(x):
+    # This can be removed when attrs have been removed in the minimum scipp version.
+    try:
+        return x.deprectaed_meta
```
nvaytet, line 94:
```suggestion
        return x.deprecated_meta
```

jl-wynen: Should've run the tests locally.
Fixed this now. But found that the mantid and streamaing tests didn't run. Working on it...

jl-wynen:
The tutorials need some work. I may have to rewrite a significant part of 'exploring data' to use data groups.

jl-wynen:
Now also fixes #462

jl-wynen:
I changed it to always fall back it `deprecated_(attrs|meta)` is not available. This should not work with scipp 23.07 and 23.08+

Also, I added a CI matrix entry to test with an up to date scipp version.

nvaytet:
> This should not work with scipp 23.07 and 23.08+

Did you mean "not work" or "now work"?

nvaytet (approved):

jl-wynen:
> > This should not work with scipp 23.07 and 23.08+
> 
> Did you mean "not work" or "now work"?

I meant 'now'

PR #464, opened 2023-10-24
Use py 3.10 and mantid 6.8 in CI

jl-wynen:


SimonHeybrock:
What about 3.9? Does/should ScippNeutron still support it?

jl-wynen:
> What about 3.9? Does/should ScippNeutron still support it?

I'd say it should. And it does. But if we want to test it, then we need to test both with and without mantid because we have no way of running mantid tests with 3.9.

SimonHeybrock:
> > What about 3.9? Does/should ScippNeutron still support it?
> 
> I'd say it should. And it does. But if we want to test it, then we need to test both with and without mantid because we have no way of running mantid tests with 3.9.

I fear that 1 minute after we drop 3.9 testing someone will merge something that breaks support? If we plan to drop 3.9 soon anyway then maybe we should simply make a clear decision?

jl-wynen:
In scipp, we should stick to the numpy schedule, i.e., support 3.9 until next April. But downstream, we maybe can be more flexible. I don't know if we have any dependencies in ess that don't support it yet, though.

SimonHeybrock:
Yes, definitely for Scipp (and any non-ESS packages (I am counting ScippNeutron as ESS-package)).

Is it much work to have CI run for 3.8 without Mantid? Are we fine with "hoping" that nothing breaks 3.8 in the next 6 months?

jl-wynen:
It should run with both versions now. Pytest should automatically skip mantid tests when mantid is not installed.
The main downside I see is that we have theses long running / hanging tests. So doubling the number of runs increases the risk of that happening.

jl-wynen:
Actually, it's more complicated because there seems to be no easy way to filter out mantid from the buildconfig files. I can duplicate them for 3.9. This would not be a huge issue as this will be temporary. What do you think?

SimonHeybrock:
I don't have strong opinions here, go with whatever you prefer.

jl-wynen:
It's implemented now. So let's keep it

.github/workflows/pr_and_main.yml
```diff
@@ -37,10 +37,12 @@ jobs:
...
-        - {os: ubuntu-20.04, build-config: ci-linux}
-        - {os: macos-11, build-config: ci-macos}
-        - {os: windows-2019, build-config: ci-windows}
-        python-version: ["3.8"]
+        - {os: ubuntu-20.04, python-version: "3.10", build-config: ci-linux}
+        - {os: macos-11, python-version: "3.10", build-config: ci-macos}
+        - {os: windows-2019, python-version: "3.10", build-config: ci-windows}
+        - {os: ubuntu-20.04, python-version: "3.9", build-config: ci-linux-39}
+        - {os: macos-11, python-version: "3.9", build-config: ci-macos-39}
+        - {os: windows-2019, python-version: "3.9", build-config: ci-windows-39}
```
SimonHeybrock, line 45: I you worry about longer build times, I suppose keeping 3.9 just for a single `os` would be sufficient, since it is highly unlikely we will add platform-specific code that works on all 3.10 but only some 3.9 platforms.

SimonHeybrock (approved):

.github/workflows/pr_and_main.yml
jl-wynen: True.

PR #466, opened 2023-10-31
Chopper toolkit

jl-wynen:
Fixes https://github.com/scipp/ess/issues/166

Waiting for review of the maths by ECDC. But the code is ready except for the TODO in `nexus_chopper.py`.


src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    slits: Optional[int] = None
+    """Number of slits."""
+    slit_angle: Optional[sc.Variable] = None
+    """Angular opening of slits."""
+    slit_edges: Optional[sc.Variable] = None
+    """Edges of the slits as angles measured anticlockwise from top-dead-center.
+
+    On init, a 1d array of the form ``[begin_0, end_0, begin_1, end_1, ...]``
+    with ``begin_i < end_i``.
+    The order of slits is arbitrary.
```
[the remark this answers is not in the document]
jl-wynen: In what way? Apply the transformation in this class directly in scippnexus? This would be fine, I think.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+            raise RuntimeError("No top dead center has been defined")
+        res = (
+            relative_time.to(unit=self.top_dead_center.unit, dtype=int, copy=False)
+            + self.top_dead_center
+        )
+        if self.delay is not None:
+            res += self.delay.to(unit=res.unit, copy=False)
+        return res
+
+    def to_svg(self, image_size: int = 400) -> str:
```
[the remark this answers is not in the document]
jl-wynen: Yes. I have something similar in Scitacean, so I should be able to copy a lot of the code.

src/scippneutron/chopper/nexus_chopper.py
```diff
@@ -0,0 +1,62 @@
...
+import dataclasses
+from typing import Optional, Union
+
+import scipp as sc
+import scippnexus as snx
+
+from .disk_chopper import DiskChopper, DiskChopperType
+
+
+class NXDiskChopper(snx.NXdisk_chopper):
```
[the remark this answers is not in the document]
jl-wynen: What do you suggest?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+      all required information.
+
+    Slit openings
+    -------------
+
+    The terminology here differentiates slit 'begin' and 'end' from 'open' and 'close'.
+    The former refer to the angles relative to TDC as shown in the image above.
+    The latter refer to the opening and closing times of the slit.
+
+    It is possible to have ``end > 360 deg`` (clockwise) or ``begin < 0 deg``
```
nvaytet, line 93: I thought that George said the angles specified did not depend on the direction of rotation? (unless I misunderstood and 'clockwise' and 'anticlockwise' here are not refering to the direction of rotation?)

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    See https://manual.nexusformat.org/classes/base_classes/NXdisk_chopper.html
+    """
+    rotation_speed: Union[sc.Variable, sc.DataArray]
+    """Rotation frequency of the chopper."""
+    beam_position: Optional[sc.Variable] = None
+    """Angle where the beam crosses the chopper."""
+    delay: Optional[Union[sc.Variable, sc.DataArray]] = None
+    """Difference between global facility time and chopper time."""
+    pair_separation: Optional[sc.Variable] = None
+    """Disk spacing in direction of beam."""
```
nvaytet, line 134:
```suggestion
    """Disk spacing in direction of beam (for double choppers only)."""
```

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    phase: Optional[sc.Variable] = None
+    """Phase of the chopper rotation relative to the source pulses."""
+    radius: Optional[sc.Variable] = None
+    """Radius of the chopper."""
+    ratio: Optional[sc.Variable] = None
+    """Pulse reduction factor in relation to other choppers/fastest pulse."""
+    slits: Optional[int] = None
+    """Number of slits."""
+    slit_angle: Optional[sc.Variable] = None
```
nvaytet, line 143: Not sure I understand what the `slit_angle` is... Is it the angular size of the slit windows (i.e. slit edges end - begin?)
Am I right that the `slit_edges` contain both the begin and end angles. Is the `slit_angle` information then redundant?

I know some people (e.g. the V20 folk) prefer to define the choppers in term of a list of slit centers, with a list of slit angular widths.
But I don't see an option for the slit centers in the nexus standard.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    with ``begin_i < end_i``.
+    The order of slits is arbitrary.
+
+    After init, a 2d array of shape ``[slit, edge]`` where ``slit`` indexes the chopper
+    slits and ``edge`` is of length 2 where ``edge=0`` is the beginning edge and
+    ``edge=1`` the ending edge of the slit.
+    The dim names depend on the input.
+    """
+    slit_height: Optional[sc.Variable] = None
+    """Radial length of the slits."""
```
nvaytet, line 157: Can we be more specific here, e.g. saying it is the distance from the chopper outer edge to the bottom of the slit, or the distance from the chopper center to the bottom of the slit (I think it's the first one? correct me if I'm wrong)

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    """
+
+    position: sc.Variable
+    """Position of the chopper.
+
+    This is the center point of the chopper's axle in the face towards the source.
+    See https://manual.nexusformat.org/classes/base_classes/NXdisk_chopper.html
+    """
+    rotation_speed: Union[sc.Variable, sc.DataArray]
+    """Rotation frequency of the chopper."""
```
nvaytet, line 128: Should we add that this can be a single scalar or a 1d flat array of values?

Also, can the rotation speed be a derived quantity in some cases? i.e. derived from the TDC timestamps?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+
+    @property
+    def angular_frequency(self) -> sc.Variable:
+        """Rotation speed as an angular frequency in ``rad * rotation_speed.unit``."""
+        return sc.scalar(2.0, unit="rad") * sc.constants.pi * self.rotation_speed
+
+    def relative_time_open(self) -> sc.Variable:
+        """Return the opening times of the chopper slits.
+
+        The times are offsets of when the slit ``begin_edge`` passes by the
```
nvaytet, line 212: Not really understanding this sentence, is there a typo?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+        DiskChopper.relative_time_angle_at_beam:
+            Computes times relative to the chopper's top-dead-center.
+        """
+        return self._relative_to_absolute_time(self.relative_time_angle_at_beam(angle))
+
+    def _relative_to_absolute_time(self, relative_time: sc.Variable) -> sc.Variable:
+        if self.top_dead_center is None:
+            raise RuntimeError("No top dead center has been defined")
+        res = (
+            relative_time.to(unit=self.top_dead_center.unit, dtype=int, copy=False)
```
nvaytet, line 401: Why do we have to convert to integers here? Is it because the TDC are datetimes?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,471 @@
...
+    folded = edges.fold(edges.dim, sizes={edges.dim: -1, edge_dim: 2})
+    if sc.any(folded[edge_dim, 0] > folded[edge_dim, 1]):
+        raise ValueError(
+            "Invalid slit edges, must be given as "
+            "[begin_0, end_0, begin_1, end_1, ...] where begin_n < end_n"
+        )
+    return folded
+
+
+def _require_frequency(name: str, x: sc.Variable) -> None:
```
nvaytet, line 477:
```suggestion
def _require_frequency_unit(name: str, x: sc.Variable) -> None:
```
?

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
jl-wynen: Added, see screenshot.

Concerning the imports of protected symbols, see https://github.com/scipp/scipp/issues/3305.
The try-except blocks around imports are used because the html repr was moved in a recent scipp version.

chopper

tests/chopper/disk_chopper_test.py
```diff
@@ -0,0 +1,581 @@
...
+        position=sc.vector([0, 0, 0], unit='m'),
+        rotation_speed=sc.scalar(-7.21, unit='Hz'),
+        slit_edges=sc.array(dims=['slit'], values=[90.0, 180.0], unit='deg'),
+        phase=phase,
+    )
+    factor = (
+        sc.constants.pi.value
+        / 180  # to rad
+        / (2 * sc.constants.pi.value)  # to angular frequency
+        / abs(-7.21)
```
nvaytet, line 278: I can't exactly remember from the Google book when it was ok to make a helper function for tests, but maybe this code is a candidate?

tests/chopper/disk_chopper_test.py
```diff
@@ -0,0 +1,581 @@
...
+        sc.constants.pi.value
+        / 180  # to rad
+        / (2 * sc.constants.pi.value)  # to angular frequency
+        / abs(9.31)
+    )  # to time based on rotation speed
+    assert sc.identical(
+        ch.time_open(),
+        sc.datetimes(
+            dims=['slit'],
+            values=[int(30 * factor) + 642 + 4, int(170 * factor) + 642 + 4],
```
nvaytet, line 592: So we're losing precision if the units of the tdc and delay differ?

tests/externalfile.py
```diff
@@ -0,0 +1,40 @@
...
+    Uses the same setup in order to share files."""
+    import pooch
+
+    return pooch.create(
+        path=pooch.os_cache('scippnexus-externalfile'),
+        env='SCIPPNEXUS_DATA_DIR',
+        retry_if_failed=3,
+        base_url='login.esss.dk:/mnt/groupdata/scipp/testdata/scippnexus/',
+        registry={
+            '2023/BIFROST_873855_00000015.hdf': 'md5:eb180b09d265c308e81c4a4885662bbd',
```
nvaytet, line 19: Can this file be put on the public http server with the other files?

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: Looks great!

src/scippneutron/chopper/nexus_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: Anything wrong with `scippneutron.NXdisk_chopper`? Then at least the spelling is not confusing.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: :+1: The first one is my understanding, too.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: > Should we add that this can be a single scalar or a 1d flat array of values?

Can do. But I'm unsure which fields this applies to. The speed is relatively obvious. But what about phase, delay, beam_position? Also, some fields are annotated to allow data arrays while other aren't. This is based on the BIFROST file I saw. But potentially all can be NXlogs and thus data arrays.

> Also, can the rotation speed be a derived quantity in some cases? i.e. derived from the TDC timestamps?

I guess it could be. I didn't want to add this kind of extra conversion on top of nexus. I'd prefer it if the user would calculate missing quantities,

src/scippneutron/chopper/disk_chopper.py
jl-wynen: Does it make a difference?

tests/chopper/disk_chopper_test.py
jl-wynen: I guess it's up to interpretation. It would look something like this:
```python
def deg_angle_to_time_factor(rotation_speed: float) -> float:
    # Multiply by the returned value to convert an angle in degrees
    # to the time when the point at the angle reaches tdc.
    to_rad = sc.constants.pi.value / 180
    angular_frequency = abs(rotation_speed) * (2 * sc.constants.pi.value)
    return to_rad / angular_frequency
```

tests/chopper/disk_chopper_test.py
jl-wynen: Yes. We could alternatively raise if the units are different or convert to the more precise one (which can also be lossy since those are ints).

tests/externalfile.py
jl-wynen: Maybe. I didn't want to duplicate it. It is outdated and needs to be replaced in the tests here anyway.

src/scippneutron/chopper/disk_chopper.py
nvaytet: I guess for now, yes?

src/scippneutron/chopper/disk_chopper.py
nvaytet: I asked on the nexus slack channel, no reply so far

src/scippneutron/chopper/disk_chopper.py
nvaytet: I think my point was: do we want to allow creating a chopper with TDC timestamps and no rotation speed? (right now it is a required argument)

tests/chopper/disk_chopper_test.py
nvaytet: I think if you need to change something in the future, then you only have to change one place if you add the function?

tests/chopper/disk_chopper_test.py
nvaytet: Raising maybe sounds safer? Unless it happens very often and becomes annoying?

jl-wynen:
Waiting for https://github.com/scipp/scippnexus/pull/173 now to fully implement the nexus loader.

src/scippneutron/chopper/nexus_chopper.py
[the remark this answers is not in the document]
jl-wynen: Doesn't match the PascalCase naming scheme of classes.

src/scippneutron/chopper/nexus_chopper.py
```diff
@@ -0,0 +1,78 @@
...
+    _SPECIAL_FIELDS = {
+        'typ',
+        'position',
+        'rotation_speed',
+        'top_dead_center',
+        '_clockwise',
+    }
+
+    def assemble(self, dg: sc.DataGroup) -> DiskChopper:
+        dg = snx.compute_positions(dg)
```
SimonHeybrock, line 39: This is going to fail in many cases. Paths may lead outside the chopper group. Not sure how we can solve this.

jl-wynen: Sounds like `compute_positions` needs to run before `assemble`. Otherwise, anything that assembles into custom types won't work in such cases.

SimonHeybrock: Would it be possible to create `DiskChopper` from already assembled chopper that ScippNexus gives you, instead of implementing a custom definition?

jl-wynen: That would have to be a post-processing step after loading the file. And it would not be able to automatically convert choppers because the output data groups don't encode nexus classes. (We should probably change that and store the NXclass as an item in the output.)

SimonHeybrock: This is what we did in `scippnexus.v1`. The problem is that this leads to eager loading and parsing of depends-on-chains, which had a lot of overhead. A caching mechanism could help, but it is not trivial.

On Slack we recently discussed storing the `NX_class` in the loaded `DataGroup`. Would that help?

jl-wynen: > On Slack we recently discussed storing the NX_class in the loaded DataGroup. Would that help?

I'd say yes. This is what I said in my last post, too.

tests/chopper/disk_chopper_test.py
nvaytet: Or maybe: what happens if one of the angles is 0.5 degrees? Or even 12.9 degrees?

SimonHeybrock: Why would you raise?

src/scippneutron/chopper/nexus_chopper.py
SimonHeybrock: By the way: This is broken for current DREAM files, an probably many others as well. Can we do this without using custom definitions for now?

SimonHeybrock: This also raises if info is incomplete, i.e., this is not useful for loading files in general. For example, I get an exception because of a missing `value` (again: for the current DREAM files). We either need to make the parser more relaxed, or do this after loading?

tests/chopper/disk_chopper_test.py
jl-wynen: TO avoid loss of precision

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,481 @@
...
+        See Also
+        --------
+        DiskChopper.time_open:
+            Computes the absolute opening time in the global timing system.
+        """
+        if self.slit_edges is None:
+            raise RuntimeError("No slits have been defined")
+        if self._clockwise:
+            return self.relative_time_angle_at_beam(self.slit_begin)
+        return self.relative_time_angle_at_beam(self.slit_end)
```
SimonHeybrock, line 233: Not sure I get this: For a chopper running at a multiple of the source frequency, e.g., 28 Hz at ESS, I would expect to get to "openings" for slit. But it seems I do not, unless I messed up.

src/scippneutron/chopper/disk_chopper.py
nvaytet: @g5t thinks the `slit_angle` is kind of old/deprecated. That it most probably represents the width of the slits, to be used in combination with `slits` (the number of slits), which would describe a number of slits evenly spaced around the chopper disk and that all have the same opening width.

Using `slit_edges` is more versatile, this making `slit_angle` mostly obsolete for ESS.

jl-wynen: So what are you suggesting? I added them as a way of representing the info in a nexus file.

nvaytet: I don't know what I'm suggesting. You can leave them, but maybe we should have a comment in the docstring of all the members that we do not use, saying something like "Currently unused by scippneutron"?

src/scippneutron/chopper/nexus_chopper.py
jl-wynen: So what do I do now? Is there a need to have a function that can assemble a chopper or can we do this manually for now? E.g.
```python
with snx.File('...') as f:
    entry = f[()]
entry = compute_positions(entry)
entry['instrument']['chopper'] = make_chopper_from_nexus(entry['instrument']['chopper'])
```
Note that we cannot detect choppers in a loaded data group until we save the nexus class to the data group. So we need a manual approach like above for now.

jl-wynen:
The core datastructure and functions are done. What's missing now are helpers for identifying when the chopper is in phase and filtering data based on it. I have implementations but didn't put them into a publishable state yet.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+        """
+        angle = self._apply_angle_repetitions(angle=angle, n_repetitions=n_repetitions)
+        angle = (
+            self.beam_position.to(unit='rad')
+            + self.phase.to(unit='rad')
+            - angle.to(unit='rad', copy=False)
+        )
+        if not self.is_clockwise:
+            angle = sc.scalar(2.0, unit='rad') * sc.constants.pi + angle
+        return angle / self.angular_frequency
```
SimonHeybrock, line 405: Does this guarantee that this always results in the smallest positive time?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+        Returns
+        -------
+        :
+            Variable of opening durations.
+        """
+        return abs(
+            self.time_offset_open(pulse_frequency=pulse_frequency)
+            - self.time_offset_close(pulse_frequency=pulse_frequency)
+        )
```
SimonHeybrock, line 423: Why using `abs`? Aren't the times define to be positive so you can just use subtraction?

SimonHeybrock (commented):
Part 1 (more later today):

pyproject.toml
```diff
@@ -52,6 +52,9 @@ all = ['scipp[all]']
 
 [tool.pytest.ini_options]
 minversion = "7.0"
+# The custom `externalfile` marker runs tests that require passwordless
+# SSH access to login.esss.dk. We do not run it by default. Make sure to
+# setup keys in your SSH agent before running `pytest -m externalfile`.
```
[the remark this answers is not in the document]
jl-wynen: A leftover. I removed the nexus loader because of our earlier discussion. Removing this comment.

src/scippneutron/chopper/nexus_chopper.py
```diff
@@ -0,0 +1,85 @@
...
+        return None
+    if edges.ndim == 1:
+        edge_dim = 'edge' if edges.dim != 'edge' else '__edge_dim'
+        folded = edges.fold(edges.dim, sizes={edges.dim: -1, edge_dim: 2})
+        if sc.any(folded[edge_dim, 0] > folded[edge_dim, 1]):
+            raise ValueError(
+                "Invalid slit edges, must be given as "
+                "[begin_0, end_0, begin_1, end_1, ...] where begin_n < end_n"
+            )
+        return folded
```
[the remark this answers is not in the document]
jl-wynen: Well, the order of slits is technically arbitrary as far as I'm aware. So I didn't want to put a check for it. And yes, this can lead to overlapping slits. But I wasn't too concerned with validating the input completely.

src/scippneutron/chopper/nexus_chopper.py
```diff
@@ -0,0 +1,85 @@
...
+    if edges.ndim == 2:
+        if edges.shape[1] != 2:
+            raise sc.DimensionError(
+                "The second dim of the slit edges must be length 2."
+            )
+        return edges
```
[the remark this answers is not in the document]
jl-wynen: What ordering do you mean?

tests/chopper/nexus_chopper_test.py
```diff
@@ -0,0 +1,89 @@
...
+        unit='deg',
+    )
+    with pytest.raises(sc.DimensionError):
+        post_process_disk_chopper(raw_nexus_chopper)
+
+
+def test_2d_slit_edges_must_have_length_2(raw_nexus_chopper):
+    raw_nexus_chopper['slit_edges'] = sc.array(
+        dims=['slit', 'edge'],
+        values=[[0, 60, 90], [124, 126, 270]],
```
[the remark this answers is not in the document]
jl-wynen: Yes, that is the purpose of this test.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+        Its position serves as a reference point for defining angles.
+
+        The ``top_dead_center`` field of a NeXus chopper stores timestamps of the
+        TDC sensor registering a full rotation.
+        This serves as a reference time for the chopper :math:`t_0`.
+        In :class:`DiskChopper`, the TDC is encoded as a component of
+        :attr:`DiskChopper.phase`.
+    * - ``beam_position``
+      - :math:`\tilde{\theta}`
+      - The angle under which the beam hits the chopper
```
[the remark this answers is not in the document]
jl-wynen: TDC. See the TDC entry in the table

src/scippneutron/chopper/nexus_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: I think I'd just check they are sorted?

src/scippneutron/chopper/nexus_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: In the 1-d case you check that begin<end, here you do not?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+      - :math:`T_0`
+      - Timestamp of a neutron pulse in global facility time.
+
+.. image:: /_static/chopper/chopper-coordinates.svg
+   :width: 400
+   :align: center
+
+:class:`DiskChopper` expects the chopper to be in phase with the source.
+It thus requires a constant rotation speed.
+And that speed must be an integer multiple of the source frequency or vice versa.
```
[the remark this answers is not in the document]
jl-wynen: I see this as an arbitrary rule that nobody actually follows. But fine, I merged the sentences.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    slit_height: Optional[sc.Variable] = None
+    """Distance from chopper outer edge to bottom of slits."""
+    radius: Optional[sc.Variable] = None
+    """Radius of the chopper."""
+
+    def __post_init__(self) -> None:
+        # Check for frequency because not all NeXus files store a unit
+        # and the name can be confusing.
+        _require_frequency('rotation_speed', self.rotation_speed)
```
SimonHeybrock, line 266: No validation of other assumptions, such as slit_edges as 2d array, and probably many more?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+            phase=_get_1d_variable(dg, 'phase'),
+            slit_edges=dg['slit_edges'],
+            slit_height=dg.get('slit_height'),
+            radius=dg.get('radius'),
+        )
+
+    @property
+    def slit_begin(self) -> sc.Variable:
+        """Beginning edges of the slits."""
+        return self.slit_edges[self.slit_edges.dims[1], 0]
```
SimonHeybrock, line 290: Looks like an anti-pattern. I think we have two options:
- Check the dims of `slit_edges`, to ensure `dims[1]` is what we want.
- Use actual dim names.

... but option 1) also means we can just use dims names.

As written now this is just as bad as working with plain NumPy arrays.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+        self, *, angle: sc.Variable, n_repetitions: int
+    ) -> sc.Variable:
+        dim = str(uuid4())
+        if n_repetitions == 1:
+            repetition_offsets = sc.scalar(0, unit=angle.unit)
+        else:
+            # -1 in the end value to exclude the end value from the array.
+            repetition_offsets = sc.arange(
+                dim, 0.0, n_repetitions * 2 * np.pi - 1, 2 * np.pi, unit='rad'
+            )
```
SimonHeybrock, line 468: Would it be clearer to make an integer arange and scale that by $\pi$? Might avoid the non-obvious way of excluding the end?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+        if n_repetitions == 1:
+            # Do not add a new dimension to the output.
+            return repeated
+        if angle.ndim == 0:
+            return repeated.flatten(to='slit')
+        # Remove aux dimension.
+        return repeated.transpose([*angle.dims[:-1], dim, angle.dims[-1]]).flatten(
+            dims=[dim, angle.dims[-1]], to=angle.dims[-1]
+        )
```
SimonHeybrock, line 483: Is all this needed because we store open and close in a single array?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+def _get_1d_variable(
+    dg: Mapping[str, Optional[sc.Variable, sc.DataArray]], name: str
+) -> sc.Variable:
+    if (val := dg.get(name)) is None:
+        raise ValueError(f"Chopper field '{name}' is missing")
+
+    msg = (
+        "Chopper field '{name}' must be a scalar variable, {got}. " "Use "
+    )  # TODO insert use
```
SimonHeybrock, line 547: TODO?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+.. math::
+
+    t_g(\theta) &= t_0 + \delta t + \begin{cases}
+    \frac{\theta-\tilde{\theta}}{|\omega|}, & \textsf{clockwise}\\
+    \frac{2\pi - (\theta-\tilde{\theta})}{|\omega|}, & \textsf{anticlockwise}
+    \end{cases}\\
+    &= t_0 + \delta t + - \frac{\theta - \tilde{\theta}}{\omega} + \begin{cases}
+    0, & \textsf{clockwise}\\
+    \frac{2\pi}{\omega}, & \textsf{anticlockwise}
+    \end{cases}
```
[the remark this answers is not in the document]
jl-wynen: It does. But that doesn't mean that it looks right on all systems. Where did you try it?

maths

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+The phase should be computed as defined about from the difference of a pulse time
+and a corresponding TDC timestamp.
+The user is responsible for determining the correct times.
+
+Slit openings
+-------------
+
+The terminology here differentiates slit 'begin' and 'end' from 'open' and 'close'.
+The former refer to the angles relative to TDC as shown in the image above.
+The latter refer to the times when a slit opens and closes for the beam.
```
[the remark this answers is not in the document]
jl-wynen: Don't know. Depends on how this works. If its enough to, e.g., define gaussian centred on the edge, we can do it downstream.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+    the rotation frequency be constant.
+
+    See Also
+    --------
+    scippneutron.chopper.disk_chopper:
+        For detailed documentation of the definitions and calculations used
+        by ``DiskChopper``.
+    """
+
+    position: sc.Variable
```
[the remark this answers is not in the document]
jl-wynen: I wanted to stick to the names in NeXus. If this is not a requirement, we can also rename other fields. E.g., `rotation_speed` is confusing because it's a frequency, not a speed.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+    def __post_init__(self) -> None:
+        # Check for frequency because not all NeXus files store a unit
+        # and the name can be confusing.
+        _require_frequency('rotation_speed', self.rotation_speed)
+
+    @classmethod
+    def from_nexus(
+        cls, dg: Mapping[str, Optional[sc.Variable, sc.DataArray]]
+    ) -> DiskChopper:
```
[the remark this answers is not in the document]
jl-wynen: > Missing docstring?

Yes.

This was intended to load a pre-processed loaded chopper. It is in general not possible to load directly from nexus because we need to filter out out-of-phase times.
Do you know a better name for the function?

src/scippneutron/chopper/disk_chopper.py
jl-wynen: No. See the picture below. (Part of the docs but I forgot to add it before). In the anticlockwise case, the short, blue slit could be placed in the first pulse.

Is this an issue? Or do you anyway need to replicate the openings a few times in order to cover all possible wavelengths?

disk-chopper-openings-dark

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: It looks fine on Mac (downloaded). Just the 2 MathJax web renders didn't like it.

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: Not sure NeXus is a good guide, so I suggest you choose what is best/clearest?

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: I think the name is fine, if it is clear/check what the input has to be?

src/scippneutron/chopper/disk_chopper.py
jl-wynen: True. I got the order in the subtraction wrong and fixed it with the `abs`.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+.. image:: /_static/chopper/disk-chopper-openings.svg
+   :class: only-light
+   :width: 700
+   :align: center
+
+.. image:: /_static/chopper/disk-chopper-openings-dark.svg
+   :class: only-dark
+   :width: 700
+   :align: center
```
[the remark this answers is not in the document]
jl-wynen: Forgot to add them. Fixed

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+        A positive frequency means anticlockwise rotation and a negative frequency
+        clockwise rotation (as seen from the source).
+    * - ``angular_frequency``
+      - :math:`\omega`
+      - :math:`\omega = 2 \pi f`, :attr:`DiskChopper.angular_frequency`.
+    * - ``delay``
+      - :math:`\delta t`
+      - Delay of the chopper timing system relative to global facility time with
+        :math:`t_g = t + \delta t`, where :math:`t_g` is a global time and :math:`t`
+        a chopper time.
```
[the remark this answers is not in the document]
jl-wynen: It effectively is because the delay is subsumed in the phase. I added it here to explain how the parameters map onto nexus. In particular, how the phase needs to be computed.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+Given the definitions above, the time in the global timing system when a point
+at angle :math:`\theta` is at the beam position is
+
+.. math::
+
+    t_g(\theta) &= t_0 + \delta t + \begin{cases}
+    \frac{\theta-\tilde{\theta}}{|\omega|}, & \textsf{clockwise}\\
+    \frac{2\pi - (\theta-\tilde{\theta})}{|\omega|}, & \textsf{anticlockwise}
+    \end{cases}\\
+    &= t_0 + \delta t + - \frac{\theta - \tilde{\theta}}{\omega} + \begin{cases}
```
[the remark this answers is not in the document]
jl-wynen: Fixed, should be `-`.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+.. math::
+
+    \Delta t_g(\theta) = t_g(\theta) - T_0 = - \frac{\theta - \tilde{\theta}
+       - \phi}{\omega}
+       + \begin{cases}
+         0, & \textsf{clockwise}\\
+         \frac{2\pi}{\omega}, & \textsf{anticlockwise}
+         \end{cases}
+
+where :math:`\phi = \omega (t_0 + \delta t - T_0)` is the ``phase``.
```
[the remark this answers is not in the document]
jl-wynen: Why? With $t_0 > T_0$, we have $sign(\phi) = sign(\omega)$.
The reason for this choice was to get $\theta = \omega t + \phi + ...$ which can be nicely plugged into $x = r\cos(\theta) = r\cos(\omega t + \phi + ...)$ which is a bit nicer than with $-\phi$. But of course, this is completely arbitrary.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+
+:meth:`DiskChopper.time_offset_angle_at_beam` can calculate :math:`\Delta t_g(\theta)`
+and :meth:`DiskChopper.time_offset_open` :meth:`DiskChopper.time_offset_close` calculate
+:math:`\Delta t_g` for slit open and close times.
+
+The definitions used here can lead to surprising results, especially when
+:math:`\tilde{\theta} \neq 0` or :math:`\phi \neq 0`.
```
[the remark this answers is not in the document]
jl-wynen: True. I wrote it because the theta=phi=0 case is easiest to think about. But I can reformulate.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+The plots below show the computed times for an angle :math:`\theta` for
+:math:`\tilde{\theta} \neq 0` and :math:`\phi = 0` (blue lines).
```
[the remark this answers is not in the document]
jl-wynen: Same deal as nonzero $\tilde{\theta}$. So would you find it easier to read if I replaced all $\tilder{\theta}$ with $\phi$?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,562 @@
...
+Note in particular the time ranges for :math:`\theta \in [0, 2\pi)` (gray rectangles).
+The other lines show :math:`\Delta t_g \,\mathsf{mod}\, 1/f` which is an option
+for restricting the times onto :math:`[0, \frac{2\pi}{|\omega|}) = [0, \frac1{f})`.
+
+.. image:: /_static/chopper/disk-chopper-time-curve.svg
```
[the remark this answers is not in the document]
jl-wynen: It is, but I found this more intuitive than writing $-\omega$.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: Some validation are done in `post_process_disk_chopper`. Do we need to repeat them here? I felt like this code already does a lot more validation than other code we write.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: But do we want to force certain dim names? That seems like just as much of an antipattern. So we would have to add arguments for this. At least in the constructor and then save the name.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: Done

src/scippneutron/chopper/disk_chopper.py
jl-wynen: Not really. The function is typically called with only open or close angles, i.e., 1d arrays. So this could be simplified by `angle.dims[-1]` -> `angle.dim`. But I initially wrote this such that it can also take `slit_edges`, or a similarly laid out array as input.

src/scippneutron/chopper/nexus_chopper.py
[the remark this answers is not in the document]
jl-wynen: Is this guaranteed by ECDC? I haven't heard that yet. And NeXus is silent on it.

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -0,0 +1,610 @@
...
+Time dependent parameters
+-------------------------
+
+In NeXus files, many chopper parameters are time-dependent, for example,
+``top_dead_center`` is an array of timestamps, or ``rotation_speed`` typically
+is an ``NXlog`` with speed measurements for different times.
+However, for simplicity and efficiency, :class:`DiskChopper` requires
+time-independent quantities.
+
+TODO:
```
jl-wynen, line 194: I need to add the tools I made for deteriming when the chopper is in phase for this section. But I never committed them and they are on my workstation which can't access until next year...

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: I didn't because it is ambiguous?

src/scippneutron/chopper/disk_chopper.py
SimonHeybrock: When is that called?

src/scippneutron/chopper/disk_chopper.py
SimonHeybrock: If we don't, how to you know what is slits and what is edges?

Why not switch to 2 1-D arrays to avoid the headache entirely?

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
jl-wynen: But then, can't you ask the same question about $\tilde{\theta}$? It would not be shown in the plot. I can just add a sentence describing their effects.

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: Right 👍

src/scippneutron/chopper/disk_chopper.py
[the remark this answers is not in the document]
SimonHeybrock: GitHub is still not showing them?

src/scippneutron/chopper/disk_chopper.py
SimonHeybrock: Yes we generally need to replicate.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: I did this now.

A reason for combining them is that this way, we guarantee that begin and end edges are associated with each other correctly.

src/scippneutron/chopper/disk_chopper.py
jl-wynen: I moved all slit processing logic into `DiskChopper`.

SimonHeybrock (commented):
Given that we have had extensive review, a lot of docs were written (👍 ), and testing is very thorough (👍 ), would it be beneficial to just move on (and merge this)? As we get more data and develop a better understanding we will likely need to make some changes anyway, and given that there are no users of this code now, I think it would improve how we can move forward, considering also interaction with other work (such as frame unwrapping).

jl-wynen:
I'm not comfortable merging this yet as I made large additions recently. Can you at least have a look at the pre-processing notebook in the user guide. Esp. the `Compute Chopper Phase` section? I'm not sure I found a sensible solution.

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4a878315-1107-4048-bd08-88c3140846b2",
+   "metadata": {},
+   "source": [
+    "We use [find_plateaus](../../generated/modules/scippneutron.chopper.filtering.find_plateaus.rst) and [collapse_plateaus](../../generated/modules/scippneutron.chopper.filtering.collapse_plateaus.rst) to find those plateaus.\n",
+    "Note the `atol` and `min_n_points` parameters, they need to be tuned for the specific input data."
+   ]
+  },
```
SimonHeybrock, line 139: What is the integer plateau coord for? Does it have a meaning?

Without having looked at other bits, it seems `time` would make more sense as the inner dim, not the outer, as it is now?

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "413583ea-931d-4357-adec-d1e0a2a29d21",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "aux = sc.DataArray(sc.scalar(0).broadcast(sizes=tdc.sizes),\n",
```
SimonHeybrock, line 286:
```suggestion
    "aux = sc.DataArray(sc.ones_like(tdc),\n",
```
maybe, for brevity?

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+    "                   coords={'time': tdc})\n",
+    "low = frequency.coords['time'][0]\n",
+    "high = frequency.coords['time'][1]\n",
+    "tdc_in_phase = aux['time', low:high].coords['time']\n",
+    "tdc_in_phase"
```
SimonHeybrock, line 291: Alternative, without helper data array?
```python
tdc_in_phase = tdc[(tdc > low) & (tdc < high)]
```

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "**Note**\n",
+    "\n",
+    "The pulse time is typically an array of timestamps and it can be difficult to determine which pulse goes with which chopper period.\n",
+    "While the choice is technically arbitrary, the times calculated by `DiskChopper` are relative to the chose pulse time.\n",
+    "\n",
+    "If the chopper rotates at the pulse frequency or an integer multiple of it, we can select any pulse time and TDC timestamp and simply use `phase = phase % (2 * sc.constants.pi)` below.\n",
+    "This corresponds to selecting the pulse and TDC times that are closest to each other.\n",
+    "\n",
```
SimonHeybrock, line 354: Not sure I get this. Is this about selecting either the "even" or "odd" pulses?

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+    "</div>\n",
+    "\n",
+    "(We multiply by 1rad to get the proper `rad*Hz` unit in `omega`.)"
```
SimonHeybrock, line 357:
```suggestion
    "(We multiply by 1 rad to get the proper `rad*Hz` unit in `omega`.)"
```

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "0a7e7d47-1c94-4ee4-92d0-a1169f8ab958",
+   "metadata": {},
+   "source": [
+    "## Build `DiskChopper`\n",
+    "\n",
+    "Finally, we can assemble all data into a [scippneutron.chopper.DiskChopper](../../generated/modules/scippneutron.chopper.disk_chopper.DiskChopper.rst) object.\n",
+    "\n",
+    "The rotation speed gets rounded (resulting in 14Hz) because `DiskChopper` requires it to be a near exact integer multiple of the pulse frequency or vice versa."
```
SimonHeybrock, line 382: Any suggestion on how to do this if the pulse frequency is not an integer?

SimonHeybrock (commented):
Review only of the pre-processing notebook:

Good description, small questions and comments:

src/scippneutron/chopper/filtering.py
```diff
@@ -0,0 +1,161 @@
...
+
+    min_n_points = (
+        min_n_points
+        if isinstance(min_n_points, sc.Variable)
+        else sc.index(min_n_points)
+    )
+    diff = abs(data.data[1:] - data.data[:-1])
+    group_id = sc.cumsum((diff > atol).to(dtype='int64'))
+    # Prepend a 0 to align the groups with the data points (diff reduces length by 1).
+    group_id = sc.concat([sc.index(0, dtype='int64'), group_id], dim=diff.dim)
```
SimonHeybrock, line 59: I don't understand how this whole function finds plateaus. From what I see here, it seems like it finds sections with "jumps" in between (by looking at adjacent differences), but a small and stead drift would be seen as a single plateau?

What am I missing?

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+   "metadata": {},
+   "source": [
+    "We can see that there is some information about the slits and geometry of the chopper as well as some timing-related data.\n",
+    "Take a look at the [NXdisk_chopper](https://manual.nexusformat.org/classes/base_classes/NXdisk_chopper.html) documentation for an overview of the fields.\n",
+    "\n",
+    "In this case, there already is a `position`.\n",
+    "This typically needs to be computed first, see [scippnexus.compute_positions](https://scipp.github.io/scippnexus/generated/functions/scippnexus.compute_positions.html).\n",
+    "\n",
+    "Some fields are nested data groups which happens when a NeXus file contains `NXlog`s.\n",
+    "We can extract the relevant arrays from them using `post_process_disk_chopper`:"
```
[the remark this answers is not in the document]
jl-wynen: Yeah, I went back and forth about whether this is post or pre. It's post from the point of view of nexus but pre from the point of view of `DiskChopper`.
Should I change it to 'pre' everywhere?

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+    "chopper"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "46ac5332-afa3-46cb-8f55-45936c17f1e0",
+   "metadata": {},
+   "source": [
+    "Some data varies with time.\n",
+    "This is difficult or resource intensive to deal with, e.g., when [unwrapping frames](./frame-unwrapping.ipynb).\n",
```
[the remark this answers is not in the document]
jl-wynen: How would you put it? I don't know which part you're unhappy with.

docs/user-guide/chopper/pre-processing.ipynb
```diff
@@ -0,0 +1,452 @@
...
+    "So we need to extract time-independent quantities from the raw chopper data.\n",
+    "\n",
+    "## Identify In-phase Regions\n",
+    "\n",
+    "Frame unwrapping is only feasible when the chopper is in-phase with the neutron source pulses because, otherwise, wavelength ranges and resolutions would change with time.\n",
```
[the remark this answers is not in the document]
jl-wynen: Not necessarily. I think, you could have an arrangement of choppers that are all out of phase but in such a way that you don't get overlaps. Then the wavelengths that can make it through vary pulse-by-pulse and so does the resolution.

docs/user-guide/chopper/pre-processing.ipynb
jl-wynen: > What is the integer plateau coord for? Does it have a meaning?

Index of the plateau. See the plot below.

> Without having looked at other bits, it seems time would make more sense as the inner dim, not the outer, as it is now?

Why does the order matter?

docs/user-guide/chopper/pre-processing.ipynb
jl-wynen: Could do. I wanted to save memory. But that shouldn't matter here

docs/user-guide/chopper/pre-processing.ipynb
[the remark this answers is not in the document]
SimonHeybrock: I think it would just be meaningless, resources or difficulty are not the problem, I think?

docs/user-guide/chopper/pre-processing.ipynb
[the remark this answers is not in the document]
SimonHeybrock: So you are thinking about some rational (N/M) phase relationship?

docs/user-guide/chopper/pre-processing.ipynb
SimonHeybrock: It does not much, but it is easier to understand the meaning (for me at least).

docs/user-guide/chopper/pre-processing.ipynb
jl-wynen: Not directly, this is about how the chopper frames are aligned with the pulses. If the chopper is half as fast as the source, then yes, this selects even or odd pulses.

docs/user-guide/chopper/pre-processing.ipynb
jl-wynen: It needs to be an integer *multiple*. See the discussions about infeasibility above for why that is required.
It may be necessary to relax the tolerance if the actual frequencies are not integers because of float precision...

src/scippneutron/chopper/filtering.py
jl-wynen: > but a small and steady drift would be seen as a single plateau?

Correct. The tolerance needs to be tuned. But I don't know of a way that doesn't have a free parameter.

docs/user-guide/chopper/pre-processing.ipynb
SimonHeybrock: Do you need to provide one pulse for every TDC, or doe you just need one?

src/scippneutron/chopper/filtering.py
SimonHeybrock: So you agree that this is not finding plateaus, and we need to come up with a different solution?

docs/user-guide/chopper/pre-processing.ipynb
jl-wynen: Let's talk about this in person tomorrow!

src/scippneutron/chopper/filtering.py
jl-wynen: Not technically, but it should be close enough.
Alternatively, we could also select all points that are in phase and identify plateaus as contiguous groups of those points. I tried that at some point and it seemed to work. But it requires a less tight tolerance on the in phase check.

SimonHeybrock: Why do you think it is close enough? I don't see that.

jl-wynen: It would only allow a drift that is smaller than the noise. I don't have enough real data to judge whether that can happen. But it seems like there would be an error in the chopper control software if it does happens.

SimonHeybrock: Why would there be an error? I think having drift smaller than noise might be perfectly possible. I don't think we have any information that would indicate otherwise?

jl-wynen: But this would mean that the data is unusable? If a user sets a frequency but the chopper actually drifts away from it without correction, then the control software is buggy.

SimonHeybrock: Not sure what you are referring to.

If the user sets a frequency, there will be a ramp-up time (and a ramp-down time), as in your figures. If the change from point to point during the ramps is less than your threshold (the noise), the ramps will be identified as plateaus.

jl-wynen: Those would be very slow ramps. In the data I've seen, the ramps are orders of magnitude steeper than the noise. The noise in that data is ~10mHz. If the ramp is less than the noise, this would correspond to a ramp up from 0 to pulse frequency of >15min. Is this realistic?

SimonHeybrock: I have no idea, but unless we are sure that this is a good assumption in 100% of cases, shouldn't we write this code in a way that identifies plateaus, instead of sections between jumps?

src/scippneutron/chopper/filtering.py
```diff
@@ -68,9 +80,41 @@ def find_plateaus(
...
+            UserWarning,
+            stacklevel=2,
+        )
     return plateaus
 
 
+def _derive(da: sc.DataArray) -> sc.Variable:
+    x = da.coords[da.dim]
+    y = da.data
+    return (y[1:] - y[:-1]) / (x[1:] - x[:-1])
```
SimonHeybrock, line 99: Would this become less noise sensitive by using a $\Delta > 1$?

jl-wynen: Somewhat because the divisor becomes larger. But it would reduce the resolution as you can only find the edges of a plateau within a down to a range of $\Delta$.

If you want to treat noise, I'd be more inclined to smooth the data.

g5t: If you're concerned about sensitivity and identifying where the slope changes, why not use the central difference? https://en.wikipedia.org/wiki/Finite_difference

src/scippneutron/chopper/filtering.py
```diff
@@ -0,0 +1,209 @@
...
+    return _is_approximate_multiple(frequency.data, ref=reference, rtol=rtol)
+
+
+def filter_in_phase(
+    frequency: sc.DataArray, *, reference: sc.Variable, rtol: sc.Variable
+) -> sc.DataArray:
+    """Remove all elements where a given frequency is not in phase with a reference.
+
+    Frequencies are considered to be in-phase if they are an integer multiple
+    of the reference or vice versa.
```
g5t, line 192: The name of this function seems misleading to me, namely I expected it to be concerned with whether the `phase` value of the chopper matches the target set point.
I agree that if the ratio between chopper and source frequencies is irrational the chopper _can not be_ in phase, but think that an 'in phase' check should be more restrictive.

Perhaps a name like `filter_is_phase_locked` would indicate that the function does not evaluate whether the recorded phase value is correct, just that the phase value is meaningful in some way?

src/scippneutron/chopper/_svg.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def _combine_slits(chopper: DiskChopper) -> sc.DataArray:
+    edges = sc.concat(
+        [chopper.slit_begin.flatten(to='slit'), chopper.slit_end.flatten(to='slit')],
+        dim='edge',
+    )
+    height = chopper.slit_height.rename_dims({chopper.slit_height.dim: 'slit'})
+    slits = sc.DataArray(
+        sc.arange('slit', edges.shape[0], unit=None),
```
g5t, line 66: In one case `edges.shape[0]` seems to be the wrong dimension. A chopper with one opening has 1-element 1-dimension `slit_begin` and `slit_end`, and produces a `edges` array where `shape[0]` corresponds to the edges instead of the slits.
Replacing `edges.shape[0]` by `edges.sizes['slit']` works in this case, and seems like it should always be correct.

src/scippneutron/chopper/filtering.py
g5t: Another suggestion: `filter_commensurate` or `filter_out_incommensurate`

docs/user-guide/chopper/pre-processing.ipynb
nvaytet: Hmm should that have been `zeros_like`?

docs/user-guide/chopper/pre-processing.ipynb
nvaytet: Done

src/scippneutron/chopper/filtering.py
nvaytet: Thanks for the suggestion. We will keep it as is now, and if we encounter stability problems later, we can change the internal method for a more robust one.

src/scippneutron/chopper/filtering.py
nvaytet: Thanks for the suggestion. It's true that `in_phase` is a little misleading.
However, we feel `(in)commensurate` would be a little unclear also?

We couldn't come up with a name that solved everything and that we were happy with, so we'll leave it as is for now. There is an explanation of what we mean by "in phase" in the docstring, so it's probably fine?

jl-wynen:
Rebased onto main and fixed ruff errors. There were some complicated merges in the docs. Need to inspect the built docs before merging.

nvaytet (approved):

PR #468, opened 2023-11-20
Chopper cascade

### LINKED ISSUE #467: Process resolution-chopper info to define `frame_offset` required for frame-unwrapping
https://scipp.github.io/scippneutron/reference/frame-unwrapping.html requires `frame_offset`. This may just be zero (or some fixed number) for instruments using the full pulse. On instruments with resolution choppers this needs to be calculated from the relevant chopper params.

This will also be required in the context of WFM, if we consider the WFM choppers as resolution choppers with multiple openings.

SimonHeybrock:
This computes the result of applying a chopper cascade to a source pulse. The goal is to use the result for defining time bounds for [frame-unwrapping](https://scipp.github.io/scippneutron/reference/frame-unwrapping.html) and WFM.

Fixes #467, but on a more generic level.

~This is not ready yet and only meant for a review of the concept. I suggest to run the added notebook.~

Here is a comparison withe the result computed by `tof` (for the [example from the tof docs](https://tof.readthedocs.io/en/stable/short-example.html)):

image

This is a stack of the "accepted" neutrons after every chopper, starting from the source pulse (blue) and the finally result propagated to the desired target distance. The blue line is the result computed by `tof`. This is consistent.

We can also draw a chopper acceptance diagram:
Screenshot 2023-11-22 at 06 44 21


src/scippneutron/tof/chopper_cascade.py
```diff
@@ -0,0 +1,383 @@
...
+        frame = self.propagate_to(chopper.distance)
+
+        # A chopper can have multiple openings, call _chop for each of them. The result
+        # is the union of the resulting subframes.
+        chopped = Frame(distance=frame.distance, subframes=[])
+        for subframe in frame.subframes:
+            for open, close in zip(chopper.time_open, chopper.time_close):
+                if (tmp := _chop(subframe, open, close_to_open=True)) is not None:
+                    if (tmp := _chop(tmp, close, close_to_open=False)) is not None:
+                        chopped.subframes.append(tmp)
```
nvaytet, line 185: Optional:

Something like `chopped.add_subframe(tmp)` ?
(in case one day we change the subframes container from a list to something else?)

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -0,0 +1,383 @@
...
+        simply indicate a problem with the chopper cascade. Attempts to handle this
+        automatically may be misleading.
+        """
+        starts = [subframe.start_time for subframe in self.subframes]
+        ends = [subframe.end_time for subframe in self.subframes]
+        # Given how time-propagation and chopping works, the min wavelength is always
+        # given by the same vertex as the min time, and the max wavelength by the same
+        # vertex as the max time. Thus, this check should generally always pass.
+        # Exceptions may be subframes that have been created manually.
+        if not all(subframe.is_regular() for subframe in self.subframes):
```
nvaytet, line 218: Is this check also needed when computing `bounds` above?

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -0,0 +1,383 @@
...
+                'coincide.'
+            )
+        wav_starts = [subframe.start_wavelength for subframe in self.subframes]
+        wav_ends = [subframe.end_wavelength for subframe in self.subframes]
+        # sort by start
+        starts, ends, wav_starts, wav_ends = zip(
+            *sorted(zip(starts, ends, wav_starts, wav_ends), key=lambda x: x[0])
+        )
+        bounds = []
+        current = (starts[0], ends[0], wav_starts[0], wav_ends[0])
```
nvaytet, line 230: Use a small dataclass instead of a tuple?

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -0,0 +1,383 @@
...
+        time = sc.concat([time_min, time_max, time_max, time_min], dim='vertex').to(
+            unit='s'
+        )
+        wavelength = sc.concat(
+            [wavelength_min, wavelength_min, wavelength_max, wavelength_max],
+            dim='vertex',
+        ).to(unit='angstrom')
+        self._frames = [
+            Frame(
+                distance=sc.scalar(0, unit='m'),
```
nvaytet, line 281: Do we allow for negative distances? e.g. if one day these numbers are taken from the `source_position` in the file?
Or will we be operating on the norm of vectors (e.g. `sample_position - source_position`) anyway, and this does not matter?

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -0,0 +1,383 @@
...
+    def draw(self) -> Any:
+        """Draw frames using matplotlib"""
+        import matplotlib.colors as mcolors
+        import matplotlib.patches as patches
+        import matplotlib.pyplot as plt
+
+        fig, ax = plt.subplots()
+        colors = list(mcolors.TABLEAU_COLORS.values())
+        colors += colors
+        colors += colors
```
nvaytet, line 326: There is a (small) chance you might run out of colors. You could instead use the `C<NUMBER>` notation for colors, which I think are pretty equivalent to the TABLEAU colors?

Below, you would just have `for i, frame in enumerate(self._frames):`
And then after that, `ax.plot(..., color=f'C{i}')`.
It just automatically does a modulo on the colors, so it wraps round after 10 colors.

docs/reference/chopper-cascade.ipynb
```diff
@@ -0,0 +1,327 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def from_tof_chopper(chopper: tof.Chopper) -> chopper_cascade.Chopper:\n",
+    "    \"\"\"Convert a chopper from the tof package to a chopper for the chopper-cascade package.\"\"\"\n",
+    "    return chopper_cascade.Chopper(\n",
+    "        distance=chopper.distance,\n",
+    "        time_open=chopper.open_close_times()[0].to(unit='s'),\n",
```
nvaytet, line 178: By supplying no end time to the `chopper.open_close_times()`, you are only performing a single chopper rotation.
If some choppers are rotating very fast, they could rotate more than once inside a single pulse?

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: I don't think so, since the global bounds are always "unique". Here we have the check because we need to merge subframes, and the logic would (I think) fall apart if this condition is not fulfilled.

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: We will be operating on the differences, yes.

docs/reference/chopper-cascade.ipynb
SimonHeybrock: I didn't know that, thanks!

Note`tof` will be removed here anyway, I think, since I did not intend to make it a dependency of `scippneutron`. Will get times from the chopper classes in ScippNeutron once merged.

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: 👍

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: I case we do that, we can still change it then?

docs/reference/chopper-cascade.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "wfm1 = chopper_cascade.Chopper(\n",
+    "    distance=sc.scalar(6.6, unit='m'),\n",
+    "    time_open=sc.array(\n",
+    "        dims=('cutout',),\n",
+    "        values=[\n",
+    "            -0.008500,\n",
+    "            -0.006246,\n",
+    "            -0.004152,\n",
+    "            -0.002206,\n",
+    "            -0.000396,\n",
```
nvaytet, line 55: I'm not sure you need the negative times in your case.
In `tof`, we have them because we need to make sure the last slits from the previous rotation also appear in the diagram, but for your purposes I think you can leave them out?

SimonHeybrock: Ok, seems to work, removed.

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -416,6 +417,46 @@ def draw(
...
+        frames = FrameSequence(
+            [frame.propagate_to(source_distance) for frame in self.frames]
+        )
+        # Reset frame distance for plotting labels. This is a bit of a hack.
+        # We should use chopper names if available.
+        for i, frame in enumerate(frames.frames):
+            frame.distance = self.frames[i].distance
+        blue = np.linspace(0.1, 1, len(frames.frames))
+        colors = plt.cm.Blues(blue)
+        fig, ax = frames.draw(
```
nvaytet, line 457: Just to check: are you guaranteed that the order of the frames means that the furthest chopper is plotted last?
This would be needed to make sure that the polygons are visible, and not hidden under a larger polygon.

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -416,6 +417,46 @@ def draw(
...
+        source_distance = self.frames[0].distance
+        frames = FrameSequence(
+            [frame.propagate_to(source_distance) for frame in self.frames]
+        )
+        # Reset frame distance for plotting labels. This is a bit of a hack.
+        # We should use chopper names if available.
+        for i, frame in enumerate(frames.frames):
+            frame.distance = self.frames[i].distance
+        blue = np.linspace(0.1, 1, len(frames.frames))
+        colors = plt.cm.Blues(blue)
```
nvaytet, line 456: Can you make the colormap an argument to the function? Some people might not like the simple shades of blue?

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: Yes, the choppers must be ordered, there is a check that does not allow for propagating backwards.

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: How? What do you have in mind?

Can we fix it once someone says they don't like it?

nvaytet: Have a `cmap` argument, that defaults to `'Blues'`, and use `getattr(plt.cm, cmap)`?

I would be the first one to say "I'd like to change the colormap" ;-)

SimonHeybrock: That would make Matplotlib a mandatory requirement, I suppose?

nvaytet (approved):

PR #471, opened 2023-11-28
Use copier template

jl-wynen:


jl-wynen:
Ready for review, seems to be working now.

.github/workflows/pr_and_main.yml
```diff
@@ -59,19 +43,3 @@ jobs:
 
       - run: conda develop src
       - run: python -m pytest
-
```
SimonHeybrock, line 62: Re above: How is (and was?) this installing Mantid? The env file has the `mantid` *channel*, but I cannot see Mantid being installed?

conda/meta.yaml
```diff
@@ -30,8 +14,8 @@ requirements:
     - h5py
     - numpy>=1.20.0
     - pooch
-    - python
-    - scipp>23.08.0
+    - python>=3.9
+    - scipp>23.07.0
```
SimonHeybrock, line 18: Why was it scipp>23.08 previously, but reduced now?

docs/about/index.md
```diff
@@ -0,0 +1,38 @@
...
+Please cite the following:
+
+[![DOI](https://zenodo.org/badge/305999676.svg)](https://zenodo.org/doi/10.5281/zenodo.7447231)
+
+To cite a specific version of ScippNeutron, select the desired version on Zenodo to get the corresponding DOI.
+
+## Older versions of the documentation
+
+Older versions of the documentation pages can be found under the assets of each [release](https://github.com/scipp/scippneutron/releases).
+Simply download the archive, unzip and view locally in a web browser.
```
SimonHeybrock, line 26: Should we manually upload the old versions before they are wiped from `gh_pages`?

docs/api-reference/index.md
```diff
@@ -0,0 +1,86 @@
+# API Reference
+
+## Top-level functions
+
+### Mantid Compatikbility
```
SimonHeybrock, line 5:
```suggestion
### Mantid Compatibility
```

docs/api-reference/index.md
```diff
@@ -0,0 +1,86 @@
...
+## Frame handling
+
+```{toctree}
+---
+maxdepth: 1
+---
+
+chopper-cascade
+frame-unwrapping
+```
```
SimonHeybrock, line 86: Aren't those auto-documented via the` tof` submodule above?

pyproject.toml
```diff
@@ -19,18 +18,23 @@ classifiers = [
...
+
+# IMPORTANT:
+# Run 'tox -e deps' after making changes here. This will update requirement files.
+# Make sure to list one dependency per line.
 dependencies = [
     "h5py",
     "numpy>=1.20",
     "pooch",
-    "scipp>23.08.0",
+    "scipp>=23.07.0",
```
SimonHeybrock, line 37: Can you explain?

.github/workflows/pr_and_main.yml
jl-wynen: It's right here: https://github.com/scipp/scippneutron/blob/c89e916728c8f4bad4b00134826bf575ec88b149/.buildconfig/ci-linux.yml#L16

SimonHeybrock: Hmm, I was looking at the ci-linux-39 file, why do we have both?

conda/meta.yaml
jl-wynen: This was part of the migration away from attrs. But we since found out that 23.08 is incompatible with mantid and I changed the code to work with both old and new Scipp versions. But I forgot to revert the requirement. So it was impossible to install a new ScippNeutron with Mantid.

docs/about/index.md
jl-wynen: Yes

pyproject.toml
jl-wynen: Same as above

docs/api-reference/index.md
jl-wynen: No. Those are notebooks in `docs/api-reference`.
It's a bit strange to have them in "API reference" but this is where they were before.

Issue #469, opened 2023-11-23
Include non-zero transition-time from open<->closed in `tof.chopper_cascade`


SimonHeybrock:
This is due to a non-zero guide side. See https://backend.orbit.dtu.dk/ws/portalfiles/portal/102784335/1.4900958.pdf Fig. 3 and, e.g., Eq. (6) and (7).

SimonHeybrock:
@nvaytet Should this be moved to `essreduce`?

nvaytet:
We left the `chopper_cascade` module in `scippneutron`, so I would say no, it should stay here?
If you want to make more accurate chopper acceptance diagrams, it should still be added here.

In `essreduce`, we will either use `tof` or `McStas` to compute time-of-flight.
McStas already does this. I guess it should be possible to add to `tof` relatively easily? (have some sort of transmission function convoluted to the opening of each chopper? although currently neutrons either go through or not, so choppers would now have a non-boolean opening, so it might be a larger change...)

nvaytet:
Because of the move towards using `tof` and McStas to compute time-of-flight, this has moved down the priority list.
I'm removing this from triage for now.

SimonHeybrock:
Question is, should it be reformulated to "Include non-zero transition-time from open<->closed in `tof` simulation"? Isn't this effect still equally relevant? Or are you counting on using McStas in production so this is addressed automatically?

nvaytet:
I was going to rely on mcstas, but we can also transfer this issue to `tof`, so that it cna also handle this case?

PR #472, opened 2023-11-30
Frame unwrapping and WFM

SimonHeybrock:
This represents the latest approach to [frame unwrapping](https://scipp.github.io/scippneutron/reference/frame-unwrapping.html) and WFM (see old #415).

This clearly makes a bunch of assumptions which may turn out to be wrong in practice. At this point I have little confidence that this will be anything close to the final solution (or even that there is a solution at all &mdash; in the end we might still be forced to fall back to defining bounds more manually).

Testing is obviously challenging, and given that it is still not clear how all the choppers will be handled, I believe we must iterate on the whole solution (including what @jl-wynen has kicked of with #466) in the coming months.

Notes:
- The old `tof.frames` was removed.
- Adds `tof.fakes`, which I hope we can develop further for these and similar tests. This is currently my only idea on how to move forward: Build more realistic fakes, fix the chopper and unwrapping code, iterate.
- Unlike the old unwrapping code, this can also handle `event_time_zero`. This will be essential for experiments with high time-resolution requirements.
- Relies on the recently added `scippneutron.tof.chopper_cascade` module
- My initial benchmarks indicate that using `scipp.lookup` over `scipp.where` and `scipp.bin` during unwrapping and WFM stitching (respectively) has equivalent or better performance. I think it is also clearer.
- I added Sciline to the docs builds (so we can show the graphs), but not as a Scippneutron requirement. That is, we will provide providers (:grimacing: ) and domain types, but not more.

Fixes scipp/ess#165.

Follow-up work:
- #482 
- #483 
- #469


nvaytet:
I had a short look but I think it would be much easier for me if we looked at the graphs together and you can explain the different steps, as well as the differences between the different graphs.

src/scippneutron/tof/unwrap.py
```diff
@@ -142,31 +145,36 @@ def frame_period(
...
         time_max=source_time_range[1],
         wavelength_min=source_wavelength_range[0],
         wavelength_max=source_wavelength_range[1],
     )
     frames.chop(choppers.values())
-    return frames[-1].propagate_to(l1)
+    # Find last frame before ltotal
+    frame_before_detector = None
+    for frame in frames:
+        if frame.distance > ltotal:
```
nvaytet, line 164: Just checking: I guess we are guaranteed that the frames are ordered according to distance?

src/scippneutron/tof/unwrap.py
```diff
@@ -264,9 +278,23 @@ def offset_from_wrapped(
 
 
 def source_chopper(
-    choppers: Choppers, source_chopper_name: SourceChopperName
+    choppers: Choppers,
+    source_time_range: SourceTimeRange,
+    source_chopper_name: Optional[SourceChopperName],
```
nvaytet, line 283: Should `source_chopper_name` have a default `None` value?

src/scippneutron/tof/unwrap.py
```diff
@@ -264,9 +278,23 @@ def offset_from_wrapped(
...
 ) -> SourceChopper:
-    return choppers[source_chopper_name]
+    """
+    Return the chopper defining the source location and time-of-flight time origin.
+
+    If there is no pulse-shaping chopper, then the source-pulse begin and end time
+    are used to define a fake chopper.
+    """
+    if source_chopper_name is not None:
+        return choppers[source_chopper_name]
```
nvaytet, line 292: So if the `source_chopper_name` exists, then `source_time_range` is unused. Should that also be an optional input?

src/scippneutron/tof/unwrap.py
SimonHeybrock: Yes, `chopper_toolkit` sorts the choppers before applying them.

src/scippneutron/tof/unwrap.py
SimonHeybrock: Sciline does not use default values.

src/scippneutron/tof/unwrap.py
SimonHeybrock: I don't think so, because then the pipeline would fail at runtime if neither is given?

nvaytet: Say you know the source chopper name, but not the source time range, you'd still have to supply a source time range as input.
If you don't know it, you could always make it to be nonsense because it won't be used, but it seems a little strange to require it?

SimonHeybrock: Without a source time range the whole chopper-cascade math won't work. This is required anyway for other parts of the workflow.

SimonHeybrock: How would that be possible?

nvaytet: Couldn't we check that the chopper open and close times match the `source_time_range`? Or maybe I misunderstood something and in some part of the pipeline you need the source chopper, and in others you need the `source_time_range` and they can in principle be different? (e.g. if the source chopper is not at a distance of 0.0?)

SimonHeybrock: The "source chopper" may define a different time range, i.e., it is a pulse-shaping chopper to increase the resolution. It will typically have a tighter time range.

nvaytet (commented):
I still need to take a closer look at the tests, a couple of comments for now

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def pulse_wrapped_time_offset(da: RawData) -> PulseWrappedTimeOffset:
+    """
+    In NXevent_data, event_time_offset is the time offset since event_time_zero,
+    which is the start of the most recent pulse. This is not the same as the start
+    of the pulse that emitted the neutron, so event_time_offset is "wrapped" around
+    at the pulse period.
+    """
+    if da.bins is None:
```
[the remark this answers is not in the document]
SimonHeybrock: In `NXdetector` it is also named `time_of_flight` so it might just work, but I would not expect to get this right without knowing what we actually get. For now I am assuming we only have event mode detectors.

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,462 @@
...
+    if len(source_chopper.time_open) != subframe_bounds['time'].sizes['subframe']:
+        raise NotImplementedError(
+            "Source chopper openings do not match the subframe count, this is ."
+            "not supported yet."
+        )
+    times = subframe_bounds['time'].flatten(dims=['subframe', 'bound'], to='subframe')
+    neg_shift = sc.zeros(
+        dims=['subframe'], shape=[times.sizes['subframe'] + 1], unit='s'
+    )
+    padding = sc.scalar(1000.0, unit='s').to(unit=times.unit)
```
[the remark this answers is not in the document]
SimonHeybrock: Added both.

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,462 @@
...
+    source_chopper :
+        Chopper defining the source location and time-of-flight time origin.
+    """
+    if len(source_chopper.time_open) != 1:
+        raise NotImplementedError(
+            "Source chopper with multiple openings not supported yet."
+        )
+    source_time_open = source_chopper.time_open[0]
+    source_time_close = source_chopper.time_close[0]
+    time_zero = 0.5 * (source_time_open + source_time_close)
```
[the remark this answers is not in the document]
SimonHeybrock: Extended param description in docstring

nvaytet:
I have no further comments here, but are we still waiting to have an in-person discussion about this?

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,491 @@
...
+    return OffsetFromTimeOfFlight(time_offset - time_zero)
+
+
+def offset_to_time_of_flight_wfm(
+    time_offset: OffsetFromWrapped,
```
SimonHeybrock, line 381: I think this may actually be wrong. I think we need to unwrapped time offset, not just the offset from wrapped to perform the WFM TOF offsetting.

Converting to draft until verified.

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,491 @@
...
+    pulse-skipping mode, i.e., the caller must know which pulses are in use.
+    """
+    frames = chopper_cascade.FrameSequence.from_source_pulse(
+        time_min=source_time_range[0],
+        time_max=source_time_range[-1],
+        wavelength_min=source_wavelength_range[0],
+        wavelength_max=source_wavelength_range[-1],
+    )
+    frames = frames.chop(choppers.values())
+    return FrameAtDetector(frames[ltotal])
```
jokasimr, line 180: Doesn't this assume:

* that the neutrons move in straight lines to the detector element
* that there is no inelastic scattering

SimonHeybrock: Correct. See my comment in the PR:

> This clearly makes a bunch of assumptions which may turn out to be wrong in practice. At this point I have little confidence that this will be anything close to the final solution (or even that there is a solution at all — in the end we might still be forced to fall back to defining bounds more manually).

SimonHeybrock:
@nvaytet @jl-wynen @jokasimr I'd like some input here, on the latest changes.

docs/api-reference/frame-unwrapping.ipynb
jl-wynen, line 1: Typo in l73: `"For, e.g., detector 1 when then observe:\n",` should be `"For, e.g., detector 1 we then observe:\n",`. And 'then' can be left out, too.

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
-    "   $$\n",
-    "   t_{\\text{tof}} =  t_{\\text{min}} - t_\\text{pivot} +\\Delta t + \n",
-    "       \\begin{cases}\n",
-    "       L_0, & \\text{for } \\Delta t \\lt t_\\text{pivot}\\\\\n",
-    "       0, & \\text{for } \\Delta t \\ge t_\\text{pivot}\n",
-    "       \\end{cases}\n",
-    "   $$\n",
-    "   \n",
-    "Note that there are other valid definitions of $t_\\text{pivot}$, differing only in how neutrons in the intermediate \"background\" region (region between two pulses) are mapped to frames."
+    "As illustrated in the figure, the pivot time $t_\\text{pivot}$ depends on the detector or rather the distance of the detector (or monitor) from the scattering position.\n"
```
jl-wynen, line 83: In the diagram, the pivot for detector 1 is at the black dashed line, i.e., the beginning of the orange frame. But in detector 2 it is in the middle between the frames.

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "ee2e08b2-6e25-49c7-a2d4-07cf90b04496",
-   "metadata": {},
-   "source": [
-    "Our fake data illustrates what we might observe when placing many neutron monitors along our beamline.\n",
-    "Note that intensities are not modelled correctly, since this is irrelevant here:"
+    "After unwrapping, the time-of-flight can be computed, by defining an origin time and distance $L_1$.\n",
```
jl-wynen, line 142:
```suggestion
"After unwrapping, the time-of-flight can be computed by defining an origin time and a distance $L_1$.\n",
```

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
+    "If we, e.g., define the start of time-of-flight at a chopper we must also adapt the distance $L_1$ between the origin and the sample, which will be used later on to compute, e.g. the wavelength or energy of the neutron.\n",
```
jl-wynen, line 143:
```suggestion
    "If we, e.g., define the start of time-of-flight at a chopper, we must also adapt the distance $L_1$ between the origin and the sample which will be used later on to compute, e.g., the wavelength or energy of the neutron.\n",
```

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
+    "If defined via, e.g., a calibration this can also be used to deal with curved guides or other non-linearities in the neutron path."
```
jl-wynen, line 144:
```suggestion
    "If defined via, e.g., a calibration, this can also be used to deal with curved guides or other non-linearities in the neutron path."
```

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
-    "unwrapped.plot()"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "973b8629-fcd5-440b-9587-9cdce789737a",
-   "metadata": {},
-   "source": [
-    "At the sample or a detector we would only see this for a particular `Ltotal`:"
+    "Choppers may be used to skip pulses, for the purpose of a simultaneous study of a wider wavelength range.\n",
```
jl-wynen, line 206: What do you mean by `simultaneous` here?

src/scippneutron/tof/unwrap.py
```diff
@@ -0,0 +1,573 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+"""
+This module provides functionality for unwrapping raw frames of neutron time-of-flight
+data.
+
+The modules handles standard unwrapping, unwrapping in pulse-skipping mode, and
```
jl-wynen, line 8:
```suggestion
The module handles standard unwrapping, unwrapping in pulse-skipping mode, and
```

jl-wynen (commented):
I don't have anything enlightening to say. This just looks like what we talked about.

src/scippneutron/tof/unwrap.py
```diff
@@ -378,7 +388,9 @@ def time_of_flight_origin_from_chopper(
     source_time_open = source_chopper.time_open[0]
     source_time_close = source_chopper.time_close[0]
     time_zero = 0.5 * (source_time_open + source_time_close)
-    return TimeOfFlightOrigin(time_zero)
+    return TimeOfFlightOrigin(
+        time=time_zero, L1=sample_distance - source_chopper.distance
```
nvaytet, line 392: Is the `TimeOfFlightOrigin` ill-defined if there is no sample in the run, or if you are trying to unwrap times for a monitor?

src/scippneutron/tof/unwrap.py
```diff
@@ -423,7 +437,10 @@ def time_of_flight_origin_wfm_from_chopper(
     ).rename_dims(cutout='subframe')
     # Set offsets before, between, and after subframes to NaN
     neg_shift[::2] = sc.scalar(math.nan, unit='s')
-    return TimeOfFlightOrigin(sc.DataArray(neg_shift, coords={'subframe': times}))
+    return TimeOfFlightOrigin(
+        time=sc.DataArray(neg_shift, coords={'subframe': times}),
+        L1=sample_distance - source_chopper.distance,
```
nvaytet, line 442: Should they be distances or should they be defined as vector positions in space and take the norm of the difference?

nvaytet (commented):
A couple of questions.

Also, maybe I've asked this before, but why is this notebook in the API reference? It makes it pretty difficult to find in the docs I feel.

docs/api-reference/frame-unwrapping.ipynb
```diff
@@ -77,420 +77,165 @@
...
-    "We refer to each of the emitted bundles of neutrons as *frame*.\n",
+    "### Time-of-flight origin\n",
     "\n",
-    "In practice, NeXus files record event data using `NXevent_data`, and in particular the `event_time_zero` and `event_time_offset` fields.\n",
-    "`event_time_zero` gives the time of the source pulse.\n",
-    "When pulse-skipping is used, the neutrons from each *frame* may nevertheless be assigned to *pulses* in the NeXus files.\n",
-    "That is, while we have a frame-period of $2 \\cdot 71~\\text{ms}$ (in this example), `event_time_zero` is in steps of $71~\\text{ms}$.\n",
-    "The `event_time_offset` that we would obtain from a NeXus file thus records the modulus of the `time`:"
+    "The origin time (used above) could be defined by a chopper, a calibration, or even set by hand.\n",
+    "The module currently providers a naive definition via the chopper cascade:"
```
nvaytet, line 166:
```suggestion
    "The module currently provides a naive definition via the chopper cascade:"
```

jokasimr (commented):
I still don't really understand the WFM case and why it's necessary to treat it separately. My understanding was that given *any* chopper configuration we can propagate the source pulse through the cascade and compute the wavelength-time diagram at every detector(element), and from it + the event time data we can compute the time of flight. From that perspective the WFM case is just one particular chopper configuration and it's not clear to me why it needs a separate code path.

But this shouldn't be an obstacle to merge this, just a heads up that this is probably something we should talk about again.

docs/api-reference/frame-unwrapping.ipynb
SimonHeybrock: Not sure what you are describing. Doesn't it look like in the docs for you?
image

docs/api-reference/frame-unwrapping.ipynb
SimonHeybrock: You can use every pulse, measure 1-4 Angstrom, change the choppers, measure 4-8 Angstrom, and combine the results. That would be serial, as opposed to simultaneous (I think this is the term I have heard scientists use).

SimonHeybrock:
> Also, maybe I've asked this before, but why is this notebook in the API reference? It makes it pretty difficult to find in the docs I feel.

It used to be more visible, but was moved there during the switch to the new theme. @jl-wynen can comment on why it is here or where we can move it.

src/scippneutron/tof/unwrap.py
SimonHeybrock: It does not matter if there is no sample, this is about the definition of L1. This (or `sample_position`) has to be defined even if there is no "sample", since all the coord transforms in ScippNeutron rely on it.

But you have actually stumbled upon a related problem: If we are unwrapping a monitor then the coord transforms defined in ScippNeutron do not use `L1` but `Ltotal`, so we should be correcting that.

src/scippneutron/tof/unwrap.py
SimonHeybrock: The `chopper_cascade` using a simple distance, so this is what we have to use here. This will hopefully allow for hiding details such as curved guides (we do not want to `chopper_cascade` toolkit to deal with such complications, if possible).

SimonHeybrock:
> From that perspective the WFM case is just one particular chopper configuration and it's not clear to me why it needs a separate code path.

It does not a different code part, for the most part. But the time-of-flight origin needs to be defined differently, in particular different neutrons within a frame have a different origin.

src/scippneutron/tof/unwrap.py
```diff
@@ -311,7 +314,7 @@ def offset_from_wrapped(
         + sc.concat([frame_period, sc.zeros_like(frame_period)], dim),
         coords={dim: time.to(unit=elem_unit(wrapped_time_offset))},
     )
-    return OffsetFromWrapped(sc.lookup(offset, dim=dim)[wrapped_time_offset])
+    return DeltaFromWrapped(sc.lookup(offset, dim=dim)[wrapped_time_offset])
```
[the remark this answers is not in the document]
SimonHeybrock: It can handle event data, `interpolate` cannot, and may be slower.

src/scippneutron/tof/unwrap.py
```diff
@@ -334,10 +337,7 @@ def source_chopper(
     )
 
 
-def offset_to_time_of_flight(
-    time_offset: OffsetFromWrapped,
-    source_chopper: SourceChopper,
-) -> OffsetFromTimeOfFlight:
+def time_of_flight_origin(source_chopper: SourceChopper) -> TimeOfFlightOrigin:
```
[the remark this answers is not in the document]
SimonHeybrock: No, not really, it is not necessarily the first chopper, and it might not even be an actual chopper, but a "virtual" one that one might define (e.g., to use the source pulse as "source chopper").

src/scippneutron/tof/unwrap.py
```diff
@@ -135,14 +128,24 @@
...
+Detector data with time-of-flight and time zero coordinates.
 """
 
 
-OffsetFromTimeOfFlight = NewType('OffsetFromTimeOfFlight', sc.Variable)
-OffsetFromWrapped = NewType('OffsetFromWrapped', sc.Variable)
+TimeOffset = NewType('TimeOffset', sc.Variable)
+"""Unwrapped time offset relative to the pulse time."""
+
+DeltaFromWrapped = NewType('DeltaFromWrapped', sc.Variable)
```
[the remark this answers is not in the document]
SimonHeybrock: Did you review an outdated diff?

src/scippneutron/tof/unwrap.py
SimonHeybrock: @nvaytet Please have a look at the latest commit, this should fix the aforementioned issue.

nvaytet: :+1: Remember to add a test that checks for this when you get round to adding tests.

nvaytet:
Green light from me for this round of review.

Do I remember correctly that you said you were going to add tests after this round of review?

SimonHeybrock:
@nvaytet Tests updated!

src/scippneutron/tof/unwrap.py
```diff
@@ -505,14 +508,21 @@ def to_time_of_flight(
     if da.bins is not None:
         da = da.copy(deep=False)
         da.data = sc.bins(**da.bins.constituents)
-        da.bins.coords['tof'] = time_offset + delta
-        da.bins.coords['time_zero'] = da.bins.coords['pulse_time'] - delta.to(
+        da.bins.coords['tof'] = time_offset - delta
```
nvaytet, line 511: :+1:

nvaytet (approved):

Issue #473, opened 2023-12-04
Metadata utilities


jl-wynen:
## Overview

We need to handle metadata in a few places. And some of that metadata is common to many techniques and workflows. In particular, we may/will have SciCat metadata in many workflows. But there are also technique-specific file formats that encode metadata, e.g., NXCanSAS, ORSO, CIF. And they each use different names and layouts even though they have many fields in common.

To provide a common ground and unify metadata handling (at least to an extend) I propose adding common classes and maybe functions that
- let users specify metadata,
- parse input metadata (mainly SciCat) into a common form,
- write output metadata to various formats.

ScippNeutron seems to me like a good place to put these utilities as it is used by all ess* packages that implement the actual workflows. We could of course also use a separate package but that seems a little overkill.

## Suggested implementation

We can use Pydantic models to encode an validate metadata. For example, the following are commonly used:
```python
class Author(pydantic.BaseModel):
    name: str  # free form even though some file formats require a specific format
    email: pydantic.EmailStr | None = None
    orcid: ORCID | None = None  # custom type, already implemented in Scitacean
    affiliation: str | None = None
    role: str | None = None  # CIF has a fixed set of allowed roles
    corresponding: bool = False  # aka 'contact'; formatted differently in some formats
    owner: bool = True  # owner of the data
   

class Beamline(pydantic.BaseModel):
    name: str
    facility: str | None = None
    site: str | None = None
    version: str | None = None  # or 'revision'?
```
And maybe
```python
class Software(pydantic.BaseModel):
    name: str
    version: str
    url: str | None = None  # can be a DOI; or should there be a separate field?

class ProcessLog(pydantic.BaseModel):
    software: list[Software]
    steps: list[str] | None = None
    graph: str | None = None  # not neatly possible with most output formats
    log: str | None = None  # anything logged by the workflow; can be too long to embed in output files
```

The `Beamline` model could of course be extended to include technical info about the beamline / instrument such as flight path length, choppers, sample holder, etc. But the fields above are a minimum set of strictly metadata that should be common to all large scale facilities.

### Example

```python
jane = Author(
    name="Jane Doe",
    email="j.doe@psi.ch",
    orcid="https://orcid.org/0000.0000.0000.0000",
    affiliation="Paul Scherrer Institut",
    corresponding=True,
)
amor = Beamline(
    name="Amor",
    facility="SINQ",
    site="Paul Scherrer Institut",
)
```

jl-wynen:
[...]

PR #477, opened 2023-12-07
Add CIF writer

jl-wynen:
Fixes https://github.com/scipp/ess/issues/212

@celinedurniak, @AndrewSazonov any feedback is welcome!

This is a first version of a CIF writer that is still fairly low-level. But it is ready for testing in workflows.

Below there is an example for the POWGEN reduction in our docs. It is possible to add wrappers for encoding the remaining metadata. But I'd wait for #473 before doing that.

Note that the tof coord is basically made up because we have no proper focussing implementation.

```python
import scipp as sc
from scippneutron.io import cif
from datetime import datetime, timezone

da_dspacing = sc.io.load_hdf5("data/powgen_reduced_dspacing.h5")

intensity = da_dspacing.data
dspacing = sc.midpoints(da_dspacing.coords['dspacing']).to(unit='Å', copy=False)
tof = dspacing.copy(deep=False)
tof.unit = 'us'
reduced_data = sc.DataArray(
    intensity,
    coords={'tof': tof}
).rename_dims(dspacing='tof')

cal = sc.DataArray(
    sc.array(dims=['cal'], values=[1.0]),
    coords={'power': sc.array(dims=['cal'], values=[1])}
)

now = datetime.now(tz=timezone.utc)
block = cif.Block(
    "LaB6",
    [
        {
            'diffrn_radiation.probe': 'neutron',
            'diffrn_source.beamline': 'POWGEN',
            'diffrn_source.device': 'spallation',
            'diffrn_source.facility': 'Spallation Neutron Source',
        },
        {
            'pd_proc.info_datetime': now,
            'computing.diffrn_reduction': 'essdiffraction v0',
            'pd_proc.info_data_reduction': '''Normalized by proton charge
Normalized by Vanadium run 4866
Calibrated using FERNS_d4832_2011_08_24''',
        },
        {
            'audit.creation_date': now,
            'audit_contact_author.name': 'Jan-Lukas Wynen',
            'audit_contact_author.id_orcid': 'https://orcid.org/0000-0002-3761-3201',
            'audit_contact_author.email': 'jan-lukas.wynen@ess.eu',
        },
    ],
)
block.add_powder_calibration(cal, comment='This calibration was made up')
block.add_reduced_powder_data(reduced_data)

cif.save_cif('data/powgen_reduced.cif', block)
```


celinedurniak:
- What will be the final structure of the `cal` DataArray (with DIFA, DIFC...) ?
- Will the versioning of `essdiffraction` use the same convention as Scipp (i.e., month.year)?
- Will you provide links to the other files required for reduction, like Vanadium, sample container, empty instrument and the reduction script or Jupyter notebook...?
- Will the author name be one of the Scipp developers or the PI of the related proposal or the team taken from the proposal (PI + local contact + collaborators)?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+            Defines a mapping from keys (a.k.a. tags) to values.
+        comment:
+            Optional comment that can be written above the chunk in the file.
+        schema:
+            CIF Schema used for the chunk.
+            Content is not checked against the schema, but the schema is written
+            to the file.
+        """
+        self._pairs = dict(pairs) if pairs is not None else {}
+        self._comment = _encode_non_ascii(comment)
```
SimonHeybrock, line 155: Call setter to avoid duplication?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+            Optional comment that can be written above the loop in the file.
+        schema:
+            CIF Schema used for the loop.
+            Content is not checked against the schema, but the schema is written
+            to the file.
+        """
+        self._columns = {}
+        for key, column in columns.items():
+            self[key] = column
+        self._comment = _encode_non_ascii(comment)
```
SimonHeybrock, line 227: Use setter?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+        for key in self._columns:
+            f.write(f'_{key}\n')
+        formatted_values = [
+            tuple(map(_format_value, row)) for row in zip(*self._columns.values())
+        ]
+        sep = (
+            '\n'
+            if any(';' in item for row in formatted_values for item in row)
+            else ' '
+        )
```
SimonHeybrock, line 281: Can you explain this? Maybe also in a comment?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+            Optional comment that can be written above the block in the file.
+        schema:
+            CIF Schema used for the block.
+            Content is not checked against the schema, but the schema is written
+            to the file.
+        """
+        self._name = ''
+        self.name = name
+        self._content = _convert_input_content(content) if content is not None else []
+        self._comment = _encode_non_ascii(comment)
```
SimonHeybrock, line 324: Use setter?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+          ... )
+
+        Add to a block:
+
+          >>> from scippneutron.io import cif
+          >>> block = cif.Block('reduced-data')
+          >>> da = sc.DataArray(intensity, coords={'tof': tof})
+          >>> block.add_reduced_powder_data(da)
+        """
+        self.add(_make_reduced_powder_loop(data, comment=comment))
```
SimonHeybrock, line 426: If we have a free helper function anyway, given that this wrapper does nearly nothing, is it even worth having it?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+    data_name = _normalize_reduced_powder_name(data.name or 'intensity_net')
+
+    res = Loop({coord_name: sc.values(coord)}, comment=comment, schema=PD_SCHEMA)
+    if coord.variances is not None:
+        res[coord_name + '_su'] = sc.stddevs(coord)
+    res[data_name] = sc.values(data.data)
+    if data.variances is not None:
+        res[data_name + '_su'] = sc.stddevs(data.data)
+
+    if data.unit != 'one':
```
SimonHeybrock, line 652: This might break, can we compare to `sc.Unit('')`?

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,677 @@
...
+    return s.encode('ascii', 'backslashreplace').decode('ascii')
+
+
+def _format_value(value: Any) -> str:
+    if isinstance(value, sc.Variable):
+        if value.variance is not None:
+            without_unit = sc.scalar(value.value, variance=value.variance)
+            s = f'{without_unit:c}'
+        else:
+            s = str(value.value)
```
SimonHeybrock, line 577: I have never thought about this before, but is formatting depending on the system locale, i.e., will people in some Germanic countries (like Sweden) get a comma instead of a dot in floating-point numbers? If so, we should probably avoid this, i.e., ensure we always write with dot?

tests/io/cif_test.py
```diff
@@ -0,0 +1,920 @@
...
+                'audit.creation_date': dt,
+            }
+        ],
+    )
+    res = write_to_str(block)
+    assert (
+        res
+        == '''data_datetime
+
+_audit.creation_date 2023-12-01T15:09:45+00:00
```
SimonHeybrock, line 128: iirc timezone info was deprecated in NumPy and Scipp does not support it, i.e., everything is UTC. But here you test `datetime`, so we get the `+00:00 when formatting?

SimonHeybrock (commented):
Thorough testing 👍 

Some minor questions:

jl-wynen:
> * What will be the final structure of the `cal` DataArray (with DIFA, DIFC...) ?

As shown in the docstring:
```python
cal = sc.DataArray(
    sc.array(dims=['cal'], values=[3.4, 0.2]),
    coords={'power': sc.array(dims=['cal'], values=[0, 1])},
)
```
which corresponds to `tzero = 3.4`, `DIFC = 0.2`.
This is per data block, i.e., for one group of pixels that are focussed in the same way. I don't know yet how to handle focussing into multiple groups because that will need multiple data blocks.

>     * Will the versioning of `essdiffraction` use the same convention as Scipp (i.e., month.year)?

Yes. Are you asking because of the `computing.diffrn_reduction` field in the example? If so, we still need to decide how to encode software because there will be multiple programs that need to be listed.

>     * Will you provide links to the other files required for reduction, like Vanadium, sample container, empty instrument and the reduction script or Jupyter notebook...?

I'd very much like to. But I don't think CIF has any way of doing so. Does it?

>     * Will the author name be one of the Scipp developers or the PI of the related proposal or the team taken from the proposal (PI + local contact + collaborators)?

It will be whatever the person running the code decides. For autoreduction, we can define authorship ourselves. But otherwise, this will be up to the user. (Same as for SciCat datasets. This is one reason for a unified representation, a la #473.)

src/scippneutron/io/cif.py
jl-wynen: Done. It's about handling multi-line strings as tested in `test_write_block_single_loop_multi_line_string`:
```python
loop_
_diffrn.ambient_environment
_diffrn.id
; water
and some salt
;
123
sulfur
x6a
```

src/scippneutron/io/cif.py
jl-wynen: I only added it for convenience since we use method chaining in other interfaces. But I agree that it is not a strong argument here because most 'interesting' blocks won't be created and written in a single expression. Should I remove?

src/scippneutron/io/cif.py
jl-wynen: How so? I thought the two are equivalent?

src/scippneutron/io/cif.py
jl-wynen: I'm pretty sure this is not the case. I just changed my system's number formatting to German and I still get periods, not commas.
Doing some cursory searching, Python's `'{}'` uses periods as separators, unless specified otherwise in the braces. C++ uses the 'C locale' by default. So unless someone calls `setlocale`, we're fine. But this is global, so it's possible that another library that we load into Python does this. But even then, I'm unsure if C++'s streams respect it.

tests/io/cif_test.py
jl-wynen: The test is about support for `datetime` objects so that we don't require wrapping everything in Scipp Variables.
It is actually unfortunate that we don't print the timezone info with Scipp objects as the CIF dictionary kind of hints that they should be present.

src/scippneutron/io/cif.py
SimonHeybrock: Generally I think having fewer functions and methods is good, but if you prefer keeping it I am fine with that as well.

src/scippneutron/io/cif.py
SimonHeybrock: Suppose we change the string formatting to return something that is not `'one'`?

src/scippneutron/io/cif.py
SimonHeybrock: 👍

src/scippneutron/io/cif.py
jl-wynen: This doesn't use string formatting. The equality operator is implemented on `Unit` to parse the RHS to a unit and then compare.

SimonHeybrock: Right...

rozyczko (approved):
Code looks good and produces well structured CIF files for instrumental/experiment data.

src/scippneutron/io/cif.py
```diff
@@ -0,0 +1,654 @@
...
+  ...     tof_loop,
+  ... ])
+  >>> cif.save_cif('example.cif', block)
+
+This results in a file containing
+
+.. code-block::
+
+  #\\#CIF_1.1
+  data_example
```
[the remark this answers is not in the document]
jl-wynen: As I understand it, this depends on the dictionary, not the format. The 1.1 format already allowed periods in names: https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax (see the grammar). This is one reason why I made it so that it writes the dict version into the file.
But I can change it to use the 2.0 header. The files don't use any CIF 2 features but should be compatible with it. The question is, do the readers care or would they reject the file with this change?
Also, CIF2 allows a byte order mark, should I add it?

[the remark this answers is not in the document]
rozyczko: No, I was just a bit concerned about versioning. 1.1 seems fine and is surely more prevalent than 2.0.
Even the 2.0 page ( https://www.iucr.org/resources/cif/cif2 ) calls it an alternative format, not replacement.

Let's stick with 1.1, since as you say, it does have support for the new syntax.

AndrewSazonov:
Everything looks good, at least for the moment. When we have more TOF support in EasyDiffraction, we can check if something needs to be extended or changed. A couple of comments below:

> > * What will be the final structure of the `cal` DataArray (with DIFA, DIFC...) ?
> 
> As shown in the docstring:
> 
> ```python
> cal = sc.DataArray(
>     sc.array(dims=['cal'], values=[3.4, 0.2]),
>     coords={'power': sc.array(dims=['cal'], values=[0, 1])},
> )
> ```
> 
> which corresponds to `tzero = 3.4`, `DIFC = 0.2`. This is per data block, i.e., for one group of pixels that are focussed in the same way. I don't know yet how to handle focussing into multiple groups because that will need multiple data blocks.

The simplest way is to use only one diffraction measurement in a data block. Otherwise, the `diffractogram_id` label can be used to identify the diffraction measurement to which the data presented in the `PD_DATA` (`PD_CALC`, `PD_MEAS`, `PD_PROC`) category belong to. The same label is also available in the `PD_CALIB_D_TO_TOF` category to identify the diffractogram to which the calibration relates.

> > ```
> > * Will you provide links to the other files required for reduction, like Vanadium, sample container, empty instrument and the reduction script or Jupyter notebook...?
> > ```
> 
> 
> I'd very much like to. But I don't think CIF has any way of doing so. Does it?

I haven't found any built-in ways to do this in CIF. Perhaps we could create a loop with custom keys for the file description and urls to SciCat or GitHub?

jl-wynen:
Thanks for you comments, @AndrewSazonov! We should probably revisit these questions when you have a reader in easyDiffraction. Because we can then try out different solutions so see what works.

PR #485, opened 2024-01-24
Update copier and dependencies

jl-wynen:


docs/_typehints.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any, NewType, Optional
```
nvaytet, line 1: Can you explain what the code in this file is doing? How differently do things render in the docs when this is used?

.buildconfig/ci-linux-39.yml
```diff
@@ -13,28 +13,16 @@ dependencies:
...
   - pythreejs==2.4.2
-  - scipp==23.08.0
-  - scippnexus==23.11.0
-  - scipy==1.11.3
+  - scipp==23.12.0
+  - scippnexus==23.12.1
+  - scipy==1.12.0
   - tox==4.11.3
-
-  # docs
```
nvaytet, line 30: Why were the docs deps removed? Is it because we no longer want to build the docs on py39?

tox.ini
```diff
@@ -9,16 +9,14 @@ setenv =
...
 
 [testenv:unpinned]
 description = Test with unpinned dependencies, as a user would install now.
 deps =
+  -r requirements/basetest.txt
   scippneutron
-  pytest
 commands = pytest
 
 [testenv:static]
```
nvaytet, line 22: **Not related to this line**

Are we running too many checks? There seems to be a mix of ubuntu versions, and python versions? Do we still have old CI workflows running alongside new ones?
Screenshot at 2024-01-24 13-14-03
Screenshot at 2024-01-24 13-14-51

docs/_typehints.py
jl-wynen: See https://github.com/scipp/copier_template/pull/124

.buildconfig/ci-linux-39.yml
jl-wynen: We can't build on 3.9 because there is no compatible mantid version. So I simplified the file

tox.ini
jl-wynen: True. We have both `ci.yml` and `pr_and_main.yml`. Must have been a messed up merge at some point.

SimonHeybrock: I thought this was intentional, to have builds with Mantid (using conda) as well as the common setup (no Mantid).

.github/workflows/pr_and_main.yml
```diff
@@ -24,7 +24,6 @@ jobs:
         - {os: ubuntu-20.04, python-version: "3.10", build-config: ci-linux}
```
nvaytet, line 24:
```suggestion
        - {os: ubuntu-22.04, python-version: "3.10", build-config: ci-linux}
```
It's `22.04` everywhere else.

nvaytet (approved):

PR #487, opened 2024-01-26
Remove `__init__.py` files from tests

jl-wynen:
Part of https://github.com/scipp/scipp/issues/3344


tests/mantid_test.py
```diff
@@ -14,7 +14,15 @@
 import scippneutron as scn
 from scippneutron._utils import get_attrs
 
-from .mantid_helper import mantid_is_available
+
+def mantid_is_available():
```
nvaytet, line 18: Can we use `pytest`'s `importorskip` here? We've used it in other projects, e.g. [plopp](https://github.com/scipp/plopp/blob/main/tests/backends/plotly/plotly_figure_test.py#L11)

tests/mantid_test.py
```diff
@@ -14,7 +14,15 @@
...
+    try:
+        import mantid  # noqa: F401
+
+        return True
+    except ModuleNotFoundError:
+        return False
+
```
SimonHeybrock, line 25: `pytest.importorskip` might simplify all this?

nvaytet: too slow ;-)

nvaytet (approved):

PR #496, opened 2024-02-19
deps: add matplotlib

### LINKED ISSUE #495: Fix ScippNeutron unpinned tests
See https://github.com/scipp/scippneutron/actions/workflows/unpinned.yml.

jokasimr:
Fixes #495

The alternative solution would be to treat matplotlib as an optional dependency, and only run the tests that use it if it is installed.

### DIFF
--- a/requirements/basetest.in
+++ b/requirements/basetest.in
@@ -2,6 +2,7 @@
 # Do not make an environment from this file, use test.txt instead!
 
 hypothesis
+matplotlib
 psutil
 pytest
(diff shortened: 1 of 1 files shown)


jl-wynen:
> The alternative solution would be to treat matplotlib as an optional dependency, and only run the tests that use it if it is installed.

It should be optional just like in scipp. Can you instead add it in `basetest.in`?

jl-wynen (approved):

PR #502, opened 2024-03-18
fix: json nexus loader

### LINKED ISSUE #501: JSONGroup/JSONDataset not compatible with the latest json nexus template.
It is not compatible with the latest json nexus templates.

For example,
https://github.com/scipp/scippneutron/blob/0988a279a8b2fc480bca867bdc4a06233d0e4e2b/tests/io/json_nexus_examples/entry.json#L1-L24

Now looks like:
```json
{
  "children": [{
      "name": "entry",
      "type": "group",
      "attributes": [{ "name": "NX_class", "dtype": "string", "values": "NXentry"}],
      "children": [{
          "module": "dataset",
          "config": { "name": "title", "values": "$TITLE$", "type": "string"}
        }]
    }]
}
```

We should make it compatible and update the examples.

jokasimr:
Fixes #501

### DIFF
--- a/src/scippneutron/io/nexus/_json_nexus.py
+++ b/src/scippneutron/io/nexus/_json_nexus.py
@@ -8,7 +8,6 @@ from dataclasses import dataclass
 from typing import Any, Dict, List, Tuple, Union
 
 import numpy as np
-from scippnexus.typing import H5Group
 
 _nexus_class = "NX_class"
 _nexus_units = "units"
@@ -16,6 +15,7 @@ _nexus_name = "name"
 _nexus_path = "path"
 _nexus_values = "values"
 _nexus_dataset = "dataset"
+_nexus_config = "config"
 _nexus_group = "group"
 _nexus_children = "children"
 _nexus_link = "link"
@@ -87,12 +87,13 @@ def make_json_dataset(name: str, data) -> dict:
             "size": data.shape,
             "type": numpy_to_filewriter_type[data.dtype.type],
         }
-
     return {
-        "type": "dataset",
-        "name": name,
-        "values": data,
-        "dataset": dataset_info,
+        'module': _nexus_dataset,
+        _nexus_config: {
+            **dataset_info,
+            _nexus_name: name,
+            _nexus_values: data,
+        },
         "attributes": [],
     }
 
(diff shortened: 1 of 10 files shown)


src/scippneutron/io/nexus/_json_nexus.py
```diff
@@ -371,12 +320,10 @@ def __iter__(self):
 
     def visititems(self, callable):
         def skip(node):
-            return node['type'] == _nexus_link or contains_stream(self)
+            return _is_link(node) or contains_stream(self)
```
SimonHeybrock, line 323: Not related to your change, but is this (skipping links) actually what h5py's visititems does?

src/scippneutron/io/nexus/_json_nexus.py
```diff
@@ -117,100 +118,43 @@ def _get_attribute_value(
...
-        self.parent = parent
-        self.name = name
-        self.file = file
+def _visitnodes(root: Dict):
+    for child in root.get(_nexus_children, ()):
+        yield child
+        yield from _visitnodes(child)
 
 
-def _visit_nodes(
```
SimonHeybrock, line 128: I see you did quite a bit of refactoring/cleanup below. Was it not working, overly complicated, or just unused?

SimonHeybrock (commented):
Looks good overall. 👍 for adding more tests. A few questions:

src/scippneutron/io/nexus/_json_nexus.py
jokasimr: It seemed overly complicated and it was only used in one place so I thought it was safe to refactor.

src/scippneutron/io/nexus/_json_nexus.py
jokasimr: Yes that seems correct
```python
import h5py
with h5py.File('testfile.hdf5','w') as f:
    group = f.create_group("group")
    ds = group.create_dataset("test", (5, 5))
    f["soft"] = h5py.SoftLink('/group/test')
    f["hard"] = h5py.SoftLink('/group/test')
    f.visititems(print)
```
prints
```
group <HDF5 group "/group" (1 members)>
group/test <HDF5 dataset "test": shape (5, 5), type "<f4">
```

SimonHeybrock (approved):

jokasimr:
I think there are some things left to do here:
I added a test for loading the ymir instrument description from gitlab, this is what we will need to load in beamlime (is that right?), but the test does not pass yet.
The test was marked as skipped because I thought the test failure was related to the other skipped test with the skip message: `Stream handling with log not implemented`, but now I'm not sure that it is.

The test fails with the message:
```
UserWarning: Failed to load /entry/instrument/laser_monitor as NXmonitor: Signal is not an array-like.
```
not sure what the issue is there, but it seems like either
 * something about the `laser_monitor` group doesn't conform to our idea of a `NXmonitor`
 * or something about the way the structure is handled by the json loader is not correct and that causes the error.
 
Do you have an idea about what the issue might be @SimonHeybrock ?

SimonHeybrock:
> I think there might be one thing left to fix here: I added a test for loading the ymir instrument description from gitlab, this is what we will need to load in beamlime (is that right?),

I think we will mainly be using Loki.

> but the test does not pass yet. The test was initially marked as skipped because I thought the test failure was related to the other skipped test with the skip message: `Stream handling with log not implemented`, but I'm not sure that it is.
> 
> The test fails with the message:
> 
> ```
> UserWarning: Failed to load /entry/instrument/laser_monitor as NXmonitor: Signal is not an array-like.
> ```
> 
> not sure what the issue is there, but it seems like either
> 
>     * something about the `laser_monitor` group doesn't conform to our idea of a `NXmonitor`
> 
>     * or something about the way the structure is handled by the json loader is not correct and that causes the error.
> 
> 
> Do you have an idea about what the issue might be @SimonHeybrock ?

This sounds like a problem in either ScippNexus or the file, not the JSON adapter.

SimonHeybrock:
> > I think we will mainly be using Loki.
> 
> I didn't mean the `ymir.json` file specifically, just that we will need to read instrument descriptions like those in the gitlab repo. For the record, loading the `loki.json` file fails with a similar error message:
> 
> ```
> UserWarning: Failed to load /entry/instrument/loki_detector_0 as NXdetector: Signal is not an array-like.
> ```
> 
> > This sounds like a problem in either ScippNexus or the file, not the JSON adapter.
> 
> Okay then I'll mark the test as TODO and leave it as it is right now.

Actually you simply need to ignore the warning. The loading actually works, it is just a fallback that issues a warning. Since we have strict warning policies (tests fail if there are warnings), you have to allow this warning in this test.

jokasimr:
> Actually you simply need to ignore the warning. 

:+1:

tests/io/load_nexus_json_test.py
```diff
@@ -200,3 +200,25 @@ def test_nexus_json_load_log_utf8_unit(examples):
         },
     )
     sc.testing.assert_identical(loaded, expected)
+
+
+# TODO: remove warning filter
+# Loading a NXmonitor fails with
+# 'Signal is not array like' error and issues a warning.
+# But this should probably be fixed elsewhere.
```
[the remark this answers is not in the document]
jokasimr: Should we create an issue for figuring this out with ECDC?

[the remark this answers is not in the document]
SimonHeybrock: If I remember correctly the issue was that the value datasets did not exist (since there were no values). If they would instead use an empty dataset it would work fine. But I am not sure what you see is the same issue. What I mention was something I observed in NeXus files, so this may be different.

PR #504, opened 2024-04-17
Peak fitting and removal

jl-wynen:
This implements a generic peak-removal routine that can be used for https://github.com/scipp/essdiffraction/issues/24

For now, I could only test this with POWGEN data as all other test data I could get either contains no peaks or peaks with a strange shape. I talked with @celinedurniak about test data fro DREAM. But it seems unclear at this point how best to test it.

I am opening this PR so that the code can be reviewed as it should not change significantly for new data.


src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+) -> list[FitResult]:
+    background = _parse_model_spec(background, prefix='bkg_')
+    peak = _parse_model_spec(peak, prefix='peak_')
+    constraints = FitConstraints(
+        min_p_value=min_p_value,
+        max_peak_width_factor=max_peak_width_factor,
+        min_peak_width_factor=min_peak_width_factor,
+        guess_background_range=guess_background_range,
+        neighbor_separation_factor=neighbor_separation_factor,
+    )
```
SimonHeybrock, line 148: Seems like `constraints: FitConstraints` might make sense as a direct argument to `fit_peaks`, instead of forwarding all the args and constructing one here?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+    for i in range(windows.sizes[peak_estimates.dim]):
+        window = windows[peak_estimates.dim, i]
+        data_in_window = data[data.dim, window[0] : window[1]]
+        results.append(
+            _fit_peak_sc(data_in_window, window, background, peak, constraints)
+        )
+    return results
+
+
+def _fit_peak_sc(
```
SimonHeybrock, line 171: What is the `sc` suffix?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+    return (fwhm < constraints.min_peak_width_factor * bin_width).value
+
+
+def _guess_background(
+    data: sc.DataArray, model: Model, constraints: FitConstraints
+) -> dict[str, sc.Variable]:
+    # 2* because the range is split between beginning and end of window
+    n = len(data) // (2 * constraints.guess_background_range)
+    tails = sc.concat([data[:n], data[-n:]], dim=data.dim)
+    return model.guess(tails)
```
SimonHeybrock, line 382: So this is assuming that the peak is somewhere in the middle?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+        right_neighbor - left_neighbor
+    ) * constraints.neighbor_separation_factor
+    lo = left_neighbor + min_separation
+    hi = right_neighbor - min_separation
+    # Do not adjust the left edge of the first window and the right edge of the
+    # last window because there are no neighbors on those sides.
+    left_edge = windows['range', 0][1:]
+    right_edge = windows['range', 1][:-1]
+    left_edge[:] = sc.where(left_edge < lo, lo, left_edge)
+    right_edge[:] = sc.where(right_edge > hi, hi, right_edge)
```
SimonHeybrock, line 444: Is the underlying assumption in the functions above that the distances between peaks are large enough such that the windows never overlap?

src/scippneutron/peaks/model.py
```diff
@@ -0,0 +1,556 @@
...
+    def __add__(self, other: Model) -> CompositeModel:
+        """Combine two models into a :class:`CompositeModel`."""
+        if not isinstance(other, Model):
+            return NotImplemented
+        return CompositeModel(left=self, right=other, prefix='')
+
+    def with_prefix(self, prefix: str) -> Model:
+        """Return a copy of the model with a new prefix."""
+        model = deepcopy(self)
+        model._prefix = prefix
```
SimonHeybrock, line 197: Does this ever need to be chained, i.e., we should combine with the existing prefix?

src/scippneutron/peaks/model.py
```diff
@@ -0,0 +1,556 @@
...
+            x, **{name: params[name] for name in self._right.param_names}
+        )
+        return left + right
+
+    def _guess(self, x: sc.Variable, y: sc.Variable) -> dict[str, sc.Variable]:
+        data = sc.DataArray(y, coords={y.dim: x})
+        return {
+            **self._left.guess(data),
+            **self._right.guess(data),
+        }
```
SimonHeybrock, line 261: Assumes that left and right do not have overlapping params?

src/scippneutron/peaks/model.py
```diff
@@ -0,0 +1,556 @@
...
+        )
+
+    @property
+    def degree(self) -> int:
+        """The degree of the polynomial."""
+        return len(self._param_names) - 1
+
+    def _call(self, x: sc.Variable, params: dict[str, sc.Variable]) -> sc.Variable:
+        # The loop is rather complex to ensure that `val` has the correct shape after
+        # broadcasting (assuming that all params have the same shape).
```
SimonHeybrock, line 303: So this is for the case of non-0-D params?

SimonHeybrock (commented):
Nice work!

src/scippneutron/peaks/_remove_peaks.py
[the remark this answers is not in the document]
jl-wynen: There aren't any because I was hoping I could get better test data.

tests/peaks/fit_peaks_test.py
```diff
@@ -0,0 +1,204 @@
...
+            windows=sc.scalar(1.0),
+            background='linear',
+            peak='gaussian',
+        )
+
+
+def test_fit_peaks_rejects_disordered_coord():
+    rng = np.random.default_rng(93)
+    data = sc.DataArray(sc.ones(sizes={'x': 10}), coords={'x': sc.arange('x', 10.0)})
+    data = data[list(rng.permutation(len(data)))]
```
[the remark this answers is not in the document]
jl-wynen: No, because the rng has a fixed seed and I checked that for this seed, it is fine.

The only risk would be that numpy change their default RNG.

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+            data, **{name: sc.values(val) for name, val in self.popt.items()}
+        )
+
+    def eval_peak(self, data: sc.Variable) -> sc.Variable:
+        return self.peak(
+            data,
+            **{
+                name: sc.values(val)
+                for name, val in self.popt.items()
+                if name.startswith('peak_')
```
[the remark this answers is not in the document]
jl-wynen: The way the models are constructed, they are guaranteed to use the `peak_` and `bkg_` prefixes.
Ultimately, I picked copied this solution from lmfit and didn't spend time to come up with a better one. Do you think there is a need for it?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,496 @@
...
+    peak_points_down = enum.auto()
+    p_too_small = enum.auto()
+    window_too_narrow = enum.auto()
+
+    @property
+    def success(self) -> bool:
+        return self in (
+            FitAssessment.accept,
+            FitAssessment.candidate,
+        )
```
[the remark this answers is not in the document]
jl-wynen: It means that the fit is good enough to be used.

Initially, my idea was to allow `assess_fit` to choose whether a fit is so good that it should be taken as the result as is (`accept`), or that a fit is ok but it should try a different model to maybe get a better result (`candidate`). I did not actually implement this because I don't know how make that judgement. So I can probably remove `candidate` and define `success == accept`. Opinions?

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: leftover from when there was a `_fit_peak` that used lmfit. Going to remove

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: Yes. It's not actually true for all initial estimates and window sizes. But it worked well so far.

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: No. The windows do overlap a lot. The assumption is that each window only contains one peak which is why the windows are moved away from peak estimates.

src/scippneutron/peaks/model.py
jl-wynen: That would make it harder to identify peak vs background params.
I don't currently see a use for it because `fit_peaks` enforces specific prefixes for its models. Maybe this is another indication that this approach is too brittle

src/scippneutron/peaks/model.py
jl-wynen: This is enforced in `__init__`

src/scippneutron/peaks/model.py
jl-wynen: Yes. Which would probably cause problems elsewhere, now that I think about it. In fact, `curve_fit` only supports 0d params. So I'll simplify

src/scippneutron/peaks/_fit_peaks.py
SimonHeybrock: Is it really a significant convenience difference? Because the user would have to import `FitConstraints`?

src/scippneutron/peaks/_fit_peaks.py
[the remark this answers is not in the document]
SimonHeybrock: 🤷

src/scippneutron/peaks/_fit_peaks.py
SimonHeybrock: Maybe add a comment?

src/scippneutron/peaks/_fit_peaks.py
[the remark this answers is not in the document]
jl-wynen: It's related to https://github.com/scipp/scippneutron/pull/504#discussion_r1570072849 The user can set any prefix they like. But they get overridden so that the name based matching works.

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: > Because the user would have to import FitConstraints?

Yes. And because they need to call its constructor. It's not a huge deal and would probably simplify the interaction between `remove_peaks` and `fit_peaks`.

SimonHeybrock: Personally I would go for the "more typing, but cleaner" solution.

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: Done in the docstring

src/scippneutron/peaks/model.py
jl-wynen: Simplified it a little

src/scippneutron/peaks/_fit_peaks.py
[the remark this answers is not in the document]
jl-wynen: Sun Young and I looked into using a nested structure of parameters instead. This would have the problem that `curve_fit` requires a flat dict of params. So the nested structure would have to be flattened when performing the fit. At that stage, we would again have to rely on having unique names. This would be ok if it were contained to the fit function. But it would also impact the interface of `CompositeModel` and thus leak out into user code.

I think that this would make the code more complicated overall. So I would stick with the current mechanism.

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,726 @@
...
+        """Format an overview of the fit result."""
+        param_name_length = max(map(len, self.popt.keys()))
+        params = '\n- '.join(
+            f'{{:{param_name_length}s}}: {val:c}'.format(key)
+            for key, val in self.popt.items()
+        )
+        return f'''FitResult: {self.message}
+In window: ({self.window[0].value}, {self.window[1].value}) {self.window.unit}
+Models:
+- background: {self.background.__class__.__name__}(prefix='{self.background.prefix}')
```
nvaytet, line 190: What are the types of these values? Are they mostly scipp variables, data arrays?
If so, can we return a DataGroup which has a html repr etc..?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,726 @@
...
+- Red. chi-squared: {self.red_chisq.value:.2e}
+- p:                {self.p_value.value:.3f}'''
+
+
+class FitAssessment(enum.Enum):
+    """Indicates whether the fit was successful or how if failed."""
+
+    success = enum.auto()
+    failed = enum.auto()
+    background_is_better = enum.auto()
```
nvaytet, line 205: I think the other names are explicit enough but I don't really understand what `background_is_better` means. Does it mean fitting the background gave a better result? Even that does not make sense...

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,726 @@
...
+        Must be a 1d variable with dimension ``data.dim``.
+    windows:
+        If a scalar, the size of fit windows.
+        A window is constructed for each peak estimate centered on the estimate
+        with a width equal to ``windows`` (adjusted to the data range and to maintain
+        a separation between peaks, see
+        :attr:`FitParameters.neighbor_separation_factor`).
+
+        If a 2d array, the windows for each peak.
+        Must have sizes ``{data.dim: len(data), 'range': 2}`` where
```
nvaytet, line 319: It's interesting that I need a similar data structure for specifying wavelength bands in SANS, which can be overlapping.
Would there be any value in making a helper class?

src/scippneutron/peaks/_fit_peaks.py
```diff
@@ -0,0 +1,726 @@
...
+            fit_parameters=fit_parameters,
+            fit_requirements=fit_requirements,
+        )
+        if result.assessment == FitAssessment.success:
+            return result
+        if candidate_result is None:
+            # Keep a potentially failed result to have something to
+            # return in case all fits fail.
+            candidate_result = result
+        # else: reject
```
nvaytet, line 397: Commented code?

src/scippneutron/peaks/model.py
```diff
@@ -0,0 +1,556 @@
...
+
+        f(x) = \\text{left}(x) + \\text{right}(x)
+
+    Composite models can be constructed by adding models, e.g.,
+
+    .. code-block:: python
+
+        left = PolynomialModel(degree=2)
+        right = GaussianModel()
+        composite = left + right
```
nvaytet, line 218: Would you always just have two models? Couldn't you have something like `CompositeModel([model1, model2, model3])` in a case where you need more than two functions?

nvaytet (commented):
Some comments. I didn't look at the tests yet.
Also, as asked on Slack, should we take a look at EasyCore for fitting? Could we avoid re-writing a lot of stuff?

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: No, there are a bunch of other types: models, string, enum, dict

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: It is described in the 'Fit procedure' section in the docs of `fit_peaks`. Does this answer your question? If not, it needs to be updated.

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: What would that class do?

What did you call the 'range' dimension?

src/scippneutron/peaks/_fit_peaks.py
jl-wynen: No. This is meant to explain what happens implicitly in the else case.

src/scippneutron/peaks/model.py
jl-wynen: Then you would need to nest composite models. E.g., `model1 + model2 + model3` would do that for you.

src/scippneutron/peaks/_fit_peaks.py
nvaytet: > What would that class do?

Not entirely sure, maybe some validation (ensuring the size of `range` is 2?), maybe some method to easily get the left/right edges as an array?

> What did you call the 'range' dimension?

`'band'` for wavelength band, but it's not a great name.

jl-wynen: > Not entirely sure, maybe some validation (ensuring the size of range is 2?), maybe some method to easily get the left/right edges as an array?

Could be useful. But really, this sounds like it would be best served by a dedicated dtype. If only we had user-defined structured dtypes...

nvaytet:
I have no further comments. Looks good as far as I'm concerned 👍

SimonHeybrock (approved):

PR #505, opened 2024-04-18
Fix instrument view

nvaytet:
After plopp 24.04 was released, the `scatter3d` function now returns a figure instead of a `VBox`.


jl-wynen (approved):

tests/io/load_nexus_test.py
```diff
@@ -247,8 +247,17 @@ def test_does_not_load_events_if_index_not_ordered(load_function: Callable):
         Detector(detector_numbers=np.array([0, 1]), event_data=event_data_1)
     )
 
-    with pytest.warns(UserWarning, match="Invalid index in NXevent_data at "):
+    # It seems that pytest changed the way multiple warnings are returned.
+    # We now record all warnings, and make sure there is at least one warning with
+    # the expected message
+    with pytest.warns(UserWarning) as record:
```
MridulS, line 253: Hmm, maybe we should be catching the multiple warnings that are emitted? The user would see them right?

Something like:
``` python
    warn_load = pytest.warns(UserWarning, match="Failed to load")
    warn_invalid_index = pytest.warns(UserWarning, match="Invalid index in NXevent_data at ")
    with warn_load, warn_invalid_index:
        load_function(builder)
```

nvaytet: The issue is that there are about 4 different warnings. In addition, `load_nexus` will be removed / deprecated from scippneutron and we will use instrument specific loaders instead, which is why I didn't put more effort into this.

MridulS: Ah makes sense!

MridulS (approved):

PR #506, opened 2024-04-19
Harmonize CI with other projects

nvaytet:
Instead of having two `.yml` workflows, run Mantid tests in a separate job but same workflow, and remove `pr_and_main.yml` workflow.


nvaytet:
> Could this lead to harder-to-resolve conflicts when we update the template? I don't know how it behaves when a chunk is added in the middle of a file.

I just tried a copier update from this branch and updating the `ci.yml` file was fine.

nvaytet:
I'm not sure if we want to keep the `.buildconfig` files for the other platforms around? I guess they may be useful?
It's just that we are not keeping them up to date.

Maybe if we do https://github.com/scipp/copier_template/issues/158, we could (partly?) generate the buildconfig files from the `ptproject.toml`?

jl-wynen:
> not sure if we want to keep the `.buildconfig` files for the other platforms around? I guess they may be useful? It's just that we are not keeping them up to date.
> 
> Maybe if we do [scipp/copier_template#158](https://github.com/scipp/copier_template/issues/158), we could (partly?) generate the buildconfig files from the `ptproject.toml`?

Maybe we can generate it. But our philosophy in other packages is to only test on Linux. Since we don't interact with Mantid through C++, this should be fine here, too. So I think the files can be removed.

.github/workflows/pr_and_main.yml
```diff
@@ -1,44 +0,0 @@
...
-    defaults:
-      run:
-        shell: bash -l {0}  # required for conda env
-    strategy:
-      fail-fast: false
-      matrix:
-        variant:
-        - {os: ubuntu-22.04, python-version: "3.10", build-config: ci-linux}
-        - {os: macos-11, python-version: "3.10", build-config: ci-macos}
-        - {os: windows-2019, python-version: "3.10", build-config: ci-windows}
```
YooSunYoung, line 26: Do we want to keep this matrix in the `ci.yml` as well? Cuz the default one from the template only has ubuntu.

nvaytet: No I think we want to only test on ubuntu. I think testing on 3 platforms is a remnant from the days when scippneutron also had a C++ component?

jl-wynen: Right now, we should be fine without. This may change if/when we use compiled components (through numba or otherwise).

jl-wynen (approved):
Just needs a change to the branch protection rules.

PR #514, opened 2024-05-23
Two theta gravity

jl-wynen:
Fixes https://github.com/scipp/esssans/issues/90

~This is a draft because it does not include the reflectometry version of this function yet. But please review so I know whether the general approach is good.~

Reflectometry code postponed to later work.

~This version does not expose a way of computing unit vectors of the 'beam aligned coord system'. This would be needed to compute `phi` without gravity in SANS. But I don't know how to even define the coord system in that case. And the current implementation in ESSsans is not entirely clean because it has functions in the graph that depend on gravity.~

The basis vectors are now available. Projecting `scattered_beam` onto them not currently.


nvaytet:
> And the current implementation in ESSsans is not entirely clean because it has functions in the graph that depend on gravity.

I think we need gravity because we need to know which way is pointing up/down. Whether we include a correction for gravity in the computation of `two_theta` is separate, it depends on whether the function that computes `two_theta` takes gravity as an input or not?

jl-wynen:
> I think we need gravity because we need to know which way is pointing up/down.

True.

> Whether we include a correction for gravity in the computation of `two_theta` is separate, it depends on whether the function that computes `two_theta` takes gravity as an input or not?

Technically, yes. This differs a little from other cases, though. Consider your proposal for ESSdiffraction where the presence of a `TwoThetaBins` param triggers focussing into two theta bins. So it seems that the presence of a gravity parameter would trigger gravity correction.

But we also need gravity to define a coordinate system. I currently think that it would be better to explicitly provide the basis vectors or at least an `up` vector instead of gravity.

src/scippneutron/conversion/beamline.py
```diff
@@ -217,19 +298,157 @@ def two_theta(
...
+
+        y' = y + \frac{|g| m_n^2}{2 h^2} L_2^2 \lambda^2
+
+    Where :math:`|g|` is the strength of gravity, :math:`m_n` is the neutron mass,
+    :math:`h` is the Planck constant, and :math:`\lambda` is the wavelength.
+    This gives the gravity-corrected scattering angles:
+
+    .. math::
+
+        \mathsf{tan}(2\theta) &= \frac{\sqrt{x^2 + y'^2}}{z} \\
```
nvaytet, line 413: I think the math checks out but I was quite confused by your drawing of the coordinate system. I can't really understand the Phi angle and the dashed lines. In my head, it looks more like this
IMG_0840

tests/conversion/beamline_conversions_test.py
```diff
@@ -118,6 +138,410 @@ def test_two_theta_orthogonal_beams():
...
+
+    sc.testing.assert_allclose(
+        res['two_theta'], sc.scalar(0.0, unit='rad'), atol=sc.scalar(1e-15, unit='rad')
+    )
+    # When polar = 0, the azimuthal angle is ill-defined, so there is no test.
+
+
+def test_scattering_angles_with_gravity_drops_in_expected_direction():
+    wavelength = sc.scalar(1.6, unit='Å')
+    gravity = sc.vector([0.0, -9.81, 0.0], unit='m/s^2')
```
nvaytet, line 310:
```suggestion
    gravity = sc.vector([0.0, sc.constants.g.value, 0.0], unit=sc.constants.g.unit)
```

tests/conversion/beamline_conversions_test.py
```diff
@@ -118,6 +138,410 @@ def test_two_theta_orthogonal_beams():
...
+    assert sc.all(with_gravity['two_theta'][1] < without_gravity[1]).value
+    sc.testing.assert_allclose(
+        with_gravity['phi'],
+        sc.array(dims=['det'], values=[np.pi / 2, -np.pi / 2], unit='rad'),
+    )
+
+
+def test_scattering_angles_with_gravity_beams_aligned_with_lab_coords():
+    wavelength = sc.array(dims=['wavelength'], values=[1.6, 0.9, 0.7], unit='Å')
+    gravity = sc.vector([0.0, -9.81, 0.0], unit='m/s^2')
```
nvaytet, line 341:
```suggestion
    gravity = sc.vector([0.0, sc.constants.g.value, 0.0], unit=sc.constants.g.unit)
```

src/scippneutron/conversion/beamline.py
jl-wynen: This is another possible representation. I tried to show that phi is defined in the x-y plane as in your graphic on the right.

I will try to improve it based on your suggestion.

tests/conversion/beamline_conversions_test.py
jl-wynen: Can do but doesn't ultimately impact the test.

jl-wynen:
Do you have any comments about performance, especially memory usage? Compared to the implementation in ESSsans.

src/scippneutron/conversion/beamline.py
nvaytet: hmm I think I now see what you tried to do. Maybe ask someone else before you change it?
I don't know if adding a ring or disk in the plane normal to z with radius `sqrt(x^2 + y^2)` on your figure would help?

nvaytet:
> Do you have any comments about performance, especially memory usage? Compared to the implementation in ESSsans.

Isn't it trying to do the same (or pretty close to it?).
One thing is that in `essans` we still need to define `cylindrical_x` and `cylindrical_y` for the beam center finder. How would the changes here fit in with that?

src/scippneutron/conversion/beamline.py
jl-wynen: I don't want to make the figure even busier. So splitting it into two might be a good idea.
I did a review with Johannes and Sunyoung, and they suggested to split it up. And independently, I had Mridul compare your figure to mine and he seemed to understand yours more easily.

jl-wynen:
> Isn't it trying to do the same (or pretty close to it?).

Close. It differs in the definition for backscattering. The implementation here matches the definition in the regular `two_theta` function. This is why my implementation ultimately calls `atan2` instead of `asin`.

> One thing is that in essans we still need to define cylindrical_x and cylindrical_y for the beam center finder. How would the changes here fit in with that?

This could be done by adding a function / functions to compute those coordinates. They could then be inputs to `scattering_angles_with_gravity`. I did the computation inline here so that `x` can be dropped early to reduce memory usage.

Alternatively, we can provide a function to compute the basis vectors. (That is, make `_beam_aligned_unit_vectors` public.) The beam centre finder would then have to project `scattered_beam` onto those vectors itself and thus repeat that computation. But that is already the case in ESSsans.

Speaking of this, is the check in https://github.com/scipp/scippneutron/blob/e8319b850d4db4db9750cb12a18bd2fd7108faaf/src/scippneutron/conversion/beamline.py#L339-L347 too strict for the beam centre finder? Or for other applications? ESSsans simply assumed that gravity is orthogonal to the incident beam.

jl-wynen:
@nvaytet This should be ready now. I updated the image and made `beam_aligned_unit_vectors` public. Should the computation of `phi` without gravity in ESSsans also be moved here?

nvaytet:
> is the check too strict for the beam centre finder? Or for other applications? ESSsans simply assumed that gravity is orthogonal to the incident beam.

I guess it depends when it would be enforced. For example, in relfectometry, the beam would not necessarily be horizontal for some instruments. For the beam center finder, you do get shifts that are of several millimeters, so the check could be too strict. We should try it.

nvaytet:
> Should the computation of phi without gravity in ESSsans also be moved here?

Could do, but leave for another PR?

nvaytet:
> I updated the image

Figures look good :+1:

jl-wynen:
> > is the check too strict for the beam centre finder? Or for other applications? ESSsans simply assumed that gravity is orthogonal to the incident beam.
> 
> I guess it depends when it would be enforced. For example, in relfectometry, the beam would not necessarily be horizontal for some instruments. For the beam center finder, you do get shifts that are of several millimeters, so the check could be too strict. We should try it.

Ok. I will make a banch in ESSsans that uses the new code here to try it out.

jl-wynen:
It works in ESSsans: https://github.com/scipp/esssans/pull/143

nvaytet:
> It works in ESSsans: [scipp/esssans#143](https://github.com/scipp/esssans/pull/143)

By "works", did you check that the old and new results are identical?

jl-wynen:
### The problem

There is indeed a discrepancy in the result. For example, take `compute_Q` of the SANS2d workflow. It produces `CleanQ[SampleRun, Numerator]`. Modifying `compute_Q` to convert to `two_theta` instead of `Q` and histogramming the result in `two_theta`, I get the following. The plot shows `sc.values((new-old)/old)` where `new` is the proposed code and `old` is the current main of ESSsans.
two_theta_discrepancy

Looking at the `two_theta` coordinate in isolation, I get a max difference of ~1e-5 per spectrum-wavelength bin.

### The cuprit

The difference stems from using `atan2` in https://github.com/scipp/scippneutron/blob/33398a1f9d0280babfae3d33fc781d17d45fcb07/src/scippneutron/conversion/beamline.py#L516 where the implementation on ESSmain uses `asin`: https://github.com/scipp/esssans/blob/4d51289fe1205c0f7b113a52a944cccd61020ecd/src/ess/sans/conversions.py#L113
More concretely:
```python
# proposed
two_theta = sc.atan2(y=sc.sqrt(x**2 + y**2), x=z)
# ESSsans main
two_theta = sc.asin(sc.sqrt(x**2 + y**2) / L2)
```
Using the `asin`-based code in the PR makes the SANS2D results (final and intermediate) agree perfectly with the implementation on ESSsans main.

I don't currently understand why this makes such a large difference. Naively, I would expect `atan2` to have better precision than `asin`. But I can't tell whether this is the case here without further investigation.

docs/user-guide/auxiliary/gravity_approx/README.md
```diff
@@ -0,0 +1,8 @@
+# Approximation of L_2 for gravity correction
+
+The file `gravity_approx.typ` contains a derivation and description of the approximation used in `scattering_angles_with_gravity`.
```
nvaytet, line 3: Why did you decide to go with this instead of using a notebook, running it and including it in the docs? Isn't it useful to have the figure in the docs, where we can just point people to it / share it (and link from the docstring of `scattering_angles_with_gravity`)?

docs/user-guide/coordinate-transformations.ipynb
```diff
@@ -27,7 +27,7 @@
     "\n",
```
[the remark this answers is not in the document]
jl-wynen: Yes! I wasn't even aware of this section.

jl-wynen:
@nvaytet Can you take another look?

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "\n",
+    "## Derivation and approximation\n",
+    "\n",
+    "After scattering, the neutron travels along curve\n",
+    "$$\n",
+    "C: \\vec{c}(t) = \\begin{pmatrix}v_x t \\\\ -\\frac{g}{2}t^2 + v_y t \\\\ v_z t\\end{pmatrix}.\n",
+    "$$\n",
+    "with initial velocity\n",
+    "$$\n",
+    "\\vec{v} = v \\begin{pmatrix}\\cos(\\phi)\\sin(2\\theta) \\\\ \\sin(\\phi)\\sin(2\\theta) \\\\ \\cos(2\\theta)\\end{pmatrix}, \\quad v = \\frac{h}{m_n \\lambda}.\n",
```
nvaytet, line 24:
```suggestion
    "\\vec{v} = v \\begin{pmatrix}\\cos(\\phi)\\sin(2\\theta) \\\\ \\sin(\\phi)\\sin(2\\theta) \\\\ \\cos(2\\theta)\\end{pmatrix}, \\quad v = \\frac{h}{m_{\rm n} \\lambda}.\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "$$\n",
+    "For a detector located at $z_d = v_z t_d$, this means that the neutron is detected at\n",
```
nvaytet, line 26:
```suggestion
    "For a detector located at $z_{\rm d} = v_{\rm z} t_{\rm d}$, this means that the neutron is detected at\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "$$\n",
+    "\\vec{p} = \\begin{pmatrix}x_d \\\\ y_d \\\\ z_d\\end{pmatrix} = \\begin{pmatrix}v_x t_d \\\\ -\\frac{g}{2}t_d^2 + v_y t_d \\\\ v_z t_d\\end{pmatrix}.\n",
```
nvaytet, line 28:
```suggestion
    "\\vec{p} = \\begin{pmatrix}x_{\rm d} \\\\ y_{\rm d} \\\\ z_{\rm d}\\end{pmatrix} = \\begin{pmatrix}v_{\rm x} t_{\rm d} \\\\ -\\frac{g}{2}t_{\rm d}^2 + v_{\rm y} t_{\rm d} \\\\ v_{\rm z} t_{\rm d}\\end{pmatrix}.\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "$$\n",
+    "To determine the scattering angles, we can consider the point where the neutron would be detected without gravity:\n",
+    "$$\n",
+    "\\vec{p}' = \\begin{pmatrix}x_d \\\\ y'_d \\\\ z_d\\end{pmatrix} = \\begin{pmatrix}v_x t_d \\\\ v_y t_d \\\\ v_z t_d\\end{pmatrix} = v t_d \\begin{pmatrix}\\cos(\\phi)\\sin(2\\theta) \\\\ \\sin(\\phi)\\sin(2\\theta) \\\\ \\cos(2\\theta)\\end{pmatrix}\n",
```
nvaytet, line 32:
```suggestion
    "\\vec{p}' = \\begin{pmatrix}x_{\rm d} \\\\ y'_{\rm d} \\\\ z_{\rm d}\\end{pmatrix} = \\begin{pmatrix}v_{\rm x} t_{\rm d} \\\\ v_{\rm y} t_{\rm d} \\\\ v_{\rm z} t_{\rm d}\\end{pmatrix} = v t_{\rm d} \\begin{pmatrix}\\cos(\\phi)\\sin(2\\theta) \\\\ \\sin(\\phi)\\sin(2\\theta) \\\\ \\cos(2\\theta)\\end{pmatrix}\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "$$\n",
+    "Thus\n",
+    "$$\n",
+    "\\tan(2\\theta) = \\frac{\\sqrt{x_d^2 + y_d^{\\prime\\,2}}}{z_d}, \\quad y_d' = y_d + \\frac{g}{2} t_d^2 = y_d + \\frac{g m_n^2 \\lambda^2}{2 h^2} L_2^{\\prime\\,2}\n",
```
nvaytet, line 36:
```suggestion
    "\\tan(2\\theta) = \\frac{\\sqrt{x_{\rm d}^2 + y_{\rm d}^{\\prime\\,2}}}{z_{\rm d}}, \\quad y_{\rm d}' = y_{\rm d} + \\frac{g}{2} t_{\rm d}^2 = y_{\rm d} + \\frac{g m_{\rm n}^2 \\lambda^2}{2 h^2} L_2^{\\prime\\,2}\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "$$\n",
+    "with\n",
+    "$$\n",
+    "L'_2 = \\frac{z_d}{\\cos(2\\theta)}\n",
```
nvaytet, line 40:
```suggestion
    "L'_2 = \\frac{z_{\rm d}}{\\cos(2\\theta)}\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "This equation is also an approximation.\n",
+    "But it would be exact if it used $L'_2$ in the denominator instead of $L_2$.\n",
+    "\n",
+    "With SANS2D data, we found a relative difference of $~10^{-5}$ in the resulting $2 \\theta$.\n",
+    "This amounted to a relative difference of $~10^{-4}$ in $I(Q)$ which is within statistical uncertainties.\n",
+    "However, LOKI is rather long and allows for measuring relatively large angles.\n",
+    "So it is a priori not clear that the approximation holds.\n",
+    "\n",
+    "To test it, we\n",
+    "1. Pick a set of values of $\\lambda$, $z_d$, $2 \\theta$, $\\phi$.\n",
```
nvaytet, line 60: Add a blank line before 1. so that the itemized list renders in sphinx?

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "2. Compute $\\vec{p}$ from those.\n",
+    "3. Compute $2 \\theta$ using $\\vec{p}$ and the equation for $\\tan(2\\theta)$.\n",
+    "\n",
+    "The code below outputs a figure that shows the relative difference $(\\text{true\\_two\\_theta} - \\text{approx\\_two\\_theta}) / \\text{true\\_two\\_theta}$ for $\\phi = 90 \\degree$.\n",
```
nvaytet, line 64: There is a parsing error somewhere in the math expression here, it renders as
Screenshot at 2024-06-21 10-32-47

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
nvaytet, line 60:
```suggestion
    "1. Pick a set of values of $\\lambda$, $z_{\rm d}$, $2 \\theta$, $\\phi$.\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "true\\_two\\_theta is the input value for $2 \\theta$ and approx\\_two\\_theta is the result of the computation described above.\n",
```
nvaytet, line 65:
```suggestion
    "`true\\_two\\_theta` is the input value for $2 \\theta$ and `approx\\_two\\_theta` is the result of the computation described above.\n",
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "error = sc.DataArray(\n",
+    "    error,\n",
+    "    coords={\"two_theta\": two_theta, \"z_det\": z_det, \"phi\": phi, \"wavelength\": wavelength},\n",
+    ")\n",
+    "\n",
+    "fig, axs = plt.subplots(2, 3, layout=\"constrained\")\n",
+    "for i, ax in zip(range(error.sizes[\"wavelength\"]), axs.flat, strict=True):\n",
+    "    da = error[\"wavelength\", i]\n",
+    "    da.plot(ax=ax)\n",
+    "    ax.set_title(f\"$\\\\lambda$={da.coords['wavelength'].value:.0f}$\\\\AA$\")"
```
nvaytet, line 131:
```suggestion
    "    da.plot(ax=ax, title=f\"$\\\\lambda$={da.coords['wavelength'].value:.0f}$\\\\AA$\")"
```

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "approx = two_theta_approx(\n",
+    "    wavelength=wavelength, two_theta=two_theta, phi=phi, z_det=z_det\n",
+    ")\n",
+    "error = (two_theta - approx) / two_theta\n",
+    "error = sc.DataArray(\n",
+    "    error,\n",
+    "    coords={\"two_theta\": two_theta, \"z_det\": z_det, \"phi\": phi, \"wavelength\": wavelength},\n",
+    ")\n",
+    "\n",
+    "fig, axs = plt.subplots(2, 3, layout=\"constrained\")\n",
```
nvaytet, line 127: There is a bit of horizontal space left: maybe make the figure a bit larger with e.g. `figsize=(9, 7)`?
Also, to try and fix the overlapping axis labels, instead of `layout=\"constrained\"` you could try to do `fig.tight_layout()` after you've created it?
Screenshot at 2024-06-21 10-35-21

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "This includes SANS and reflectometry.\n",
+    "In those cases, we need to take gravity into account when computing the scattering angle $2\\theta$.\n",
+    "This document derives the expression that ScippNeutron uses in this case.\n",
+    "Further, this expression contains an approximation that explained and analyzed.\n",
+    "\n",
+    "## Derivation and approximation\n",
+    "\n",
+    "After scattering, the neutron travels along curve\n",
+    "$$\n",
+    "C: \\vec{c}(t) = \\begin{pmatrix}v_x t \\\\ -\\frac{g}{2}t^2 + v_y t \\\\ v_z t\\end{pmatrix}.\n",
```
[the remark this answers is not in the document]
jl-wynen: I have never heard that rule. And it honestly sounds like a lot of typing because there are many subscripts. Is it because you see the `y` subscript as a concrete value of a general index of `v`?

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
jl-wynen: I would say `m_n` is a combined name. `m` doesn't have an index where `n` is a concrete value of that index.

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
jl-wynen: Same as `m_n` above.

docs/user-guide/algorithms-background/two_theta-gravity-correction.ipynb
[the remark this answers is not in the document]
nvaytet: I was told (but maybe it's not a rule everywhere) that an italic font indicates something that can have different values, whereas `x, y, z, d` here are just to say it's the `x` component of `v`.

But I really don't mind, you can also leave it.
And because it was a lot of typing, I thought I would do the work for you ;-)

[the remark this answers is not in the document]
jl-wynen: Thanks, I appreciate the help! Who told you this? Was it in astronomy or in neutron scattering? I could imagine different communities having different rules.

nvaytet (approved):

PR #519, opened 2024-05-28
Fix conda build

jl-wynen:
The type hints were bad because `mp.Queue` is a function, not a class. This was not caught earlier because we don't run mypy and `Optional[mp.Queue]` is allowed. But `mp.Queue | None` is not allowed because the pipe operator is only implemented for types, not functions.

I added sciline as a dependency because we were skipping the frame unwrapping tests both in the package build and CI.


YooSunYoung (approved):

YooSunYoung (changes requested):
Docs build still fails though...

MridulS:
(probably) relavent issue https://github.com/sphinx-doc/sphinx/issues/12300

jl-wynen:
@YooSunYoung it is working now

PR #528, opened 2024-06-25
Scattering angle for reflectometry

jl-wynen:
First step for https://github.com/scipp/essreflectometry/issues/54

This computes $\gamma$ as defined in https://doi.org/10.1016/j.nima.2016.03.007. ESSreflectometry will subtract $\omega$ to obtain $\theta$ because, this way, `scattering_angle_in_yz_plane` is more generic.

### DIFF
--- a/src/scippneutron/conversion/beamline.py
+++ b/src/scippneutron/conversion/beamline.py
@@ -536,3 +536,30 @@ def scattering_angles_with_gravity(
     two_theta_ = sc.atan2(y=y, x=z, out=y)
 
     return {'two_theta': two_theta_, 'phi': phi}
+
+
+def scattering_angle_in_yz_plane(
+    incident_beam: sc.Variable,
+    scattered_beam: sc.Variable,
+    wavelength: sc.Variable,
+    gravity: sc.Variable,
+    sample_rotation: sc.Variable,
+) -> sc.Variable:
+    match beam_aligned_unit_vectors(incident_beam=incident_beam, gravity=gravity):
+        case {
+            'beam_aligned_unit_y': ey,
+            'beam_aligned_unit_z': ez,
+        }:
+            pass
+        case _:
+            raise RuntimeError('Unexpected return value of beam_aligned_unit_vectors')
+
+    y = _drop_due_to_gravity(
+        distance=sc.norm(scattered_beam), wavelength=wavelength, gravity=gravity
+    )
+    y += sc.dot(scattered_beam, ey).to(dtype=elem_dtype(wavelength), copy=False)
+    y = sc.abs(y, out=y)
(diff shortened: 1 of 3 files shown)


tests/conversion/beamline_conversions_test.py
```diff
@@ -544,6 +544,346 @@ def test_scattering_angles_with_gravity_supports_mismatching_units():
...
+        scattered_beam=scattered_beam,
+        wavelength=wavelength,
+        gravity=gravity,
+    )
+
+    sc.testing.assert_allclose(res, sc.scalar(polar, unit='rad'))
+
+
+@pytest.mark.parametrize('polar', [np.pi / 3, np.pi / 2, 2 * np.pi / 3, np.pi])
+def test_scattering_angle_in_yz_plane_reproduces_angles_azimuth_greater_pi(
```
nvaytet, line 618: I may be wrong but it looks to me like this test is doing exactly the same as the previous test? Is it that the values in the parametrize should be different?

src/scippneutron/conversion/beamline.py
```diff
@@ -536,3 +542,83 @@ def scattering_angles_with_gravity(
...
+    ey = unit_vectors['beam_aligned_unit_y']
+    ez = unit_vectors['beam_aligned_unit_z']
+
+    y = _drop_due_to_gravity(
+        distance=sc.norm(scattered_beam), wavelength=wavelength, gravity=gravity
+    )
+    y += sc.dot(scattered_beam, ey).to(dtype=elem_dtype(wavelength), copy=False)
+    y = sc.abs(y, out=y)
+    z = sc.dot(scattered_beam, ez).to(dtype=elem_dtype(y), copy=False)
+    return sc.atan2(y=y, x=z, out=y)
```
nvaytet, line 624: Looks good but same question as before: have you checked that it gives the same results as before?

I guess we also expect a difference because we used `asin` in the [reflectometry code](https://github.com/scipp/essreflectometry/blob/main/src/ess/reflectometry/conversions.py#L70), but if we replace that by an `atan` we should get identical results?

tests/conversion/beamline_conversions_test.py
jl-wynen: You are right. This was copied from a test for the regular scattering angles that includes the azimuthal angle. But I removed that angle...

src/scippneutron/conversion/beamline.py
jl-wynen: It does not give the same results because of the sign error in the old code. I did not see any difference from the asin / atan change. But I am still investigating.

jl-wynen:
See https://github.com/scipp/essreflectometry/pull/55

nvaytet (approved):

PR #529, opened 2024-07-04
Add notebook for computing WFM time-of-flight

nvaytet:
- Add notebook that computes `tof` for neutron events from WFM beamline
- Add a new provider to specify TimeOfFlightOrigin parameters manually.

Draft because no unit test yet, but still worth reviewing the notebook and changes in providers.


src/scippneutron/tof/chopper_cascade.py
```diff
@@ -505,6 +507,36 @@ def __getitem__(self, key) -> Chopper:
             time_close=self.time_close[key],
         )
 
+    @classmethod
+    def from_disk_chopper(
+        cls,
+        disk_chopper: DiskChopper,
+        pulse_frequency: sc.Variable,
+        npulses: int = 1,
```
SimonHeybrock, line 515: Maybe use a different (larger) default. I think there is probably no harm in making this larger than necessary.

src/scippneutron/tof/unwrap.py
```diff
@@ -404,6 +413,38 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
     return TimeOffset(unwrapped.coords['time_offset'])
 
 
+def time_of_flight_origin_wfm(
+    distance: TimeOfFlightOriginDistance,
+    time_open: TimeOfFlightOriginTime,
```
SimonHeybrock, line 418: Bad name, since it is not the time when the slits open? How about `subframe_origin_time`?

src/scippneutron/tof/unwrap.py
```diff
@@ -404,6 +413,38 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
+    )
+    shift = sc.concat([shift, nans[0]], 'subframe')
+    return TimeOfFlightOrigin(
+        time=sc.DataArray(shift, coords={'subframe': times}),
+        distance=distance,
+    )
+
+
+# TODO: add a helper provider that can make time of flight origin from a pair of WFM
+# choppers
```
SimonHeybrock, line 445: I suppose then either the above or below functions goes way, since it is split into two as per our discussion?

docs/user-guide/wfm/wfm-time-of-flight.ipynb
```diff
@@ -0,0 +1,625 @@
...
+   "source": [
+    "Ltotal = sc.scalar(26., unit='m')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "360be644-4c0d-4a0c-9497-af5a603cbfe1",
+   "metadata": {},
+   "source": [
+    "## Creating some neutron events\n",
```
SimonHeybrock, line 182: Note a lot of overlap above and below with the fakes that were being introduced in https://github.com/scipp/scippneutron/blob/main/src/scippneutron/tof/fakes.py. Can we extend and use that?

docs/user-guide/wfm/wfm-time-of-flight.ipynb
```diff
@@ -0,0 +1,625 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "43755616-c1ce-4218-9255-0a2a7405daf1",
+   "metadata": {},
+   "source": [
+    "## Using the chopper cascade to chop the pulse\n",
+    "\n",
+    "The `chopper_cascade` module (abbreviated here as `cc`) can now be used to chop a pulse of neutrons using the choppers created above.\n",
```
SimonHeybrock, line 311: I'd avoid importing under an alias unless a module is used many times, since it just adds to the cognitive overhead. Spell it out!

docs/user-guide/wfm/wfm-time-of-flight.ipynb
```diff
@@ -0,0 +1,625 @@
...
+    ")\n",
+    "\n",
+    "pl[uw.PulsePeriod] = sc.reciprocal(source.frequency)\n",
+    "pl[uw.PulseStride] = None\n",
+    "pl[uw.SourceTimeRange] = pulse_times.min(), pulse_times.max()\n",
+    "pl[uw.SourceWavelengthRange] = pulse_wavs.min(), sc.scalar(11.0, unit='angstrom')\n",
+    "pl[uw.Choppers] = choppers\n",
+    "\n",
+    "# Use provider to set time of flight origin manually\n",
+    "pl.insert(uw.time_of_flight_origin_wfm)\n",
```
SimonHeybrock, line 435: - I think we should just have a single provider, I don't think there is need to modify the pipeline? Just set the origin by hand, which will replace the provider which computed it?
- Remember that we are moving away from the "pipeline" terminology, use "workflow", don't mention "provider". See https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#c-5-use-a-fixed-pattern-for-creating-manipulating-and-running-workflows.

SimonHeybrock (commented):
Small comments, looks good otherwise:

docs/user-guide/wfm/wfm-time-of-flight.ipynb
nvaytet: I tried to use `FakeSource` and `FakeBeamline` to create multiple pulses of neutrons that span different wavelength ranges, and then just merge all the events in a single data array.

Something like:
```Py

# Make the ESS source with tof
N = 500_000  # Number of neutrons in a pulse
source = tof.Source(facility='ess', neutrons=N)

# Histogram the wavelengths in coarse bins, to make 20 pulses below
neutrons_per_wavelength = source.data.squeeze().hist(wavelength=sc.linspace('wavelength', 0.05, 11.0, 21, unit='angstrom'))

choppers = {
    key: cc.Chopper.from_disk_chopper(
             chop,
             pulse_frequency=source.frequency,
             npulses=1,
         )
    for key, chop in disk_choppers.items()
}

events = None

for n in neutrons_per_wavelength:
    s = fakes.FakeSource(
        frequency=sc.scalar(14.0, unit='Hz'), run_length=sc.scalar(10.0, unit='s'),
        events_per_pulse=int(n.value)
    )

    p = fakes.FakePulse(
        time_min=sc.scalar(0.0, unit='ms'),
        time_max=sc.scalar(3.0, unit='ms'),
        wavelength_min=n.coords['wavelength'].min(),
        wavelength_max=n.coords['wavelength'].max(),
    )
    
    beamline = fakes.FakeBeamline(
        source=s,
        pulse=p,
        choppers=choppers,
        monitors={'monitor': Ltotal},
        detectors={},
    )
    # Need try/except as some wavelength ranges have 0 neutrons that make it through
    try:
        mon, ref = beamline.get_monitor('monitor')
        if events is None:
            events = mon
        else:
            events = events.bins.concatenate(mon)
    except (ValueError, NotImplementedError):
        pass
```
but the final events don't look great. The different pulses have sharp edges, and some overlap where we see spikes in the histogrammed plot:
Screenshot at 2024-07-05 14-03-58

Compared to when I generated the events with `tof`:
Screenshot at 2024-07-05 14-04-13

nvaytet: Hmm if I use more bins, I get something that's better. Maybe I can tweak this further...
Screenshot at 2024-07-05 14-24-12

docs/user-guide/wfm/wfm-time-of-flight.ipynb
nvaytet: I tried to replace the old provider with one that uses a pair of WFM choppers.
I've also left the possibility of setting the time of flight origin and distance manually (I hope!).
See update.

src/scippneutron/tof/unwrap.py
```diff
@@ -430,17 +438,50 @@ def time_of_flight_origin_wfm_from_chopper(
...
 
 
+def time_of_flight_origin_time_wfm_from_choppers(
+    wfm_choppers: WFMChoppers,
+) -> TimeOfFlightOriginTime:
+    times = []
+    for ch in wfm_choppers:
+        times.append(ch.time_open)
+        times.append(ch.time_close)
+    return TimeOfFlightOriginTime(sc.concat(times, dim='dummy').mean('dummy'))
```
SimonHeybrock, line 482: Fine, but not also https://scipp.github.io/generated/functions/scipp.reduce.html#scipp-reduce.

src/scippneutron/tof/unwrap.py
```diff
@@ -404,24 +424,12 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
-    WFM choppers, in which case the time-of-flight origin may be defined using some
-    combination of the two choppers. This is not supported yet, as we only support
-    a single source chopper input.
-    """
-    if len(source_chopper.time_open) != subframe_bounds['time'].sizes['subframe']:
-        raise NotImplementedError(
-            "Source chopper openings do not match the subframe count, this is ."
-            "not supported yet."
-        )
+    """ """
```
SimonHeybrock, line 432: Docstring removed by accident?

src/scippneutron/tof/unwrap.py
```diff
@@ -430,17 +438,50 @@ def time_of_flight_origin_wfm_from_chopper(
...
     high = times[-1] + padding
     times = sc.concat([low, times, high], 'subframe')
-    shift[1::2] += 0.5 * (
-        source_chopper.time_open + source_chopper.time_close
-    ).rename_dims(cutout='subframe')
-    # Set offsets before, between, and after subframes to NaN
-    shift[::2] = sc.scalar(math.nan, unit='s')
+
+    # Select the first n time origins to match subframe bounds
+    subframe_origin_time = subframe_origin_time[: subframe_bounds.sizes['subframe']]
```
SimonHeybrock, line 443: Um, why?

src/scippneutron/tof/unwrap.py
```diff
@@ -430,17 +438,50 @@ def time_of_flight_origin_wfm_from_chopper(
...
+
+    # We need to add nans between each subframe_origin_time offsets for the bins before,
+    # after, and between the subframes.
+    nans = sc.full_like(subframe_origin_time, value=math.nan)
+    shift = (
+        sc.concat([nans, subframe_origin_time], dim=uuid4().hex)
+        .transpose()
+        .flatten(to='subframe')
+    )
+    shift = sc.concat([shift, nans[0]], 'subframe')
```
SimonHeybrock, line 453: Unused `shift` from above`? Also, why not use `full_like` to make `shift`, then use setitem with `[1::2]` to set the times?

src/scippneutron/tof/unwrap.py
nvaytet: Not sure what else to do here? The `subframe_origin_time` has maybe a length of e.g. 30, but we know the choppers only have 6 openings. The dimensions wouldn't agree further down without the slicing, I think?

SimonHeybrock: Sure, but why the first 6? It could just as well be the last 6, or some in the middle?

nvaytet: I think this is the last thing I have not yet addressed. But I am unsure how to proceed. Do you have a suggestion?

SimonHeybrock: Not really, unfortunately. Require that only the relevant times and/or chopper openings are passed in?

nvaytet: See if the update is reasonable.

SimonHeybrock (approved):
👍 Fingers crossed that this is one of the last few iterations on the WFM work.

PR #530, opened 2024-07-09
Remove load_nexus and data streaming

jl-wynen:
Removes the deprecated `load_nexus` function and surrounding utilities. Also removes the data streaming package because it depends on `load_nexus`. The streaming code should have been superseded by Beamlime by now. 

I did not find any usages of either `load_nexus` or data streaming in downstream projects. (Except for `ess` which we no longer use anyway.)


YooSunYoung (approved):

docs/developer/file-loading.rst
```diff
@@ -1,9 +1,10 @@
 File loading
 ============
 
-Scippneutron is able to load Nexus files from ESS experimental runs,
-as well as data files from other neutron facilities.
-There are currently two different implementations of file loading.
+Scippneutron does not provide routines to read files itself.
+However, it is able to load data files from many neutron facilities by using Mantid.
+For NeXus files, prefer `ScippNeXus <https://scipp.github.io/scippnexus/>`_
```
SimonHeybrock, line 6:
```suggestion
For NeXus files, prefer `ScippNexus <https://scipp.github.io/scippnexus/>`_
```

src/scippneutron/io/nexus/_json_nexus.py
```diff
@@ -1,387 +0,0 @@
...
-    @property
-    def file(self):
-        return self._file
-
-    @property
-    def parent(self):
-        return self._parent
-
-
-class JSONDataset(JSONNode):
```
SimonHeybrock, line 244: These wrappers are in use by Beamlime (indirectly), for loading JSON Nexus data using ScippNexus. Have they been moved somewhere?

SimonHeybrock (changes requested):

src/scippneutron/io/nexus/_json_nexus.py
jl-wynen: @YooSunYoung can you check? As I understood it, you no longer use the code here.

SimonHeybrock: @jl-wynen Remember the Nexus loaders you added in ESSreduce, and how we wanted them to work with JSON Nexus? This is making use of these wrappers.

YooSunYoung: Ah ha...! Yes this was the JsonNexus wrapper I meant.
Sorry I missed this one... it is indeed used in the beamlime.
Should we move this to `scippnexus`...?

src/scippneutron/io/nexus/_json_nexus.py
```diff
@@ -1,387 +0,0 @@
...
-        return self[name]
-
-
-@dataclass
-class StreamInfo:
-    topic: str
-    flatbuffer_id: str
-    source_name: str
-    dtype: Any
-    unit: str
```
YooSunYoung, line 355: This part about stream is probably not needed.
They are parsed in a different way now.

src/scippneutron/io/nexus/_json_nexus.py
SimonHeybrock: > Should we move this to scippnexus...?

This was discussed previously: Since the JSON-Nexus is not standard but ESS specific we felt it should not be part of ScippNexus (and maybe not even ScippNeutron, i.e., ESSreduce might make more sense?).

jl-wynen: I can move them to ESSreduce.

jl-wynen: See https://github.com/scipp/essreduce/pull/52

SimonHeybrock (approved):

PR #543, opened 2024-08-29
Update Scipp and Plopp

jl-wynen:
This should fix the broken `nightly_at_main` build. Do not merge until plopp has been released!


nvaytet (approved):

nvaytet:
I think we're just missing cbar=True in the call to scatter3d.

jl-wynen:
There seem to be some breaking changes in Mantid. I don't know how to fix the last test:
```python
    def test_advanced_geometry_with_absent_shape(self):
        # single bank 3 by 3
        ws = mantid.CreateSampleWorkspace(
            NumBanks=1, BankPixelWidth=3, StoreInADS=False
        )
        # Save and reload trick to purge sample shape info
        file_name = "example_geometry.nxs"
        geom_path = os.path.join(tempfile.gettempdir(), file_name)
        mantid.SaveNexusGeometry(ws, geom_path)  # Does not save shape info
        assert os.path.isfile(geom_path)  # sanity check
        out = mantid.LoadEmptyInstrument(
            Filename=geom_path, StoreInADS=False
        )  # reload without sample info
        os.remove(geom_path)

        assert not out.componentInfo().hasValidShape(0)  # sanity check
        dg = scn.mantid.from_mantid(out, advanced_geometry=True)
        # Shapes have zero size
        assert sc.identical(
            sc.sum(dg['shape']), sc.vector(value=[0, 0, 0], unit=sc.units.m)
        )
```
This fails in `mantid.LoadEmptyInstrument(Filename=geom_path, StoreInADS=False)` with
```
RuntimeError: LoadEmptyInstrument-v1: No instrument XML definition found in /tmp/example_geometry.nxs at mantid_workspace_1/instrument/instrument_xml or at raw_data_1/instrument/instrument_xml
```

It looks like Mantid now requires an instrument file for every instrument. But we don't have one for the dummy instrument in this test. What should we do? Can we remove the test? Or do we need to change it to use a real instrument?

nvaytet:
> What should we do? Can we remove the test? Or do we need to change it to use a real instrument?

Mark the test as skipped? It's blocking work downstream that need to update dependencies.

PR #548, opened 2024-09-05
Add a builder class for CIF files

### LINKED ISSUE #547: CIF writer should mark empty strings
Currently, the CIF writer emits nothing for empty strings. This is fine (I think) in pairs. But in loops, it makes it impossible to find out which columns are empty. So we should instead write `''`.

jl-wynen:
This PR adds a new class to simplify writing CIF files. When implementing https://github.com/scipp/essdiffraction/pull/89, I found that the existing low-level interface is tedious to use. And I needed to implement some functionality that should be in ScippNeutron and not ESS*, such as author and beamline handling. This will in particular be relevant once #473 is implemented. (Note that the current PR is meant as a stop-gap solution until the models from that issue are added.)

Fixes #547

### DIFF
--- a/src/scippneutron/io/cif.py
+++ b/src/scippneutron/io/cif.py
@@ -63,10 +63,10 @@ from __future__ import annotations
 
 import io
 import warnings
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Mapping, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


src/scippneutron/io/cif.py
```diff
@@ -104,17 +173,306 @@ def save_cif(
...
+    content:
+        One or more CIF data blocks or a ``CIF`` object to write to the file.
+    comment:
+        Optional comment that can be written at the top of the file.
     """
-    if isinstance(blocks, Block):
-        blocks = (blocks,)
+    if isinstance(content, CIF):
+        if comment:
+            raise ValueError('Cannot specify a comment when saving a CIF object')
```
YooSunYoung, line 183: Can't we just replace the comment if another one is given...?
Or will it be too flacky...?

YooSunYoung (approved):

src/scippneutron/io/cif.py
jl-wynen: You are right. I changed to it overwrite the comment. I don't think this is too flaky.

PR #560, opened 2024-10-14
Update of WFM tof workflow to take into account DREAM choppers

nvaytet:
The DREAM pulse-shaping choppers (PSC) are different from the V20 or ODIN WFM choppers in the way that they have 8 cutouts that can shape the pulse in different ways.

In the example McStas simulation that we have, the detector sees 2 frames made from 8 cutouts.
The workflow that computed time-of-flight from event-time-offset that we had assumed that the number of frames was equal to number of cutouts.

This PR updates the workflow to handle both cases.
Note that we re-wrote the workflow graph to simplify it, from
old-graph
to
new-graph


src/scippneutron/tof/unwrap.py
```diff
@@ -428,121 +441,97 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
-    """
-    times = subframe_bounds['time'].flatten(dims=['subframe', 'bound'], to='subframe')
+    ltotal:
+        The total distance between the source and the detector.
+
+    Notes
+    -----
+    To find the time-of-flight origin, we ray-trace the fastest and slowest neutrons of
+    the subframes back to the first chopper to determine the time-of-flight origin.
+    The assumption here is that the first chopper is one of the two wfm choppers.
```
SimonHeybrock, line 465: Is this the case for all ESS instruments with WFM?

src/scippneutron/tof/unwrap.py
```diff
@@ -428,121 +441,97 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
+    tmin = detector_subframe_bounds['time'].min('bound')
+    tmax = detector_subframe_bounds['time'].max('bound')
+    wmin = detector_subframe_bounds['wavelength'].min('bound')
+    wmax = detector_subframe_bounds['wavelength'].max('bound')
+    # Ray-trace back to the position of the first chopper (note that frame 0 is the
+    # pulse itself)
+    distance = frames[1].distance
+    dist = ltotal - distance
+    t1 = tmin - dist * chopper_cascade.wavelength_to_inverse_velocity(wmin)
+    t2 = tmax - dist * chopper_cascade.wavelength_to_inverse_velocity(wmax)
```
SimonHeybrock, line 518: `chopper_cascade` can propagate forwards, e.g., to the detector position. Can't the same code be used to go backwards?

tests/tof/unwrap_test.py
```diff
@@ -334,21 +334,15 @@ def test_pulse_skipping_unwrap_histogram_mode_not_implemented(
 
 def test_wfm_unwrap(ess_10s_14Hz, ess_pulse) -> None:
     distance = sc.scalar(20.0, unit='m')
-    choppers = fakes.wfm_choppers.copy()
-    # We currently have no way of detecting which cutouts of the "source" chopper
-    # are used (not blocked by other choppers), for this test we remove those we
-    # know are not used.
```
SimonHeybrock, line 340: Maybe replace by comment that explains (and assertion) that there *are* extra openings but the code knows how to deal with them automatically?

SimonHeybrock (commented):
Big congrats on getting this to work! 👍 

I have only looked at the code changes, not the notebooks. Maybe a walkthrough would be best?

src/scippneutron/tof/unwrap.py
SimonHeybrock: I think you could just use https://github.com/scipp/scippneutron/blob/17f727d078d4e51c5806f4cf5973d568208f62b0/src/scippneutron/tof/chopper_cascade.py#L28

SimonHeybrock: Or use https://github.com/scipp/scippneutron/blob/17f727d078d4e51c5806f4cf5973d568208f62b0/src/scippneutron/tof/chopper_cascade.py#L87, if we can have a proper frame setup?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -381,6 +383,10 @@ def time_offset_open(self, *, pulse_frequency: sc.Variable) -> sc.Variable:
         ----------
         pulse_frequency;
             Frequency of the neutron source.
+        start_rotation:
+            This can be used to include rotations before the first pulse by giving a
+            negative integer. This is useful for choppers that are not in phase with
+            the source. The default is 0.
```
SimonHeybrock, line 389: Is there ever a case where we don't want to include such rotations? That is, is this argument even needed?

src/scippneutron/chopper/disk_chopper.py
```diff
@@ -438,41 +454,56 @@ def time_offset_angle_at_beam(
...
             Defined anticlockwise with respect to top-dead-center.
-        n_repetitions:
-            Return this many times for each angle corresponding to multiple rotations
-            of the chopper.
+        start_rotation:
+            This can be used to include rotations before the first pulse by giving a
+            negative integer. This is useful for choppers that are not in phase with
+            the source. The default is 0.
+        end_rotation:
+            How many chopper rotations to perform.
```
SimonHeybrock, line 462: Same question as above: Maybe the existing argument was fine and we should just ensure that we always include the start?

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -248,14 +248,33 @@ class Bound:
...
+                        bound.wav_start,
+                    )
+                    merged_bounds.append(current)
+                    current = Bound(
+                        old_end,
+                        bound.end,
+                        old_wav_end,
+                        bound.wav_end,
+                    )
+                # If end is before current end, overlap is total and we merge
```
SimonHeybrock, line 270: I don't get this. What is being clipped? What is meant be "overlap is partial" and why don't we just want to merge?

src/scippneutron/chopper/disk_chopper.py
nvaytet: Changed this to now always have the -1 rotation.

nvaytet:
Note that CI will fail until https://github.com/scipp/scipp/pull/3582 is merged and released.

nvaytet:
@SimonHeybrock I believe this is ready for another look.

src/scippneutron/tof/unwrap.py
```diff
@@ -217,17 +218,26 @@ def frame_at_detector(
...
+) -> FrameAtDetector:
+    """
+    Return the frame at the detector.
+
+    This is the result of propagating the source pulse through the chopper cascade to
+    the detector. The detector may be a monitor or a detector after scattering off the
+    sample. The frame bounds are then computed from this.
+    """
+    return FrameAtDetector(frames[-1].propagate_to(ltotal))
+    # return FrameAtDetector(frames[ltotal])
```
SimonHeybrock, line 236: Commented code?

src/scippneutron/tof/unwrap.py
```diff
@@ -333,7 +343,8 @@ def offset_from_wrapped(
     """
     time_bounds = frame_bounds['time']
     frame_period = frame_period.to(unit=elem_unit(time_bounds))
-    if time_bounds['bound', -1] - time_bounds['bound', 0] > frame_period:
+    diff = (time_bounds['bound', -1] - time_bounds['bound', 0]) - frame_period
+    if any(diff.flatten(to='x') > sc.scalar(0.0, unit=frame_period.unit)):
```
SimonHeybrock, line 347: I am confused by the `flatten`, why is it needed? Can we just
```suggestion
    if (diff > sc.scalar(0.0, unit=frame_period.unit)).any():
```

src/scippneutron/tof/unwrap.py
```diff
@@ -427,120 +438,188 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
+        1. partial overlap:
+        |-------------|            ->   |--------|
+                 |-------------|   ->                 |--------|
+
+        2. total overlap:
+        |----------------------|   ->   |--------|       |-----|
+                 |-------|         ->
+
+    In the case of multiple detector pixels, we find the pixel closest to the source
+    and use that as the reference for the clipping.
```
SimonHeybrock, line 460: Hmm, shouldn't you get more overlap for pixels with the *greatest* distance?

src/scippneutron/tof/unwrap.py
```diff
@@ -427,120 +438,188 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
     """
-    Compute the time-of-flight origin in the WFM case.
-    For WFM there is not a single time-of-flight "origin", but one for each subframe.
-    In the case of a pair of WFM choppers, the distance to the time-of-flight origin
-    would typically be the average of the distances of the choppers, while the time
-    of the time-of-flight origin would be the average of the opening and closing times
-    of the WFM choppers.
+    Check for time overlap between subframes.
+    If overlap is found, we clip away the regions where there is overlap.
+    This is done by adding a fake chopper which is closed during the overlapping times.
```
SimonHeybrock, line 447: Not clear to me why we still need choppers at this point. Instead of adding a fake chopper, can't we just clip the (sub)frame bounds?

src/scippneutron/tof/unwrap.py
```diff
@@ -427,120 +438,188 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
+    for subframe in last_frame.subframes:
+        f = chopper_cascade.Frame(distance=last_frame.distance, subframes=[subframe])
+        chopped = f.chop(fake_chopper)
+        subframes.extend(
+            [
+                sf
+                for sf in chopped.subframes
+                if not sc.allclose(sf.start_time, sf.end_time)
+            ]
+        )
```
SimonHeybrock, line 521: All this (and above) looks really complicated (read: very difficult to verify). I wonder if it is due to the fake-chopper. Why can't we just compute bounds, clip, and continue?

src/scippneutron/tof/unwrap.py
```diff
@@ -427,120 +438,188 @@ def time_offset(unwrapped: UnwrappedData) -> TimeOffset:
...
 
-def time_of_flight_origin_time_wfm_from_choppers(
-    wfm_choppers: WFMChoppers,
-) -> TimeOfFlightOriginTime:
-    """
-    Compute the time of the time of flight origin, in the case of a set
-    (usually a pair) of WFM choppers. The time is the mean of the opening and closing
-    times of the WFM choppers.
+    # TODO: what dimension order is the best here? Where should `detector_number` be?
+    origin_time = sc.DataArray(shift.transpose(times.dims), coords={'subframe': times})
```
SimonHeybrock, line 620: If this is meant for processing the raw data, `detector_number` should be first (slowest dim), since that is how data is sorted.

src/scippneutron/tof/unwrap.py
```diff
@@ -608,7 +687,8 @@ def to_time_of_flight(
     delta = origin.time
     if isinstance(delta, sc.DataArray):
         # Will raise if subframes overlap, since coord for lookup table must be sorted
-        delta = sc.lookup(delta, dim='subframe')[time_offset]
+        # TODO: Can we do without making a copy of delta?
+        delta = sc.lookup(delta.copy(deep=True), dim='subframe')[time_offset]
```
SimonHeybrock, line 691: Shouldn't/isn't `sc.lookup` doing this?

src/scippneutron/tof/unwrap.py
nvaytet: Yes, I actually left it in there intentionally because I wasn't sure if I should use the `getitem` syntax.
I believe I needed to make the change because inside the `getitem` it assumes that the distance passed is a scalar.

Should I fix in the `getitem` or should I keep the code I have now?

To fix in the `getitem`, we would have to handle cases where you propagate to a range of distances, some might be before the last frame, some might be after... It gets messy.
I could also just raise a helpful error message if the distance passed in the `getitem` is not a scalar?

SimonHeybrock: I don't have a strong opinion, do what you think works best.

src/scippneutron/tof/unwrap.py
nvaytet: The flatten was to handle both scalars and variables with dimensions in a generic way, but your suggestion also does this and avoids the flatten :+1:

src/scippneutron/tof/unwrap.py
nvaytet: I am using `chop` because I need to have a well-constructed frame, that I can then propagate backwards to find the time origin.
I could just clip the time range, but I also want to clip the wavelengths in the correct way, so the backward propagation is correct. I thought the best way was to use `chop`.

Do you have a another suggestion?

src/scippneutron/tof/unwrap.py
nvaytet: Yes, I think you're right.

src/scippneutron/tof/unwrap.py
SimonHeybrock: The frame-propagation is already a free function. Wouldn't it still be easier to compute time and wavelength bounds directly?

SimonHeybrock: I honestly do not remember, it has been too long since I wrote this. We can have a look together one day if you like?

nvaytet: Hmm I think I may have a way to do this, by taking the mean of the intersection from all vertices of the subframe.

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -507,6 +507,8 @@ def from_disk_chopper(
             Number of pulses to rotate the chopper for.
         """
         tpulse = 1.0 / pulse_frequency
+        # Start at -1 to ensure a rotation that is finishing when the pulse begins is
+        # also included.
```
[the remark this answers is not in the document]
nvaytet: Removed. Was left over from previous changes.

src/scippneutron/tof/unwrap.py
nvaytet: I think it doesn't matter because it will just propagate to the other distances and there are no more choppers in the way.

src/scippneutron/tof/unwrap.py
nvaytet: I've looked into simplifying this without using the fake chopper, for many pixels in a vectorized way. It's proving difficult.

My suggestion would be to merge this implementation so that it doesn't block all the other workflows.

I will revisit right after with a simpler approach but it should only be internal refactoring and shouldn't change how it is used by workflows.

src/scippneutron/tof/unwrap.py
SimonHeybrock: My point was that if there are two subframes that to not overlap at L2=5 meter, they may overload at any greater L2. Or do I misunderstand the comment?

nvaytet: Yes, you're right, the frames may not overlap on the closest pixel.
However, if we take the overlapping times at the farthest pixel, and use those to make a fake chopper, those times would not match the overlap seen at the closest pixel, they would be shifted to the right.
So the closest frame would have a small region where there is still overlap.

Basically, I think I need to find the overlap at the farthest pixel, and propagate those backwards to the closest pixel distance, and then make my fake chopper...

As discussed, this fake chopper implementation is not great.
@jokasimr and I do have a prototype of an alternative method which is working.
Question is: should we spend time on fixing this, or should we just try to get that into a production shape instead of trying to fix this implementation?

We could leave this one as is and assume this edge case won't be an issue in the next couple of weeks.
We then can roll out the alternative version a short time after?

SimonHeybrock: Why is it an edge case? Doesn't it affect all (WFM) beamlines that do not have pixels on a sphere? Or do you mean the need for clipping is an edge case (unusual DREAM settings?).

nvaytet: I mean that both the DREAM choppers are a little unusual, and that the case where some pixels have overlap and others don't is hopefully an edge case as the range of pixel distances is not too large for a given instrument?
(I guess Loki is the odd one out, but then don't have WFM choppers)

SimonHeybrock: Then we should just use the furthest right (and live with throwing away a tiny bit too many events for the closer pixels)? Otherwise you *always* get overlap for all but the closest pixel.

nvaytet: Let's find a time to chat in a call.

nvaytet: Conclusion from the call: because we are making a gap when adding the fake chopper rather than clipping with an infinitely thin vertical line, the distance it would take for the gap to completely fill in again with overlap is hopefully larger than the range of pixel distances.
So the overlap should be removed for all pixels (in most cases).
user-guide_wfm_dream-wfm_49_0

nvaytet:
@SimonHeybrock any final thoughts?

PR #561, opened 2024-10-17
Fix propagation of chopper cascade frames to an array of distances

nvaytet:
Fix propagation of chopper cascade frames to an array of distances.
We also remove the code that merges frame subbounds.
Overlapping frame bounds will now need to be clipped (or handled differently) in a separate method.
This is defered to #560 .


src/scippneutron/tof/chopper_cascade.py
```diff
@@ -195,30 +220,27 @@ def chop(self, chopper: Chopper) -> Frame:
...
-        start = sc.reduce([sub.start_time for sub in self.subframes]).min()
-        end = sc.reduce([sub.end_time for sub in self.subframes]).max()
-        wav_start = sc.reduce([sub.start_wavelength for sub in self.subframes]).min()
-        wav_end = sc.reduce([sub.end_wavelength for sub in self.subframes]).max()
+        start = sc.reduce([sub.global_start_time for sub in self.subframes]).min()
+        end = sc.reduce([sub.global_end_time for sub in self.subframes]).max()
+        wav_start = sc.reduce(
+            [sub.global_start_wavelength for sub in self.subframes]
+        ).min()
+        wav_end = sc.reduce([sub.global_end_wavelength for sub in self.subframes]).max()
```
SimonHeybrock, line 228: Why are we using the global times here? Shouldn't the bounds be per-pixel? I think the "global" in the docstring was referring to "across all subframes"?

nvaytet: Ok, but since we are no longer merging the subframes, we just use the `start_time` everywhere, and I can remove the `global_start_time`?

src/scippneutron/tof/chopper_cascade.py
```diff
@@ -105,25 +105,50 @@ def propagate_by(self, distance: sc.Variable) -> Subframe:
...
-    def end_wavelength(self) -> sc.Variable:
-        """The end wavelength of the subframe."""
+    def global_end_wavelength(self) -> sc.Variable:
+        """The global end wavelength of the subframe."""
         return self.wavelength.max()
 
+    @property
+    def start_time(self) -> sc.Variable:
+        """The start time of the subframe for each of the distances in self.time."""
+        return self.time.min('vertex' if 'vertex' in self.time.dims else None)
```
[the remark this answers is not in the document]
nvaytet: I can see that hard-coding `vertex` actually only works for something that was created with `from_source_pulse`...
I'm not sure how to determine which dim should be used. The `self.time` should have at least one dimension for the vertices (might not be called `vertex`) and may or may not have another dimension for the pixel distances (and that could have any name also).
How can I know which one to reduce?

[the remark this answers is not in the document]
nvaytet: Can we assume that `self.wavelengths` will not have the distance dimension, and therefore we can reduce along all the `self.wavelengths` dims?

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: That is what I thought, yes. Or am I missing something?

nvaytet:
@SimonHeybrock if you could have another look at this one, as I need it for the rest of the WFM work. Thanks :pray:

SimonHeybrock (approved):

PR #562, opened 2024-10-24
Make CIF writer conform with expectations of EasyDiffraction

jl-wynen:
Most importantly, this changes the default output key to `_pd_proc.intensity_norm`. The reason for this is that in `intensity_net`, the background has been subtracted. But we don't do this in data reduction; we only normalise.

### DIFF
--- a/src/scippneutron/io/cif.py
+++ b/src/scippneutron/io/cif.py
@@ -150,12 +150,12 @@ class CIFSchema:
 CORE_SCHEMA = CIFSchema(
     name='coreCIF',
     version='3.3.0',
-    location='https://github.com/COMCIFS/cif_core/blob/fc3d75a298fd7c0c3cde43633f2a8616e826bfd5/cif_core.dic',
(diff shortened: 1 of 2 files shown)


src/scippneutron/io/cif.py
```diff
@@ -298,7 +298,9 @@ def with_reduced_powder_data(self, data: sc.DataArray, *, comment: str = '') ->
         ``('tof', 'dspacing')``.
         The data array may also have a name in
         ``('intensity_net', 'intensity_norm', 'intensity_total')``.
-        If the name is not set, it defaults to ``'intensity_net'``.
+        If the name is not set, it defaults to ``'intensity_norm'``,
+        whichis appropriate for typical outputs from data reduction.
```
SimonHeybrock, line 302:
```suggestion
        which is appropriate for typical outputs from data reduction.
```

SimonHeybrock (approved):

PR #566, opened 2024-11-07
Copier update

jl-wynen:


.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
nvaytet, line 1: What is the use for this file?

YooSunYoung: https://github.com/scipp/copier_template/commit/b1fe1f9dde6579407bd5880c209edfd345f19700

For uv support I guess...?

jl-wynen: It's from here: https://github.com/scipp/copier_template/pull/210

This makes it so that uv uses the minimum Python version by default.

.github/workflows/ci.yml
```diff
@@ -74,6 +74,6 @@ jobs:
     uses: ./.github/workflows/docs.yml
     with:
       publish: false
-      linkcheck: ${{ contains(matrix.variant.os, 'ubuntu') && github.ref == 'refs/heads/main' }}
+      linkcheck: github.ref == 'refs/heads/main' }}
```
MridulS, line 77:
```suggestion
      linkcheck: github.ref == 'refs/heads/main'
```

MridulS (approved):

PR #577, opened 2024-11-25
Alternative to time-origin tof computation

nvaytet:
This is an alternative implementation of the computation of the `tof` coordinate in the `unwrap` submodule.

The motivations are the following:

1. Simplify the code in `tof/unwrap.py` (finding the tof origin with WFM was complex with a fake chopper and propagating frames backwards along the beamline)

2. Generalize the code so that we only have to use a single set of providers instead of having providers for standard unwrap, pulse-skipping and WFM (only the uncommon case where there are no choppers requires different providers).

**Implementation**

The implementation is basically described in https://tof.readthedocs.io/en/stable/wfm.html.

- We use the `chopper_cascade` module to find the ranges of time and wavelength (polygons) that are reachable at the detector pixels.
- We approximate the polygons with straight lines
- We use these to define lookup tables which give us directly the tof as a function of arrival time at detector


src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,372 @@ def chopper_cascade_frames(
...
+        # between the vertices
+        dims = set(mm.dims) - set(frame_start.dims)
+        weight = sc.sqrt(xdiff**2 + ydiff**2)
+        a = weight * mm
+        for dim in dims:
+            a = a.nansum(dim)
+        a /= weight.sum(dims)
+        # Compute the intercept as the mean of the intercepts of the curves that pass
+        # through the vertices
+        b = (y - a * x).mean('vertex')
```
jokasimr, line 370: I think we should use another method to determine the line that best approximates the polygon.
While it works well in the cases we've seen so far it is not obvious that it will always produce something reasonable.

For example, in a hypothetical situation where two vertices have almost the same x-coord the slope will grow a lot and the average slope might be larger than expected.

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,372 @@ def chopper_cascade_frames(
...
-            + delta.to(unit=time_of_flight.unit, copy=False),
-            keep_inputs=False,
-        )
-        # Generally the coord is now not ordered, might want to cut, swap, and concat,
-        # but it is unclear what to do with the split bin in the middle and how to
-        # join the two halves. The end of the last bin generally does not match the
-        # begin of the first bin, there may be a significant gap (which we could fill
-        # with NaNs or zeros), but there could also be a small overlap.
-    return UnwrappedData(da)
+    return TofCoord(toa)
```
jokasimr, line 438: Should we not assume the neutron left in the middle of the pulse? (Just because that makes a slightly better estimate.)
That is, shouldn't this be something like `toa - sc.scalar(3, unit='ms')/2`?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,372 @@ def chopper_cascade_frames(
...
-    time_of_flight_origin_wfm,
-)
-_non_wfm = (time_of_flight_origin_from_chopper,)
+def re_histogram_tof_data(da: TofData) -> ReHistogrammedTofData:
+    """
+    Histogrammed data that has been converted to `tof` will typically have
+    unsorted bin edges (due to either wrapping of `time_of_flight` or wavelength
+    overlap between subframes).
+    This function re-histograms the data to ensure that the bin edges are sorted.
+    It generates a number of events in each bin with a normal distribution around
```
jokasimr, line 468: The way I understand this function is that it's computing a change of variables of a distribution (a histogram).

That is, starting with a distribution over `toa` and a function `tof(toa)`, we want to compute the distribution over `tof`.

To do this we should use the formula here: https://en.wikipedia.org/wiki/Probability_density_function#Function_of_random_variables_and_change_of_variables_in_the_probability_density_function

If we used that I'm sure we'd end up with a simpler solution here.

jokasimr (commented):
Overall this looks more or less complete to me. Great job.
Left a couple of comments of things I think we should change before merging.

src/scippneutron/tof/unwrap.py
nvaytet: If you could give me some pointers on how you would implement this, I'd be happy to try. I don't think I fully understood what's in the link.

src/scippneutron/tof/unwrap.py
nvaytet: Yes I forgot to look at the notebook you sent.
Thanks for reminding me.

src/scippneutron/tof/unwrap.py
jokasimr: I was looking at this now and it became more complicated than I thought it would be 🤔
What you've already written might be the better solution.
Like you said, if we come up with something later it can be revisited.

src/scippneutron/tof/unwrap.py
nvaytet: The suggested method indeed yields better results :+1: 

Old:
Screenshot_20241126_142030

New:
Screenshot_20241126_141751

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
     """
     if da.bins is None:
         # Canonical name in NXmonitor
-        return PulseWrappedTimeOffset(da.coords['time_of_flight'])
-    return PulseWrappedTimeOffset(da.bins.coords['event_time_offset'])
+        toa = da.coords['time_of_flight']
+    else:
+        time_zero = da.coords['event_time_zero'] - da.coords['event_time_zero'].min()
+        coord = da.bins.coords['event_time_offset']
+        toa = coord + time_zero.to(dtype=float, unit=elem_unit(coord), copy=False)
```
SimonHeybrock, line 239: Unclear what is going on here from context and docstring. Has the outer `event_time_zero` coord been modified and now it is used to apply a shift?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
+    slopes = []
+    intercepts = []
+    subframes = sorted(frame.subframes, key=lambda x: x.start_time.min())
+    edges = []
+    dim = 'vertex'
+
+    for sf in subframes:
+        edges.extend([sf.start_time, sf.end_time])
+        x0 = sf.time - frame_start
+        y0 = (
```
SimonHeybrock, line 355: Might be more readable if use use the actual axes as symbols instead of x and y. Is it time and wavelength?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
-    shift = sc.full(sizes=sizes, value=math.nan, unit='s')
-    shift['subframe', 1::2] = time_origins
-
-    origin_time = sc.DataArray(
-        shift.transpose(times.dims).copy(deep=True), coords={'subframe': times}
-    )
-
-    return TimeOfFlightOrigin(time=origin_time, distance=at_first_chopper.distance)
+    slope = sc.lookup(lookup.slope, dim='subframe')[toa]
+    intercept = sc.lookup(lookup.intercept, dim='subframe')[toa]
```
SimonHeybrock, line 427: I would feel more natural to construct `Lookup` in the previous function. I thought that is also what the type hint in the dataclass said?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
-_non_wfm = (time_of_flight_origin_from_chopper,)
+def re_histogram_tof_data(da: TofData) -> ReHistogrammedTofData:
+    """
+    Histogrammed data that has been converted to `tof` will typically have
+    unsorted bin edges (due to either wrapping of `time_of_flight` or wavelength
+    overlap between subframes).
+    This function re-histograms the data to ensure that the bin edges are sorted.
+    It generates a number of events in each bin with a normal distribution around
+    the bin center. The width of the distribution is the bin width divided by 2.
+    The new events are then histogrammed using a set of sorted bin edges.
```
SimonHeybrock, line 485: How do you plan to handle variances?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
+    events_per_bin = 200
+    min_bin_width = sc.abs(
+        mid_tofs['time_of_flight', 1:] - mid_tofs['time_of_flight', :-1]
+    ).nanmin()
+
+    dim = uuid.uuid4().hex
+
+    spread = sc.array(
+        dims=[dim],
+        values=np.random.normal(size=events_per_bin, scale=min_bin_width.value / 2.5),
```
SimonHeybrock, line 504: 2.5?

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,387 @@ def chopper_cascade_frames(
...
+    mid_tofs = sc.midpoints(da.coords['tof'], dim='time_of_flight')
+    events_per_bin = 200
+    min_bin_width = sc.abs(
+        mid_tofs['time_of_flight', 1:] - mid_tofs['time_of_flight', :-1]
+    ).nanmin()
+
+    dim = uuid.uuid4().hex
+
+    spread = sc.array(
+        dims=[dim],
```
SimonHeybrock, line 503: Did any scientist look over this? I am skeptical about just putting such a solution in here.

SimonHeybrock (commented):
Not a full review, I presume Johannes took care of that.

src/scippneutron/tof/unwrap.py
nvaytet: I added a comment to clarify.

src/scippneutron/tof/unwrap.py
nvaytet: `x` is 'time-of-arrival' and `y` is 'time-of-flight'.
As they are both times, I feel it is less confusing if we name them x and y?

src/scippneutron/tof/unwrap.py
nvaytet: Agreed, and it was like that at one point.
However, if I want to ensure consistency of units, I would then need to also get the `toa` as an argument of the previous function, and I think the graph visualization would make less sense.

I was now thinking that the `SlopeAndInterceptLookup` were the data that can then be used to create a lookup.
But it is true that it is confusing.

Would it be ok to just rename that type to something else, implying that they are not yet lookups at that stage?

src/scippneutron/tof/unwrap.py
nvaytet: Yes, that is a point I raised with Johannes. I am not sure.
This whole function was a bit of an experiment, to try and construct something that could be used further in a workflow, as the `TofData` is a valid result, but is hard to handle because of its unsorted bin edges.

Maybe if we want to use something like this for the purposes of unit testing, it should be moved out of the source and into the tests?

src/scippneutron/tof/unwrap.py
nvaytet: No, as mentioned above, this was kind of an experiment, and I should have labeled it as is, much more clearly.

I tried making a solution where we would keep the subframe dimension, and rebin each subframe to a common axis. This works fine with a single detector pixel, but becomes tricky when the part you want to extract is pixel-dependent (because each subframe has different bounds on different pixels). This is true for both label-based indexing, and adding masks to the data.

I couldn't just directly `rebin` because the bin edges of the data I am trying to rebin are unsorted.

I am very much open to a different approach.
For now, I will leave it in but remove it from the default providers (and manually add the provider in the unit tests).

src/scippneutron/tof/unwrap.py
SimonHeybrock: How about `toa` and `tof`?

nvaytet: Hmm I think it's less readable than x/y?
```
        toa0tof1 = toa0 * tof1
        toa1tof0 = toa1 * tof0
        toa0tof1_toa1tof0 = toa0tof1 - toa1tof0

        A = ((toa0tof1_toa1tof0) / 2).sum(dim)
        toa = ((toa0 + toa1) * (toa0tof1_toa1tof0) / 6).sum(dim)
        tof = ((tof0 + tof1) * (toa0tof1_toa1tof0) / 6).sum(dim)
        toatof = (
            (toa0tof1_toa1tof0)
            * (2 * toa0 * tof0 + toa0tof1 + toa1tof0 + 2 * toa1 * tof1)
            / 24
        ).sum(dim)
        toa2 = ((toa0tof1_toa1tof0) * (toa0**2 + toa0 * toa1 + toa1**2) / 12).sum(dim)
```

jokasimr:
I'll take another look tomorrow!

src/scippneutron/tof/unwrap.py
SimonHeybrock: Looks the same to me, except that it gives more clue as to what is going on. But maybe this x/y code belong into a generic helper function with a clear name and keyword arguments mapping x/y to toa and tof for better readability.

nvaytet: Done.

src/scippneutron/tof/unwrap.py
nvaytet: I modified this slightly, to use a uniform distribution inside each bin instead of a normal distribution.
The low and high bounds for the uniform distribution now matches each bin edges exactly, instead of having events spilling over into neighbouring bins.

SimonHeybrock (approved):

jokasimr:
I'm still working on the review, it'll be done later today

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,386 @@ def chopper_cascade_frames(
...
-    if len(source_chopper.time_open) != 1:
-        raise NotImplementedError(
-            "Source chopper with multiple openings not supported yet."
-        )
-    source_time_open = source_chopper.time_open[0]
-    source_time_close = source_chopper.time_close[0]
-    time_zero = 0.5 * (source_time_open + source_time_close)
-    return TimeOfFlightOrigin(time=time_zero, distance=source_chopper.distance)
+    return TimeOfArrivalModuloPeriod(
+        toa % frame_period.to(unit=elem_unit(toa), copy=False)
```
jokasimr, line 319: Related to the above comment. What we want here is the time since the most recent `frame_time_zero`.

That is `event_time_offset + frame_time_offset`.

In a simulation we know that the first `event_time_zero` coincides with the first `frame_time_offset` and then what is implemented here is correct.

jokasimr (commented):
I think there might be a wrong assumption here that the minimum `event_time_zero` in the file is also the start of a stride in pulse skipping mode.

But what might happen is that the user starts writing the file in the middle of the stride when pulse skipping is used.

Is that taken into consideration here?

Note that this is probably not an issue for simulated data.

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,386 @@ def chopper_cascade_frames(
...
-    return PulseWrappedTimeOffset(da.bins.coords['event_time_offset'])
+        toa = da.coords['time_of_flight']
+    else:
+        # To unwrap the time of arrival, we want to add the event_time_zero to the
+        # event_time_offset. However, we do not really care about the exact datetimes,
+        # we just want to know the offsets with respect to the start of the run.
+        # Hence we use the smallest event_time_zero as the time origin.
+        time_zero = da.coords['event_time_zero'] - da.coords['event_time_zero'].min()
+        coord = da.bins.coords['event_time_offset']
+        toa = coord + time_zero.to(dtype=float, unit=elem_unit(coord), copy=False)
```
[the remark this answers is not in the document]
nvaytet: I fixed the case where a pulse skipping chopper would have a phase offset of 180 deg by propagating multiple pulses through the chopper cascade to predict frame at detector location.

I fixed the potential issue of starting to record in the middle of a pulse-skipping frame by adding a `PulseStrideOffset` integer parameter, which can be set by the user if the data doesn't look right, and which is set to 0 by default.

src/scippneutron/tof/unwrap.py
```diff
@@ -224,563 +184,386 @@ def chopper_cascade_frames(
...
+    ----------
+    da:
+        Raw detector data loaded from a NeXus file, e.g., NXdetector containing
+        NXevent_data.
     """
     if da.bins is None:
         # Canonical name in NXmonitor
-        return PulseWrappedTimeOffset(da.coords['time_of_flight'])
-    return PulseWrappedTimeOffset(da.bins.coords['event_time_offset'])
+        toa = da.coords['time_of_flight']
```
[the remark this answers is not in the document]
nvaytet: The latter. It's just the name that is given in NeXus files.

jokasimr (approved):

jokasimr:
Looks good to me!

PR #587, opened 2024-12-12
Skip horace tests if pace_neutrons is not installed

nvaytet:
`pace_neutrons` is only available on pip, not conda.
During a conda build, we cannot install it, so we remove the dependency in the recipe from the test dependencies, and skip the tests if package is not installed.

### DIFF
--- a/tests/io/sqw/sqw_horace_test.py
+++ b/tests/io/sqw/sqw_horace_test.py
@@ -8,7 +8,6 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
-import pace_neutrons
 import pytest
 import scipp as sc
 from dateutil.parser import parse as parse_datetime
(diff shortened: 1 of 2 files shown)


tests/io/sqw/sqw_horace_test.py
```diff
@@ -25,9 +24,13 @@
     SqwLineProj,
 )
 
+pytest.importorskip("pace_neutrons")
```
jl-wynen, line 27:
```suggestion
pace_neutrons = pytest.importorskip("pace_neutrons")
```

tests/io/sqw/sqw_horace_test.py
```diff
@@ -25,9 +24,13 @@
...
+
 
 @pytest.fixture(scope="module")
 def matlab() -> Any:
+    import pace_neutrons
+
```
jl-wynen, line 33:
```suggestion
```

PR #588, opened 2025-01-13
Remove uses of attrs and meta and remove deprecated Mantid functions

jl-wynen:
This should fix the nightly at main build


src/scippneutron/mantid.py
```diff
@@ -1277,108 +973,6 @@ def load_with_mantid(
         )
 
 
-def load(
```
MridulS, line 1280: Could you also update the PR title so the changelog has information about removal of these deprecated functions?

tools/make_tutorial_data.ipynb
```diff
@@ -18,7 +18,7 @@
     "edges = sc.array(dims=['tof'], unit='us', values=np.linspace(5.0, 100000.0, num=201))\n",
     "data = sc.rebin(data, 'tof', edges)\n",
     "for i in [1, 2, 3, 4, 5]:\n",
-    "    mon = data.attrs[f'monitor{i}']\n",
+    "    mon = data.coords[f'monitor{i}']\n",
```
nvaytet, line 21: Does this mean that if we then release scipp and bump the version in the deps here, we will no longer be able to load the current tutorial data? (meaning that we have to update the tutorial data)

jl-wynen: No. We don't actually use the output file directly anywhere. We use a file called `loki-at-larmor-filtered.hdf5` but I don't know where it comes from. 
The docs seem to still work as before.

It turns out that the code has been broken for a long time because the mantid loader has changed. I fixed the first cell (and split it to make it easier to debug). But I cannot seem to run the next cell because I get a certificate error.

@SimonHeybrock Do you know if this notebook is still needed and how the filtered data was made?

jl-wynen: My point is that it looks like the file was processed further after creating it with this notebook because:

- it has a different name and
- the file contains a data group where the monitors are a group item, not a data array attr. The tutorial still works without recreating the files.

SimonHeybrock: Remove it, we can always recover from git history?

nvaytet (approved):

PR #589, opened 2025-01-13
Use Monte-Carlo simulations for frame unwrapping and WFM

nvaytet:
In this PR, we move away from our analytical computations of frame boundaries and fitting the polygons with single lines for frame unwrapping and WFM.

This method yielded errors typically towards the edge of the frames where the polygon fit would be less accurate (the lines would in some cases be outside of the polygons). Using a 3rd order fit helped but still wasn't optimal.

In this work, we instead use the `tof` package to illuminate the chopper cascade with a pulse of neutrons.
We then propagate the neutrons that exit the cascade to the distances of the pixels (this is just a single broadcast operation) and build a lookup table that gives us wavelength (or time-of-flight) as a function of distance and time-of-arrival.

The results are better than with the previous method (tested on a DREAM simulation), and the code simpler.
The computation is somewhat more expensive than before, but for now it is not very noticeable.

More importantly, this opens the door to other simulation methods. Instead of `tof`, one could in the future use McStas to make a much more realistic simulation of neutrons travelling through the cascade, including effects like non-instantaneous opening of choppers or neutron reflections along the beam guide.

See the `frame-unwrapping` and `wfm-time-of-flight` notebooks for a description of how the algorithm works.

**Comparison with previous version:**

Plots of dspacing vs 2theta:

- `true`: using the true wavelengths of the neutrons from the simulation data
- `from_tof`: the current version implemented in this PR
- `from_main`: the version on main using linear fit to polygons
- `from_cubic`: an (unpublished) improvement on main using a cubic fit to polygons (see https://github.com/scipp/scippneutron/tree/cubic-polygons)

Screenshot_20250114_141344

1D dspacing plot:

Figure 1 (16)
Figure 1 (17)


src/scippneutron/tof/fakes.py
```diff
@@ -19,6 +19,7 @@
 import scipp as sc
 from numpy import random
 
+from ..chopper import DiskChopper
 from . import chopper_cascade
```
nvaytet, line 25: I don't know if it is still useful to keep the other `FakeBeamline`, as well as the non-disk choppers below.
The refactoring of tests means they are no longer used in the tests, but they may still be useful in the future?

src/scippneutron/tof/unwrap.py
```diff
@@ -1,48 +1,86 @@
...
-from collections.abc import Callable, Mapping
+from collections.abc import Mapping
 from dataclasses import dataclass
+from functools import reduce
 from typing import NewType
 
 import numpy as np
 import scipp as sc
-from scipp.core.bins import Lookup
+import tof
```
SimonHeybrock, line 19: I have only briefly skimmed through this file so far. Consider how the use of `tof` in now deeply embedded in the entire file. At the same time, you already know that you want to use (also) McStas in place of `tof` in the future. Would it be possible to follow common design practices and define a minimal and well-defined interface/wrapper, such that the architecture is hopefully modular enough so you can swap out `tof` for `McStas` without rewriting all the logic again?

nvaytet: It's not that embedded actually, it's almost entirely in a single provider (`run_tof_model`).
All one needs to do to replace the table with a mcstas one is change the provider for `SimulationResults`.
But yes, we should move the `tof` stuff into a separate module, and make room for a mcstas provider.

I guess we would then have a set of `unwrap_basic_providers` (`tof`-based), and a set of `unwrap_advanced_providers` (`mcstas`-based)?

jokasimr:
Looks very interesting! I'll take a closer look today. 
One quick comment, I think we should think about replacing the bilinear interpolation on the lookup table with something that has fewer parameters and is multithreaded.

src/scippneutron/tof/unwrap.py
```diff
@@ -140,6 +146,19 @@ class TimeOfArrivalToTimeOfFlight:
...
+    """
+    Return the period of the source pulses, i.e., time between consecutive pulse starts.
+
+    Parameters
+    ----------
+    facility:
+        Facility where the experiment is performed (used to determine the source pulse
+        parameters).
+    """
+    return PulsePeriod(1.0 / tof.facilities[facility].frequency)
```
nvaytet, line 159: Should we also remove the use of `tof` here? It doesn't influence the lookup table.
But we could also keep our own library of frequencies in `scippneutron`...

SimonHeybrock: Yes.

nvaytet:
> I think we should think about replacing the bilinear interpolation on the lookup table with something that has fewer parameters and is multithreaded.

Multithreading would be nice. What parameters are you talking about? As far as I see it, there are no parameters?

jokasimr:
> What parameters are you talking about? As far as I see it, there are no parameters?

I meant the entries in the lookup table. But maybe it's already quite small?

nvaytet:
Conclusion from discussion: the `unwrap` module will move to `essreduce`.
All comments will be addressed there instead.
I will link the new PR once it's created.

nvaytet:
The PR in `essreduce` is here: https://github.com/scipp/essreduce/pull/163

nvaytet:
Pushed to the wrong repo. Update in `essreduce`.

PR #590, opened 2025-01-20
Update dependencies

jl-wynen:


nvaytet (approved):
Looks fine.

What was the reason/motivation for updating the deps?

jl-wynen:
> What was the reason/motivation for updating the deps?

The release failed because it pulled in the latest tof version: https://github.com/scipp/scippneutron/actions/runs/12863396700/job/35859898604 because we don't pin it.

nvaytet:
I don't understand. Where is it not pinned? It's pinned in `docs.txt`: https://github.com/scipp/scippneutron/blob/main/requirements/docs.txt#L232

Is it because we are using `conda`?

nvaytet:
Seems pinned in conda as well? https://github.com/scipp/scippneutron/blob/main/.buildconfig/ci-linux.yml#L42

PR #591, opened 2025-01-21
Add metadata utililites

### LINKED ISSUE #473: Metadata utilities

jl-wynen:
Fixes #473

This is a first implementation of common metadata models for data reduction. The models are based on an analysis of the file formats we care about: [Common Metadata Schemas.md](https://github.com/user-attachments/files/18492833/Common.Metadata.Schemas.md) (This is an updated version of the document in the issue: https://github.com/scipp/scippneutron/issues/473#issuecomment-1932134279).

The `ProcessLog` is missing because I don't feel confident that we have nailed down our approach to it sufficiently.

Please read the docstrings to make sure you understand the purpose of the different models to check whether this makes sense.

I used Pydantic to make it easy to implement validation (the ORCID type is a more sophisticated version of what I implemented in Scitacean.) and to help convert inputs to the expected types. But almost all fields are plain strings that need no validation. So I can change it to use dataclasses if you think that we should not take on Pydantic as a dependency.

I am going to open PRs in ESSreduce and ESSsans that use part of this new code. See the cross refs below.


src/scippneutron/io/cif.py
```diff
@@ -26,10 +26,15 @@
 Assemble data and metadata using a :class:`CIF` builder:
 
   >>> from scippneutron.io import cif
+  >>> from scippneutron import meta
```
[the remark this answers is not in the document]
jl-wynen: I wanted to keep it short. But I'm fine with metadata.

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,432 @@
...
+    and can usually be omitted when an ORCID iD is provided.
+    """
+
+    name: str
+    """Free form name of the person."""
+    orcid: ORCIDiD | None = None
+    """ORCID iD of the person."""
+
+    corresponding: bool = False
+    """Whether the person is the corresponding / contact author."""
```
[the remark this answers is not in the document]
SimonHeybrock: I don't have a strong preference, just wanted to make sure you had a reason. But multiple persons in an experiment could be corresponding authors for separate publications, so having it here may be problematic?

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,432 @@
...
+    """URL to the concrete version of the software.
+
+    If no URL for a concrete version is available,
+    a URL of the project or source code may be used.
+    """
+    doi: str | None = None
+    """DOI of the concrete version of the software.
+
+    If there is no DOI for the concrete version,
+    a general DOI for the software may be used.
```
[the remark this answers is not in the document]
jl-wynen: My intention was to use this for archiving and provenance. That is, to indicate which software was used to produce some data. For this, I don't see a need to link to some documentation.

But if you want to create a UI from this, then we should probably add the fields you mentioned.

In general, I think we need to somehow encode URLs or DOIs in our packages so we can extract them here. Or we need a hard coded list. But I don't see a good way for either to get the DOI for a specific release. Do you?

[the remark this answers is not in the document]
SimonHeybrock: See the dunder method I mention below?

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,432 @@
...
+            A Software instance.
+        """
+        from importlib.metadata import version
+
+        return cls(
+            name=package_name,
+            version=version(package_name),
+            url=_deduce_package_source_url(package_name),
+            doi=None,
+        )
```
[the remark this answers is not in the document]
jl-wynen: This is a larger discussion. So I opened https://github.com/orgs/scipp/discussions/3634

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,432 @@
...
+
+    The ESS source is provided as ``scippneutron.meta.ESS_SOURCE``.
+    """
+
+    # Needed to allow Scipp objects
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    frequency: SourceFrequency
+    """The source frequency in Hz."""
+    pulse_duration: PulseDuration
```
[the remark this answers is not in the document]
jl-wynen: True. I tried to be generic with the different source types and probes. But it currently does not look like we will support other types of sources any time soon. So we could specialise to pulsed neutron sources. Or keep the probe generic but limit ourselves to pulsed sources.
I think I prefer the latter. How about you?

src/scippneutron/meta/_model.py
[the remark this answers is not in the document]
jl-wynen: This model anyway does not encode what the person is associated with. They can be, e.g., and instrument scientist because the model was read from an input file. But that does not mean that they should be represented in the output file.
In general, I think info about people cannot be automated in a robust manner. So I see `Person` more as something a user inputs when creating an output file. And it may be behind a domain type like `Author` or `PrincipalInvestigator`. So maybe that is the solution: Rely on concrete domain types to encode the `corresponding` property, not a model field.

src/scippneutron/meta/_model.py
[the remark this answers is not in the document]
SimonHeybrock: 🤷

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,434 @@
...
+        return {
+            PulseDuration: self.pulse_duration,
+            SourceFrequency: self.frequency,
+            SourcePeriod: self.period,
+        }
+
+
+ESS_SOURCE = Source(
+    frequency=SourceFrequency(sc.scalar(14.0, unit='Hz')),
+    pulse_duration=PulseDuration(sc.scalar(0.003, unit='s')),
```
nvaytet, line 359: I guess how to define how long the pulse is is not so clear. The number that is commonly quoted is 2.86 ms.
But how much tail should we include?

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,434 @@
...
+    source_type=SourceType.SpallationNeutronSource,
```
nvaytet, line 360: Should we also have the wavelength range of the source? The same question about the tail would also apply for wavelengths...
ess-source

src/scippneutron/meta/_model.py
```diff
@@ -0,0 +1,434 @@
...
+    if instrument_name is None:
+        return None, None
+
+    match _FACILITY_PER_INSTRUMENT.get(instrument_name.lower()):
+        case None:
+            return None, None
+        case (facility, site):
+            return facility, site
+        case facility:
+            return facility, None
```
nvaytet, line 434: Shouldn't the site also be `'ESS'` for ESS instead of being `None`?

src/scippneutron/meta/_orcid.py
```diff
@@ -0,0 +1,119 @@
...
+                cls.__str__, info_arg=False, return_schema=core_schema.str_schema()
+            ),
+        )
+
+
+def _parse_pydantic(value: str | ORCIDiD) -> ORCIDiD:
+    return ORCIDiD(value)
+
+
+_ORCID_RESOLVER: str = 'https://orcid.org'
```
nvaytet, line 85: I don't really get why the url is part of the id?
Entries should just be `0000-0000-0001-0082`, without the url prefix?

tests/meta/meta_model_test.py
```diff
@@ -0,0 +1,53 @@
...
+import scippneutron as scn
+import scippneutron.data
+from scippneutron import meta
+
+
+def test_experiment_from_nexus_entry() -> None:
+    with snx.File(scn.data.get_path('PG3_4844_event.nxs')) as f:
+        experiment = meta.Measurement.from_nexus_entry(f['entry'])
+    assert experiment.title == 'diamond cw0.533 4.22e12 60Hz [10x30]'
+    assert experiment.run_number == '4844'
```
nvaytet, line 18: I don't know where would be the place to put it, but do we need a/several field(s) for all the parameters used in the experiment?
E.g. sample rotation angle, chopper info, slit openings, etc.

I'm guessing this is more the role of Scicat? Should we then have a link to the raw data on scicat or something?

src/scippneutron/meta/_orcid.py
```diff
@@ -0,0 +1,119 @@
...
+from __future__ import annotations
+
+import itertools
+from typing import Any
+
+from pydantic import GetCoreSchemaHandler
+from pydantic_core import core_schema
+
+
+class ORCIDiD:
```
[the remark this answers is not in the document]
jl-wynen: Yes, but the official docs use a double id: https://info.orcid.org/documentation/collecting-and-sharing-orcid-ids/ They distinguish between 'ORCID' the service and 'ORCID iD' the identifier.

src/scippneutron/meta/_model.py
jl-wynen: Good point. This is what I got from ESSspectroscopy. The length might depend on the specific use case. It may be best to stick strictly to metadata here and not specify duration or frequency.

src/scippneutron/meta/_model.py
jl-wynen: See my reply above.

src/scippneutron/meta/_model.py
jl-wynen: Could be. The distinction only exists in SciCat and even there it's not strict. `creationLocation` is supposed to be `site:facility:instrument`. But I have been using `ESS:DMSC:instrument` for simulated data. This would work best if we have `site="ESS"`.

src/scippneutron/meta/_orcid.py
jl-wynen: See https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier
With these kinds of ids (same for DOI) it is generally recommended to encode them as a complete URI.

Note that this accepts inputs of the plain id without the resolver. But it adds the resolver when storing the id.

src/scippneutron/meta/_model.py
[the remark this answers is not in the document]
jl-wynen: I removed the source parameters because Neil noted that they are not so easy to define. Is the name still an issue?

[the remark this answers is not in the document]
SimonHeybrock: Ok for me. Names can always be changed later.

src/scippneutron/metadata/_model.py
```diff
@@ -0,0 +1,405 @@
...
+    Specify an ORCID iD whenever possible to make sure the
+    person can be uniquely identified.
+    The name is always required but is often not unique.
+    Other contact information like address or email address are less important
+    and can usually be omitted when an ORCID iD is provided.
+    """
+
+    name: str
+    """Free form name of the person."""
+    orcid: ORCIDiD | None = None
```
nvaytet, line 169: I don't know if it's now confusing that the class is ORCIDiD, but the member is `orcid`. Should it be `orcid_id`?
Cf. your comment:
> They distinguish between 'ORCID' the service and 'ORCID iD' the identifier.

src/scippneutron/metadata/_model.py
```diff
@@ -0,0 +1,405 @@
...
+
+    try:
+        return next(
+            url.split(',')[-1].strip() for url in urls if url.startswith("Source")
+        )
+    except StopIteration:
+        return None
+
+
+class SourceType(str, enum.Enum):
```
nvaytet, line 294: A little annoying that the entire `str` API ends up in the generated docs (also for `RadiationProbe` below), but don't know what can be done about it...
Screenshot_20250127_112156

jl-wynen: Agree. I changed it to only inherit from `Enum`. This should not affect usability because pydantic converts strings to enums anyway.

I also tried `enum.StrEnum` instead of inheriting from `str` directly. But it leads to the same docs.

src/scippneutron/metadata/_model.py
jl-wynen: Agree. I got lazy.

SimonHeybrock (approved):
Looks good for a v1 to me, provided that @nvaytet is also happy.

nvaytet (approved):

PR #592, opened 2025-01-22
Remove frame unwrapping and WFM code

nvaytet:
This code is moving to `essreduce`: see https://github.com/scipp/essreduce/pull/163


nvaytet:
Don't know if we want to keep the `fakes`. Right now they were only used in the unwrap/wfm tests which are gone.

SimonHeybrock:
> Don't know if we want to keep the `fakes`. Right now they were only used in the unwrap/wfm tests which are gone.

I think they can be removed? Where they also added in `essreduce`?

SimonHeybrock (approved):

nvaytet:
> I think they can be removed? Where they also added in `essreduce`?

Not exactly. The `FakeBeamlineEss` was moved over, but not the `FakeBeamline` you initially created. Also, the choppers were not moved, I made some new ones which are using `DiskChopper` instead of `tof.chopper_cascade.Chopper`.

SimonHeybrock:
Do I understand correctly that your last commit *does* remove them now?

nvaytet:
Yes, I removed them.
I'm of the opinion that it's ok to remove, but I thought we could revert if you prefered to keep them.

SimonHeybrock (approved):

nvaytet:
I think I forgot to remove the dependency on `tof`...

Issue #594, opened 2025-01-27
Slow import times


SimonHeybrock:
ScippNeutron is taking a long time to import. It seems related to imports of plotting libraries.

- Avoid these imports, either directly or by not importing the problematic submodules (`masking` might be the or one of the offenders).
- Consider SPEC 1 scipp/scipp#3567 as a potential solution.

Note: You can profile imports using
```bash
python -X importtime -c 'import scippneutron'
```

PR #595, opened 2025-01-27
 MAINT: Use lazy_loader for lazy loading of submodules 

MridulS:
This adds lazy loading to scippneutron, we should ideally also add this to plopp to propogate it across the ess ecosystem. Tox will not be happy here as I haven't run tox -e deps locally (libpymcr doesn't do ARM wheels).

Locally I can see a 100x speedup in import time (600ms vs 6ms) with cached modules in an env, in a freshly installed env it's ~10seconds for scippneutron.

This should fix https://github.com/scipp/scippneutron/issues/594

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ dynamic = ["version"]
 # Run 'tox -e deps' after making changes here. This will update requirement files.
 # Make sure to list one dependency per line.
 dependencies = [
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
LGTM. Merging into branch `lazy` so I can more easily update the deps (since you are working with a fork).

PR #596, opened 2025-01-28
MAINT: Use lazy_loader for lazy loading of submodules

### LINKED ISSUE #594: Slow import times

SimonHeybrock:
This is @MridulS change from #595 + `tox -e deps`.

Fixes #594.


pyproject.toml
```diff
@@ -31,6 +31,7 @@ dynamic = ["version"]
 # Run 'tox -e deps' after making changes here. This will update requirement files.
 # Make sure to list one dependency per line.
 dependencies = [
+    "lazy_loader",
```
nvaytet, line 34: I think Mantid tests are failing because you also need to add `lazy_loader` to the `.buildconfig` file?

SimonHeybrock:
Thanks @MridulS, latest changes LGTM.

PR #604, opened 2025-02-20
Convert metadata fields from sc.Variable

jl-wynen:
When we read metadata from files, values may be wrapped in `sc.Variable`s. This tripped up the build in https://github.com/scipp/esssans/pull/195. It is likely not a big problem when we use scippnexus-based loaders (the linked PR failed for ISIS data). But this PR introduces a mechanism that allows us to extract data from scalar variables so that we need to worry less about the precise input types.

### DIFF
--- a/src/scippneutron/metadata/_model.py
+++ b/src/scippneutron/metadata/_model.py
@@ -4,12 +4,25 @@ from __future__ import annotations
 
 import enum
 from datetime import datetime
+from typing import Annotated, Any
 
+import scipp as sc
 import scippnexus as snx
 from dateutil.parser import parse as parse_datetime
-from pydantic import BaseModel, EmailStr
+from pydantic import BaseModel, BeforeValidator, EmailStr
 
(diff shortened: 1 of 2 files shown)


src/scippneutron/metadata/_model.py
```diff
@@ -39,13 +52,13 @@ class Beamline(BaseModel):
     which version of the beamline was used.
     """
 
-    name: str
+    name: Annotated[str, BeforeValidator(_unpack_variable)]
```
MridulS, line 55: I'm not sure of the best practices here but does it make sense to define type aliases here?
```
ValidatedStr = Annotated[str, BeforeValidator(_unpack_variable)]
ValidatedStrOptional = Annotated[str | None, BeforeValidator(_unpack_variable)]
```

jl-wynen: Could do. I would say it depends on what we think is more readable. I'm fine either way.

MridulS (approved):

PR #613, opened 2025-03-20
Add atomic weights and masses

### LINKED ISSUE #612: Consider adding atomic masses
For PDF reduction, we need the masses of atoms. Scippneutron provides a number of neutron-scattering-related parameters of atoms. We may want to add masses to this list. See, e.g., https://gitlab.com/ase/ase/-/blob/master/ase/data/__init__.py?ref_type=heads#L82

If we don't do this, people could use, e.g., [ASE](https://wiki.fysik.dtu.dk/ase/index.html) to get the masses. But this is a rather large package just for getting some numbers.

jl-wynen:
Fixes #612

The exact choice of which value to use is complicated. But for most applications, going with the standard atomic weight should be fine.


src/scippneutron/atoms/__init__.py
```diff
@@ -1,17 +1,98 @@
...
+    @property
+    def atomic_weight(self) -> sc.Variable:
+        """Return the atomic weight is available."""
+        if self._atomic_weight is None:
+            raise ValueError(
+                f"Atomic weight for '{self.isotope}' is not defined ."
+                "This likely means that there is no standard atomic weight "
+                "for this element."
+            )
+        return self._atomic_weight
```
SimonHeybrock, line 45: Are we worried that the caller might modify this in-place, for subtle errors? Should we return a copy?

I think we have a similar in the scipp physical constants.

src/scippneutron/atoms/__init__.py
```diff
@@ -84,37 +163,57 @@ def for_isotope(isotope: str) -> ScatteringParams:
...
-            line[10], line[11], 'barn'
-        ),
-        total_scattering_cross_section=_assemble_scalar(line[12], line[13], 'barn'),
-        absorption_cross_section=_assemble_scalar(line[14], line[15], 'barn'),
-    )
+def _load_atomic_weight(element: str) -> tuple[int, sc.Variable | None]:
+    with _open_bundled_parameters_file('atomic_weights.csv') as f:
+        if line_remainder := _find_line_with_isotope(element, f):
+            z, weight, error = line_remainder.rstrip().split(',')
+            return int(z), _assemble_scalar(weight, error, 'Da')
```
SimonHeybrock, line 207: What is `Da`?

tools/atomic_weights.ipynb
```diff
@@ -0,0 +1,238 @@
+{
```
SimonHeybrock, line 1: No one might remember this notebooks exists. Can we add a reference to it in the code or the csv files, so future devs who try updating the tables can find it?

src/scippneutron/atoms/__init__.py
jl-wynen: [Dalton](https://en.wikipedia.org/wiki/Dalton_(unit)) it's the weight relative to 12C.

src/scippneutron/atoms/__init__.py
SimonHeybrock: Not sure. Another alternative might be to make the variables read-only, but I think that mechanism does not have Python bindings.

jl-wynen: Event if it was available in Python, people could still set the `values` directly.

PR #624, opened 2025-05-21
Bump deps in mantid test env

MridulS:
Not sure what's going wrong with https://github.com/scipp/scippneutron/pull/623, but it looks like some packages haven't been bumped in a while in the env used for running mantid tests.


nvaytet:
Yes, strange about the build failure... I don't understand either

jl-wynen:
I remember why I didn't update the versions in the past, it makes the docs build hang forever. I don't know which dependency this is specifically. And I can't reproduce it locally.

PR #636, opened 2025-07-28
Lower bounds

jokasimr:


docs/conf.py
```diff
@@ -33,6 +33,7 @@
     'sphinx_copybutton',
     'sphinx_design',
     'sphinxcontrib.bibtex',
+    'sphinxcontrib.autodoc_pydantic',
```
MridulS, line 36: The doc build workflow uses the mamba env in .buildconfig/ci-linux.yaml so we need to add the new deps in https://github.com/scipp/scippneutron/blob/main/.buildconfig/ci-linux.yml

.buildconfig/ci-linux.yml
```diff
@@ -43,6 +43,7 @@ dependencies:
   - sphinx-copybutton==0.5.2
   - sphinx-design==0.6.1
   - sphinxcontrib-bibtex==2.6.3
+  - autodoc-pydantic==2.1.0
```
MridulS, line 46: We should ideally sync this with the pinned versions in `requirements/*.txt` files. We should probably automate this.

jl-wynen:
Can we move this PR forward without figuring out how to automatically handle conda bounds? I had to roll out the copier update to drop support for Python 3.10 and now we have a failing nightly test for lower bounds.

PR #641, opened 2025-08-07
Update lower bound of ipykernel

jl-wynen:
This should fix the lower bound nightly test.


jl-wynen:
I am struggling to reproduce the error locally now, investigating.

SimonHeybrock (approved):

PR #651, opened 2025-11-21
Changes to SQW writer for BIFROST

jl-wynen:
The most notable changes here are
- support for writing DND arrays instead of just zeros,
- pixel data is not specified as a numpy array instead of a data array to not rely on coord naming conventions.

### DIFF
--- a/src/scippneutron/io/sqw/_sqw.py
+++ b/src/scippneutron/io/sqw/_sqw.py
@@ -435,12 +435,18 @@ def _parse_single_ix_experiment_3_0(struct: ir.Struct) -> SqwIXExperiment:
         (e,) = candidate_efix
         efix = sc.scalar(e.value, unit="meV")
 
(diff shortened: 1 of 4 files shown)


src/scippneutron/io/sqw/_ir.py
```diff
@@ -150,5 +150,7 @@ def _serialize_field(
             ty=field.ty, shape=(len(field.value),) if field.value else (), data=[field]
         )
     if isinstance(field, Array):
-        return ObjectArray(ty=field.ty, shape=field.value.shape[::-1], data=field.value)
+        return ObjectArray(
+            ty=field.ty, shape=field.value.shape[::-1], data=field.value.T
```
SimonHeybrock: `shape[::-1]` seems to make sense with `T`, does this mean it was broken before?

PR #653, opened 2025-11-24
Finish SQW docs

jl-wynen:


docs/developer/file-formats/sqw.md
```diff
@@ -7,14 +7,16 @@ ScippNeutron can write and, to a limited degree, read SQW files using {mod}`scip
 
 ```{note}
 There is no official specification for the format of SQW files.
-The contents of this document are based the little bits of documentation that are available, from reading the source code of Horace, and from experimenting with Horace and SQW files.
-It should *not* be regarded as authoritative.
+The contents of this document are based on what documentation is available, on reading the source code of Horace, and on experiments with Horace and SQW files.
+This document should *not* be regarded as authoritative.
 ```
 
+See, e.g., https://github.com/pace-neutrons/Horace/blob/3c00958d578c625a7f2274ef7b34557578abb934/documentation/add/07-sqw_redesign.md for some documentation provided by the developers of Horace.
```
nvaytet: This url is not rendering properly in sphinx
Screenshot_20251125_125448

src/scippneutron/io/sqw/_build.py
```diff
@@ -175,6 +175,28 @@ def add_empty_dnd_data(self, metadata: SqwDndMetadata) -> SqwBuilder:
...
+            Metadata describing the DND data.
+        data:
+            The values and variances of the image.
+        counts:
+            The number of observations ('pixels') per bin.
+
+        Returns
+        -------
+        :
+            Self.
```
nvaytet: Seems to not return self?

PR #656, opened 2025-11-26
Ignore links to sciencedirect

jl-wynen:
These links are now forbidden for bots.


MridulS (approved):
Maybe add this to the template too?

Issue #669, opened 2026-02-10
Weekly osx builds fail


nvaytet:
See https://github.com/scipp/scippneutron/actions/workflows/weekly_windows_macos.yml

The failed tests are
```
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle0-axis1-base0-radius1-height0] FAILED [ 24%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle0-axis1-base1-radius1-height0] FAILED [ 25%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle1-axis1-base0-radius1-height0] FAILED [ 25%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle1-axis1-base1-radius1-height0] FAILED [ 26%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle2-axis1-base0-radius1-height0] FAILED [ 27%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle2-axis1-base1-radius1-height0] FAILED [ 27%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle3-axis1-base0-radius1-height0] FAILED [ 28%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle3-axis1-base1-radius1-height0] FAILED [ 28%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle4-axis1-base0-radius1-height0] FAILED [ 29%]
tests/absorption/cylinder_test.py::test_intersection_from_center[point_on_unit_circle4-axis1-base1-radius1-height0] FAILED [ 29%]
tests/io/sqw/sqw_horace_test.py::test_horace_roundtrip_main_header FAILED [ 85%]
```
Seems rather random which tests fail?

It is consistent though; it's always the same tests that fail (I think?).

PR #671, opened 2026-02-11
Fix cylinder intersection failures on ARM64, test for almost equality

MridulS:
fixes https://github.com/scipp/scippneutron/issues/669


jl-wynen:
> > The macos tests are fixed: https://github.com/scipp/scippneutron/actions/runs/21905945879?pr=671
> 
> The osx tests passed, but the windows tests failed...

Is that still the case? I don't see the failed test

PR #691, opened 2026-04-14
New instrument view with better performance

### LINKED ISSUE #484: Fix instrument view to allow multiple inputs
We need to support things like
```Py
scn.instrument_view({'layer0': res['layer', 0].hist(),
              'layer3': res['layer', 3].hist()},
             pixel_size=0.0075)
```
It currently fails saying `dict` has no attribute `coords`.

nvaytet:
Using the [Dream instrument view](https://github.com/scipp/essdiffraction/pull/242) to be the generic one in scippneutron:

- Better performance
- Allows multiple banks as input (in a dict or datagroup)
- Adds a slider for wavelength/tof dimension

We remove the ability to show other components using pythreejs objects like cylinders or disks; this was basically never used.

Needs https://github.com/scipp/plopp/pull/553

Fixes #484


nvaytet:
We need https://github.com/scipp/plopp/pull/554 for the notebooks to work properly again.

src/scippneutron/instrument_view.py
```diff
@@ -1,290 +1,123 @@
...
-    * `center` - scipp scalar vector describing position to place item at.
-    * `size` - scipp scalar vector describing the bounding box to use in the
-    same length units as positions
-    * `type` - known shape type to use.
-    Valid types are: 'box', 'cylinder' or 'disk'.
-
-    Optional arguments are:
+) -> FigureLike:
+    """
+    Three-dimensional visualization of the DREAM instrument.
```
jokasimr: Leftover docs comment from DREAM instrument view

src/scippneutron/instrument_view.py
```diff
@@ -1,290 +1,123 @@
...
-        Key for coord/attr holding positions to use for pixels.
+    data:
+        Data to visualize. The data can be a single detector module (``DataArray``),
+        or a group of detector modules (``dict`` or ``DataGroup``).
+        The data must contain a ``position`` coordinate.
+    dim:
+        Dimension to use for the slider. No slider will be shown if this is None.
+    pos:
+        The name of the coordinate to use for the pixel positions.
     pixel_size:
```
jokasimr: Why is there both a `size` and a `pixel_size` argument? Is it backwards compatibility?
Why is only `pixel_size` documented in the docstring?

src/scippneutron/instrument_view.py
```diff
@@ -1,290 +1,123 @@
...
+    if isinstance(data, sc.DataArray):
+        data = sc.DataGroup({"": data})
+    pieces = []
+    for da in data.values():
+        da = da.drop_coords(list(set(da.coords) - {pos, dim}))
+        position_dims = tuple(da.coords[pos].dims)
+        if dim is not None:
+            # Ensure that the dims to be flattened are contiguous
+            other_dims = tuple(d for d in da.dims if d not in position_dims)
+            da = da.transpose(position_dims + other_dims)
```
jokasimr: Is there a reason why it is transposed in this order?
I'd imagine it would be faster to access a selective "time-slice" if `time` was the outermost dimension.

src/scippneutron/instrument_view.py
nvaytet: It's backward compatibility, and I guess I should only document `size` :+1:

src/scippneutron/instrument_view.py
nvaytet: I can never remember the order (I blame it on too much fortran in my previous life) if it's faster to slice `y` or `x` if your dims are `(y, x)`...
But yes, the idea would be to make it the faster index

jokasimr: I'm not 100% sure either. Might depend on if we read a small time slice or a large time slice.

My intuition is that if we have time as the outer dim then we can load a relatively small contigous chunk of data into the cpu, but otherwise we must load a number of smaller intervals of data over the entire data array, and that seems slower.

I did this quick benchmark and it seems to confirm that:
```
import numpy as np
a = np.random.random((100, 1000_000))
b = np.random.random((1000_000, 100))

# Selective "point" time

# Select random time point and sum outer time dimension
%timeit a[np.random.randint(0, 99)].sum(0)
339 μs ± 4.56 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

# Select random time point and sum inner time dimension
%timeit b[..., np.random.randint(0, 99)].sum(-1)
5.35 ms ± 46.9 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)


# Non-selective "broad" time range

# Select a random slice and sum outer "time" dimension
%timeit a[slice(*sorted(np.random.randint(0, 99, size=2)))].sum(0)
15 ms ± 656 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# Select a random slice and sum inner "time" dimension
%timeit b[..., slice(*sorted(np.random.randint(0, 99, size=2)))].sum(-1)
55.4 ms ± 8.52 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

jokasimr (approved):

PR #699, opened 2026-04-30
Increase scipp lower bound

nvaytet:
To fix the nightly lower-bound builds.

See https://github.com/scipp/plopp/pull/560 for context.

Successful lower bounds run [here](https://github.com/scipp/scippneutron/actions/runs/25158373530).


pyproject.toml
```diff
@@ -39,13 +39,13 @@ dependencies = [
     "numpy>=1.20",
     "plopp>=26.4.1",
     "pydantic>=2",
-    "scipp>=24.07.0",
+    "scipp>=25.8.0",
```
SimonHeybrock, line 42: Why this particular new minimum?

nvaytet: Because that is what Plopp requires. See [here](https://github.com/scipp/plopp/blob/main/pyproject.toml#L40) and [discussion](https://github.com/scipp/plopp/pull/560#issuecomment-4350845904).

I'd like to make Scipp a hard dependency of Plopp to avoid such issues in the future.

SimonHeybrock: But why should the scippneutron minimum depend on what Plopp needs?

But on the other hand I don't see a reason why we should support older versions?

SimonHeybrock (approved):

pyproject.toml
nvaytet: Basically, I think if we make scipp a hard dependency of plopp, and require `plopp>=26.4.1`, it will figure out all by itself that it cannot install an older version of scipp even if our lower bound is below what plopp needs.

I think this means that with our current setup in Plopp with scipp being optional, any minimum version of scipp we request in Plopp's pyproject.toml would be ignored if you only install `plopp` and not `plopp[scipp]`.
Meaning that you can end up with a very broken environment like installing `plopp=26.4.1` and `scipp==0.0.4` (maybe that is going too far) if they allow the same python version...

PR #700, opened 2026-05-07
Support multiple pulses in chopper cascade

nvaytet:
Example:
```Py
from scippneutron.tof import chopper_cascade
from ess.odin.beamline import choppers
import scipp as sc

source_position = sc.vector([0, 0, 0], unit='m')
disk_choppers = choppers(source_position)

pulse_frequency = sc.scalar(14.0, unit='Hz')
pulse_period = 1 / pulse_frequency

chops = {
    key: chopper_cascade.Chopper(
        distance=(ch.axle_position - source_position).fields.z,
        time_open=ch.time_offset_open(pulse_frequency=pulse_frequency / 4),
        time_close=ch.time_offset_close(pulse_frequency=pulse_frequency / 4),
    )
    for key, ch in disk_choppers.items()
}

frames = chopper_cascade.FrameSequence.from_source_pulse(
    time_min=sc.scalar(0.0, unit='ms'),
    time_max=sc.scalar(4.0, unit='ms'),
    wavelength_min=sc.scalar(0.0, unit='angstrom'),
    wavelength_max=sc.scalar(10.0, unit='angstrom'),
    npulses=3,
    pulse_period=pulse_period,
)
frames = frames.chop(chops.values())

at_sample = frames.propagate_to(sc.scalar(65.0, unit="m"))
at_sample.draw()
```
Figure 3 (1)

### DIFF
--- a/src/scippneutron/tof/chopper_cascade.py
+++ b/src/scippneutron/tof/chopper_cascade.py
@@ -298,6 +298,9 @@ class FrameSequence:
         time_max: sc.Variable,
         wavelength_min: sc.Variable,
         wavelength_max: sc.Variable,
+        npulses: int = 1,
+        pulse_period: sc.Variable | None = None,
+        distance: sc.Variable | None = None,
     ):
         """
         Initialize a frame sequence from min/max time and wavelength of a pulse.
@@ -311,10 +314,21 @@ class FrameSequence:
             [wavelength_min, wavelength_min, wavelength_max, wavelength_max],
             dim='vertex',
         ).to(unit='angstrom')
+
+        if npulses > 1:
+            if pulse_period is None:
+                raise ValueError('pulse_period must be provided if npulses > 1')
+            subframes = [
+                Subframe(time=time + i * pulse_period, wavelength=wavelength)
+                for i in range(npulses)
+            ]
+        else:
+            subframes = [Subframe(time=time, wavelength=wavelength)]
+
         frames = [
             Frame(
-                distance=sc.scalar(0, unit='m'),
-                subframes=[Subframe(time=time, wavelength=wavelength)],
+                distance=sc.scalar(0.0, unit='m') if distance is None else distance,
+                subframes=subframes,
             )
         ]
         return FrameSequence(frames)
--- a/src/scippneutron/tof/chopper_cascade.py
+++ b/src/scippneutron/tof/chopper_cascade.py
@@ -318,6 +318,7 @@ class FrameSequence:
         if npulses > 1:
             if pulse_period is None:
                 raise ValueError('pulse_period must be provided if npulses > 1')
+            pulse_period = pulse_period.to(unit='s')
             subframes = [
                 Subframe(time=time + i * pulse_period, wavelength=wavelength)
                 for i in range(npulses)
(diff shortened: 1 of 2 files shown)


src/scippneutron/tof/chopper_cascade.py
```diff
@@ -311,10 +314,22 @@ def from_source_pulse(
...
         ).to(unit='angstrom')
+
+        if npulses > 1:
+            if pulse_period is None:
+                raise ValueError('pulse_period must be provided if npulses > 1')
+            pulse_period = pulse_period.to(unit='s')
+            subframes = [
+                Subframe(time=time + i * pulse_period, wavelength=wavelength)
+                for i in range(npulses)
+            ]
```
SimonHeybrock, line 325: Should there be source pulses also *before* the "pulse of interest? Or is that handled by some wrapping you do afterwards?

tests/tof/chopper_cascade_test.py
```diff
@@ -493,3 +493,67 @@ def test_frame_sequence_getitem_by_distance_selects_correct_chopper(
...
+    expected_time = sc.array(dims=['vertex'], values=[0.0, 1.0, 1.0, 0.0], unit='ms')
+    expected_wavelength = sc.array(
+        dims=['vertex'], values=[1.0, 1.0, 10.0, 10.0], unit='angstrom'
+    )
+
+    for i in range(3):
+        expected_subframe = chopper_cascade.Subframe(
+            time=expected_time + i * pulse_period.to(unit='ms'),
+            wavelength=expected_wavelength,
+        )
```
SimonHeybrock, line 519: Not a fan of repeating the implementation-math in the test, but not sure what a better test would be.

src/scippneutron/tof/chopper_cascade.py
nvaytet: I do some wrapping manually afterwards in `essreduce`. I thought I would keep it simple here, so that it behaves a little like `tof`. But I don't mind either way.

tests/tof/chopper_cascade_test.py
nvaytet: I could check that the min of `frame.bounds().time` is zero, and the max is `2 * period + 1ms`?
And remove the looping over the frames. Not sure it's better, but it's different.

src/scippneutron/tof/chopper_cascade.py
SimonHeybrock: Its fine as it is, probably. But I do wonder if one could formulate this on a cyclic coord system instead?

nvaytet: Maybe something like making the period boundary act like a chopper and chop the polygons as well?

SimonHeybrock: What I mean: After every time propagation, do a `mod 71 ms` or (142 ms if pulse-skipping, etc.). But the polygon math might not be easy to port.

nvaytet: Yes I got that. You'd have to do the mod, but as you say you would run into cases where some polygons span across the period boundary.
I was thinking one easy way to split those polygons would be to use the `chop` mechanism?
You would then of course have to duplicate the pulses to the frame before the current frame.

SimonHeybrock: I don't think pulse duplication would be necessary any more, since periodicity takes care of that? Splitting the polygons that cross the boundary is a good suggestion, provided it does not interfere with other code that tries to look at subframes?

SimonHeybrock (approved):

PR #701, opened 2026-05-18
Use is_binned property instead of bins is [not] None

jl-wynen:
Working towards https://github.com/scipp/scipp/issues/3790

### DIFF
--- a/src/scippneutron/_utils/__init__.py
+++ b/src/scippneutron/_utils/__init__.py
@@ -10,11 +10,11 @@ from scipp.typing import VariableLike
 
 
 def elem_unit(var: VariableLike) -> sc.Unit:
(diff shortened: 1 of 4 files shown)


src/scippneutron/_utils/__init__.py
```diff
@@ -10,11 +10,11 @@
 
 
 def elem_unit(var: VariableLike) -> sc.Unit:
-    return var.bins.unit if var.bins is not None else var.unit
```
nvaytet, line 13: Did you also grep for `bins is None`?

jl-wynen: I searched for `.bins is` and there only seem to be these negative checks.

nvaytet (approved):

PR #711, opened 2026-06-23
copier update

jokasimr:
Not sure if the changes to `.github/workflows/docs.yml` are correct or if there are some special cases for `scippneutron` that I removed now that should have been kept.

There was a large diff from the copier template, I erred on the side of removing inconsistencies with the template to make future maintenance easier.


tox.ini
```diff
@@ -23,6 +23,28 @@ deps =
...
+deps =
+  scippneutron=={posargs}
+  {[testenv:docs]deps}
+allowlist_externals={[testenv:docs]allowlist_externals}
+commands = {[testenv:docs]commands}
+
+[testenv:linkcheck]
+description = Run Sphinx linkcheck
+deps = -r requirements/docs.txt
+commands = python -m sphinx -j2 -v -b linkcheck -d {toxworkdir}/docs_doctrees -D nbsphinx_execute=never docs html
```
jl-wynen: These can all be removed. We can't build the docs with tox because they need a conda env.

.github/workflows/docs.yml
```diff
@@ -53,19 +53,18 @@ jobs:
...
-          cache-environment: true
-          create-args: >-
-            python=3.11
-            scippneutron==${{ inputs.version }}
+          python-version-file: '.github/workflows/python-version-ci'
+      - run: python -m pip install --upgrade pip
+      - run: python -m pip install -r requirements/ci.txt
+      - run: tox -e releasedocs -- "${VERSION}"
+        if: ${{ inputs.version != '' }}
       - uses: mamba-org/setup-micromamba@v2
```
jl-wynen, line 69: The update only changed one branch of this if-else. But CI is using the other branch that still uses mamba and mantid. So CI should pass. But the changed branch should fail because it doesn't have mantid.

I doubt we can make this work in the same way as the other packages. I'd opt for reverting docs.yml and tox.ini. You'd need to check release.yml to see if it works.

jl-wynen (changes requested):

jl-wynen (approved):

PR #718, opened 2026-08-04
feat: relative width smoothing procedures

jokasimr:
Adds functionality to scippneutron for smoothing 1D curves with kernels that have a width relative to the dimension coordinate.

As an example, in the figure below the step `signal` is smoothed by a gaussian kernel with width proportional to `x`:

Figure 49

The computed `smoothed` signal and the expected `exact` signal overlap.

### DIFF
--- a/src/scippneutron/smoothing.py
+++ b/src/scippneutron/smoothing.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from numbers import Real
 from typing import Any, Protocol, TypeAlias, cast
 
 import numpy as np
@@ -12,14 +13,8 @@ from scipy.signal import convolve
 from scipy.stats import norm, triang, uniform
 
 __all__ = [
-    "smooth_gaussian",
-    "smooth_kernel",
-    "smooth_rectangle",
-    "smooth_relative_gaussian",
-    "smooth_relative_kernel",
-    "smooth_relative_rectangle",
-    "smooth_relative_triangle",
-    "smooth_triangle",
+    "smooth",
+    "smooth_relative",
 ]
 
 
@@ -44,6 +39,7 @@ _BUILTIN_KERNELS: dict[str, _Distribution] = {
     "rectangle": uniform(loc=-1.0, scale=2.0),
     "rect": uniform(loc=-1.0, scale=2.0),
     "box": uniform(loc=-1.0, scale=2.0),
+    "boxcar": uniform(loc=-1.0, scale=2.0),
     "uniform": uniform(loc=-1.0, scale=2.0),
     # Centered triangle on [-1, 1], peak at 0.
     # Users can pass triang(c=...) themselves for asymmetric triangles.
@@ -102,7 +98,7 @@ def _trim_kernel_weights(
 
 def _relative_kernel_weights(
     log_spacing: float,
-    alpha: float,
+    scale: float,
     kernel: _Kernel,
     tail: float,
     max_offset: int,
@@ -112,16 +108,16 @@ def _relative_kernel_weights(
 
     The kernel distribution describes the relative displacement Z:
 
-        q' = q * (1 + alpha * Z)
+        q' = q * (1 + scale * Z)
 
     Equivalently,
 
-        K(q, q') = 1 / (alpha * q) * f((q' - q) / (alpha * q))
+        K(q, q') = 1 / (scale * q) * f((q' - q) / (scale * q))
 
     where f is the PDF of the supplied distribution.
     """
-    if not np.isfinite(alpha) or alpha <= 0:
-        raise ValueError("alpha must be positive")
+    if not np.isfinite(scale) or scale <= 0:
+        raise ValueError("scale must be positive")
     if not np.isfinite(log_spacing) or log_spacing <= 0:
         raise ValueError("log_spacing must be positive")
     if not np.isfinite(tail) or not (0.0 < tail < 1.0):
@@ -134,15 +130,15 @@ def _relative_kernel_weights(
     # Positive physical domain:
     #
     #   q' > 0
-    #   q * (1 + alpha * z) > 0
-    #   z > -1 / alpha
-    z_domain_min = -1.0 / alpha
+    #   q * (1 + scale * z) > 0
+    #   z > -1 / scale
+    z_domain_min = -1.0 / scale
 
     p_domain_min = float(dist.cdf(z_domain_min))
     norm_mass = 1.0 - p_domain_min
 
     if not np.isfinite(norm_mass) or norm_mass <= 0.0:
-        raise ValueError("kernel has no positive-domain mass for this alpha")
+        raise ValueError("kernel has no positive-domain mass for this scale")
 
     support_min, support_max = _kernel_support(dist)
 
@@ -155,20 +151,20 @@ def _relative_kernel_weights(
     has_finite_log_support = (
         np.isfinite(z_left_exact)
         and np.isfinite(z_right_exact)
-        and (1.0 + alpha * z_left_exact > 0.0)
+        and (1.0 + scale * z_left_exact > 0.0)
     )
 
     if has_finite_log_support:
-        u_left = np.log1p(alpha * z_left_exact)
-        u_right = np.log1p(alpha * z_right_exact)
+        u_left = np.log1p(scale * z_left_exact)
+        u_right = np.log1p(scale * z_right_exact)
     else:
         probabilities = (
             p_domain_min + np.array([0.5 * tail, 1.0 - 0.5 * tail]) * norm_mass
         )
         z_left, z_right = (float(value) for value in dist.ppf(probabilities))
 
-        u_left = np.log1p(alpha * z_left)
-        u_right = np.log1p(alpha * z_right)
+        u_left = np.log1p(scale * z_left)
+        u_right = np.log1p(scale * z_right)
 
     if not np.isfinite(u_right):
         raise ValueError("right kernel bound is not finite; increase tail")
@@ -184,10 +180,10 @@ def _relative_kernel_weights(
     L = (m - 0.5) * log_spacing
     U = (m + 0.5) * log_spacing
 
-    # z = (q' - q) / (alpha q)
-    #   = (exp(u) - 1) / alpha
-    zL = np.expm1(L) / alpha
-    zU = np.expm1(U) / alpha
+    # z = (q' - q) / (scale q)
+    #   = (exp(u) - 1) / scale
+    zL = np.expm1(L) / scale
+    zU = np.expm1(U) / scale
 
     # Exact cell-integrated weights in log-space.
     w = (dist.cdf(zU) - dist.cdf(zL)) / norm_mass
@@ -198,13 +194,13 @@ def _relative_kernel_weights(
 
 def _translation_invariant_kernel_weights(
     spacing: float,
-    width: float,
+    scale: float,
     kernel: _Kernel,
     tail: float,
     max_offset: int,
 ) -> tuple[_IntArray, _FloatArray]:
-    if not np.isfinite(width) or width <= 0:
-        raise ValueError("width must be positive")
+    if not np.isfinite(scale) or scale <= 0:
+        raise ValueError("scale must be positive")
     if not np.isfinite(spacing) or spacing <= 0:
         raise ValueError("spacing must be positive")
     if not np.isfinite(tail) or not (0.0 < tail < 1.0):
@@ -219,7 +215,7 @@ def _translation_invariant_kernel_weights(
             float(value) for value in dist.ppf([0.5 * tail, 1.0 - 0.5 * tail])
         )
 
-    bounds = width * np.array([z_left, z_right])
+    bounds = scale * np.array([z_left, z_right])
     if not np.all(np.isfinite(bounds)):
         raise ValueError("kernel bounds are not finite; increase tail")
 
@@ -227,8 +223,8 @@ def _translation_invariant_kernel_weights(
     m_max = int(np.clip(np.ceil(bounds[1] / spacing - 0.5), 0, max_offset))
     m = np.arange(m_min, m_max + 1, dtype=np.int64)
 
-    lower = (m - 0.5) * spacing / width
-    upper = (m + 0.5) * spacing / width
+    lower = (m - 0.5) * spacing / scale
+    upper = (m + 0.5) * spacing / scale
     weights = np.maximum(dist.cdf(upper) - dist.cdf(lower), 0.0)
     return _trim_kernel_weights(m, weights)
 
@@ -285,7 +281,7 @@ def _smooth_with_weights(
 def _smooth_relative_kernel_on_geomgrid(
     y: ArrayLike,
     log_spacing: float,
-    alpha: float,
+    scale: float,
     kernel: _Kernel,
     tail: float,
 ) -> _FloatArray:
@@ -293,18 +289,18 @@ def _smooth_relative_kernel_on_geomgrid(
 
     if y.ndim != 1:
         raise ValueError("y must be one-dimensional")
-    if not np.isfinite(alpha) or alpha < 0:
-        raise ValueError("alpha must be non-negative")
+    if not np.isfinite(scale) or scale < 0:
+        raise ValueError("scale must be non-negative")
     if not np.isfinite(log_spacing) or log_spacing <= 0:
         raise ValueError("log_spacing must be positive")
     if not np.isfinite(tail) or not (0.0 < tail < 1.0):
         raise ValueError("tail must be between 0 and 1")
-    if y.size == 0 or alpha == 0:
+    if y.size == 0 or scale == 0:
         return y.copy()
 
     offsets, weights = _relative_kernel_weights(
         log_spacing=log_spacing,
-        alpha=alpha,
+        scale=scale,
         kernel=kernel,
         tail=tail,
         max_offset=y.size - 1,
@@ -324,7 +320,7 @@ def _validate_max_grid_points(max_grid_points: int) -> None:
 def _smooth_relative_kernel(
     x: ArrayLike,
     y: ArrayLike,
-    alpha: float = 1.0,
+    scale: float,
     kernel: _Kernel = "gaussian",
     tail: float = 1e-12,
     max_grid_points: int = 1_000_000,
@@ -336,8 +332,8 @@ def _smooth_relative_kernel(
         raise ValueError("x and y must be one-dimensional")
     if x.size != y.size:
         raise ValueError("x and y must have the same length")
-    if not np.isfinite(alpha) or alpha < 0:
-        raise ValueError("alpha must be non-negative")
+    if not np.isfinite(scale) or scale < 0:
+        raise ValueError("scale must be non-negative")
     if not np.isfinite(tail) or not (0.0 < tail < 1.0):
         raise ValueError("tail must be between 0 and 1")
     _validate_max_grid_points(max_grid_points)
@@ -347,7 +343,7 @@ def _smooth_relative_kernel(
         raise ValueError("x must be positive")
     if np.any(np.diff(x) <= 0):
         raise ValueError("x must be strictly increasing")
-    if x.size == 0 or alpha == 0 or x.size == 1:
+    if x.size == 0 or scale == 0 or x.size == 1:
         return y.copy()
 
     logx = np.log(x)
@@ -372,7 +368,7 @@ def _smooth_relative_kernel(
     yg = np.interp(xp, x, y)
     zg = _smooth_relative_kernel_on_geomgrid(
         yg,
-        alpha=alpha,
+        scale=scale,
         log_spacing=log_range / (k - 1),
         kernel=kernel,
         tail=tail,
@@ -383,7 +379,7 @@ def _smooth_relative_kernel(
 def _smooth_kernel_values(
     x: ArrayLike,
     y: ArrayLike,
-    width: float,
+    scale: float,
     kernel: _Kernel = "gaussian",
     tail: float = 1e-12,
     max_grid_points: int = 1_000_000,
@@ -395,8 +391,8 @@ def _smooth_kernel_values(
         raise ValueError("x and y must be one-dimensional")
     if x.size != y.size:
         raise ValueError("x and y must have the same length")
-    if not np.isfinite(width) or width < 0:
-        raise ValueError("width must be non-negative")
+    if not np.isfinite(scale) or scale < 0:
+        raise ValueError("scale must be non-negative")
     if not np.isfinite(tail) or not (0.0 < tail < 1.0):
         raise ValueError("tail must be between 0 and 1")
     _validate_max_grid_points(max_grid_points)
@@ -404,7 +400,7 @@ def _smooth_kernel_values(
         raise ValueError("x must contain only finite values")
     if np.any(np.diff(x) <= 0):
         raise ValueError("x must be strictly increasing")
-    if x.size == 0 or width == 0 or x.size == 1:
+    if x.size == 0 or scale == 0 or x.size == 1:
         return y.copy()
 
     spacing = np.diff(x)
@@ -426,7 +422,7 @@ def _smooth_kernel_values(
 
     offsets, weights = _translation_invariant_kernel_weights(
         spacing=x_range / (k - 1),
-        width=width,
+        scale=scale,
         kernel=kernel,
         tail=tail,
         max_offset=k - 1,
@@ -486,21 +482,33 @@ def _scipp_output(
     return out
 
 
-def _width_in_coordinate_unit(width: sc.Variable, x: sc.Variable) -> float:
-    if not isinstance(width, sc.Variable):
-        raise TypeError("width must be a scipp.Variable")
-    if width.ndim != 0:
-        raise sc.DimensionError("width must be a scalar")
-    if width.variances is not None:
-        raise sc.VariancesError("kernel widths with variances are not supported")
-    return float(width.to(unit=x.unit).value)
+def _scale_in_coordinate_unit(scale: sc.Variable, x: sc.Variable) -> float:
+    if not isinstance(scale, sc.Variable):
+        raise TypeError("scale must be a scipp.Variable")
+    if scale.ndim != 0:
+        raise sc.DimensionError("scale must be a scalar")
+    if scale.variances is not None:
+        raise sc.VariancesError("kernel scales with variances are not supported")
+    return float(scale.to(unit=x.unit).value)
 
 
-def smooth_kernel(
+def _dimensionless_scale(scale: object) -> float:
+    if not isinstance(scale, sc.Variable):
+        if not isinstance(scale, Real):
+            raise TypeError("scale must be a real number or a scipp.Variable")
+        return float(scale)
+    if scale.ndim != 0:
+        raise sc.DimensionError("scale must be a scalar")
+    if scale.variances is not None:
+        raise sc.VariancesError("kernel scales with variances are not supported")
+    return float(scale.to(unit=sc.units.dimensionless).value)
+
+
+def smooth(
     x: _ScippArray,
     y: sc.Variable | None = None,
     *,
-    width: sc.Variable,
+    scale: sc.Variable,
     kernel: _Kernel = "gaussian",
     tail: float = 1e-12,
     max_grid_points: int = 1_000_000,
@@ -508,7 +516,7 @@ def smooth_kernel(
     """Smooth sampled data with a translation-invariant kernel.
 
     The kernel describes a distribution of displacements ``Z``, with displaced
-    coordinates given by ``x' = x + width * Z``. At the boundaries, the kernel
+    coordinates given by ``x' = x + scale * Z``. At the boundaries, the kernel
     is renormalized over the available finite input domain.
 
     Input that is not uniformly spaced is interpolated to a uniform grid,
@@ -523,15 +531,17 @@ def smooth_kernel(
         Values to smooth when ``x`` contains the sample coordinates. Must be a
         one-dimensional variable with the same dimension as ``x``. Must be
         omitted when ``x`` is a data array.
-    width:
+    scale:
         Scale factor for the displacement distribution. Must be a scalar with a
         unit compatible with the coordinate. Set to zero to return a copy of the
         input without smoothing.
     kernel:
-        Kernel distribution. Supported names are ``'gaussian'``, ``'rectangle'``,
-        and ``'triangle'``, including their aliases. Alternatively, provide a
-        fully specified distribution with ``cdf``, ``ppf``, and ``support``
-        methods.
+        Kernel distribution. The canonical names are ``'gaussian'``,
+        ``'boxcar'``, and ``'triangular'``. They represent a standard normal
+        distribution, a uniform distribution on [-1, 1], and a symmetric
+        triangular distribution on [-1, 1], respectively. Other aliases are
+        accepted. Alternatively, provide a fully specified distribution with
+        ``cdf``, ``ppf``, and ``support`` methods.
     tail:
         Total probability omitted when truncating a kernel with unbounded
         support.
@@ -553,23 +563,23 @@ def smooth_kernel(
         grid exceeds ``max_grid_points``.
     scipp.DimensionError
         If the inputs are not one-dimensional, a pair of variables does not
-        have matching dimensions, or ``width`` is not scalar.
+        have matching dimensions, or ``scale`` is not scalar.
     scipp.CoordError
         If a data array has no dimension coordinate or has a bin-edge
         coordinate.
     scipp.UnitError
-        If the unit of ``width`` is incompatible with the coordinate unit.
+        If the unit of ``scale`` is incompatible with the coordinate unit.
     scipp.VariancesError
-        If the signal or ``width`` has variances.
+        If the signal or ``scale`` has variances.
     TypeError
-        If ``width`` is not a variable, ``kernel`` is not a distribution-like
+        If ``scale`` is not a variable, ``kernel`` is not a distribution-like
         object, or ``max_grid_points`` is not an integer.
     """
     x, y, template = _scipp_input(x, y)
     values = _smooth_kernel_values(
         x.values,
         y.values,
-        width=_width_in_coordinate_unit(width, x),
+        scale=_scale_in_coordinate_unit(scale, x),
         kernel=kernel,
         tail=tail,
         max_grid_points=max_grid_points,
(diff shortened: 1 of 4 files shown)


jokasimr:
The idea is that the scaled kernel has constant width on a grid that is the logarithm of the original grid, and when the kernel is constant width it can be computed efficiently using `scipy.signal.convolve`.

Here's a document describing the implementation in more detail:

[relative_kernel_smoothing_math.pdf](https://github.com/user-attachments/files/30701336/relative_kernel_smoothing_math.pdf)

jokasimr:
@SimonHeybrock Do you think this functionality should go here in Scippneutron or in Scipp?

It's mainly useful for analysis work where you have an idealized model and want to apply some smoothing from resolution effects to obtain something that is comparable to the measurement data.

SimonHeybrock:
Before looking into details, can you compare this to what we already have in https://scipp.github.io/generated/modules/scipy/scipp.scipy.ndimage.gaussian_filter.html (and https://scipp.github.io/generated/modules/scipy/scipp.scipy.ndimage.generic_filter.html)? What is added on top of that? Would it make sense to have a common API?

jokasimr:
The difference is that this is a more convenient interface for typical smoothing needs, more of a interface for users than for developers.

1. (Grid awareness) The functions implemented in this module are grid aware while `gaussian_filter` is not grid aware. That is, for `gaussian_filter`, the effect on the signal will depend on the grid it was defined on. If the intention of the user is to smooth the input with a fixed-width gaussian kernel then they have to make sure their signal is defined on a uniform grid, etc.

2. (Flexible kernel support) This module exposes a number of common smoothing kernels, and any distribution in `scipy.stats` can be used as a kernel.

3. (Relative width kernels) `smooth_relative` is a convenience function for smoothing with a kernel having a width that grows proportional to the grid coordinate. That is a common case, think about instruments where uncertainty in a coordinate is proportional to itself (`q` or `wavelength` uncertainty grows proportional to `q` or `wavelength`). But such operations can be quite hard to implement correctly and efficiently using a tool like `generic_filter`.

Since the functions implemented in this module use `scipy.signal.convolve` under the hood they are probably significantly faster than `generic_filter` in most cases. But I have not made any runtime comparisons.

The functions in this module are explicitly 1D, and not ND like the functions you mentioned. That is a limitation, but it allows a simpler interface and it covers common use cases.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+    logx = np.log(x)
+    dlog = np.diff(logx)
+    log_range = float(logx[-1] - logx[0])
+
+    # Preserve an existing geometric grid. Otherwise choose a geometric grid
+    # at least as dense as the smallest input spacing in log-space.
+    if np.allclose(dlog, dlog[0], rtol=1e-7, atol=0.0):
+        k = x.size
+    else:
+        k = int(np.ceil(log_range / np.min(dlog))) + 1
```
SimonHeybrock: Crash on plausible input: this raises `OverflowError: cannot convert float infinity to integer` rather than the intended `max_grid_points` `ValueError`.

```python
x = np.array([3.0, np.nextafter(3.0, 6.0), 6.0])
smooth_relative(sc.array(dims=['x'], values=x, unit='m'), y, scale=0.1)
# RuntimeWarning: divide by zero encountered in scalar divide
# OverflowError: cannot convert float infinity to integer
```

Also reproduces for `10.0` and `0.1`.

Root cause is the mismatch between what is validated and what is divided by: line 344 checks `np.diff(x) > 0`, but here we divide by `np.min(np.diff(np.log(x)))`. For `|log x| >= 1` one ulp of `log x` exceeds the true log-difference of adjacent doubles, so `dlog` rounds to exactly zero while `diff(x)` is still positive. `_smooth_kernel_values` is immune precisely because it divides by the same array it validated.

Suggest computing `k` in floating point and letting the `max_grid_points` check reject non-finite values, which handles the overflow and the too-many-points case in one branch:

```python
k = np.ceil(log_range / np.min(dlog)) + 1.0  # may be inf
...
if not (k <= max_grid_points):
    raise ValueError(...)
k = int(k)
```

tests/smoothing_test.py
```diff
@@ -0,0 +1,790 @@
...
+        x,
+        y,
+        scale=0.1,
+        max_grid_points=size,
+    )
+
+    assert actual.shape == y.shape
+
+
+def test_pathologically_close_coordinates_fail_before_allocation():
```
SimonHeybrock: This passes by accident. `x[0] == 1.0` and `log(1.0) == 0`, so the ulp of the log is tiny and the log-difference survives. With any other base the function raises `OverflowError` instead of the `ValueError` this test is asserting (see the comment on `smoothing.py:358`).

Worth parametrizing over a few bases, e.g. `1.0, 3.0, 10.0, 0.1`.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+    x_range = float(x[-1] - x[0])
+
+    # Preserve an existing uniform grid. Otherwise choose a uniform grid at
+    # least as dense as the smallest input spacing.
+    if np.allclose(spacing, spacing[0], rtol=1e-7, atol=0.0):
+        k = x.size
+    else:
+        k = int(np.ceil(x_range / np.min(spacing))) + 1
+
+    if k > max_grid_points:
```
SimonHeybrock: `max_grid_points` also fires on the branch where no resampling happens. When the input is already uniform `k = x.size`, so a plain uniform input larger than the limit is rejected with a message that says resampling would be required:

```python
x = sc.linspace('x', 0.0, 1.0, 11, unit='m')
smooth(x, y, scale=sc.scalar(0.1, unit='m'), max_grid_points=10)
# ValueError: uniform resampling would require too many points, exceeding max_grid_points=10
```

With the default of 1e6 this rejects a 2M-point uniform spectrum that needs no intermediate grid at all. The parameter is documented as bounding the *intermediate* grid, so I think the check belongs on the resampling branch only. Same at line 360.

SimonHeybrock (commented):
Nice work — the numerical core is careful and the test suite is the strongest part of this PR. Analytic reference solutions for the smoothed quadratic, second-order convergence checks against grid spacing, and tail-truncation scaling with a grid-limited plateau is real verification rather than smoke tests. The convolution alignment, the O(n) boundary normalization in `_valid_weight_sums`, and the direct-vs-FFT switch for non-finite input all check out.

Findings below, roughly in order of importance. One is a genuine crash (`OverflowError` on the relative path for coordinates that collide in log space), two are smaller error-handling issues, and the rest is structure and documentation.

The largest item is that the relative and translation-invariant paths are one algorithm under a coordinate map — I prototyped the unification to confirm that, details on `_translation_invariant_kernel_weights`.

Separately, and not a code review point so not raised inline: I still want to settle whether this belongs in scippneutron at all, given it is 672 lines with no neutron-specific content. Let's discuss that on the thread.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+            raise sc.CoordError("data must have a dimension coordinate")
+        if x.coords.is_edges(x.dim):
+            raise sc.CoordError("the dimension coordinate must not contain bin edges")
+        if x.masks:
+            raise ValueError("smoothing data with masks is not supported")
+        template = x
+        y = x.data
+        x = x.coords[x.dim]
+    elif not isinstance(y, sc.Variable):
+        raise TypeError("expected a DataArray or a pair of Variables")
```
SimonHeybrock: This guard only validates `y`, so a numpy `x` with a scipp `y` falls through to line 458 and raises `AttributeError: 'numpy.ndarray' object has no attribute 'dims'` instead of the intended `TypeError`.

`test_rejects_numpy_arrays` misses this because it passes numpy for both arguments.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+    zU = np.expm1(U) / scale
+
+    # Exact cell-integrated weights in log-space.
+    w = (dist.cdf(zU) - dist.cdf(zL)) / norm_mass
+    w = np.maximum(w, 0.0)
+
+    return _trim_kernel_weights(m, w)
+
+
+def _translation_invariant_kernel_weights(
```
SimonHeybrock: This and `_relative_kernel_weights` are the same function under a coordinate map, and the reduction is exact rather than approximate: substitute `u(z) = scale*z`, `z(u) = u/scale`, `z_domain_min = -inf` into the relative version and `norm_mass` collapses to 1, `has_finite_log_support` collapses to the plain `isfinite` check, and `log1p`/`expm1` collapse to linear. One level up, `_smooth_relative_kernel` and `_smooth_kernel_values` differ only by log-vs-identity plus the `x > 0` check.

I prototyped the unified version to check this isn't hand-waving: the four functions become two, parameterized on a small geometry object supplying `coordinate`, `points`, `displacement_min`, `offset` and `displacement`. The core drops from 436 to ~290 lines, all 73 tests in this PR pass unchanged, and across 180 combinations of grid (uniform, geometric, jittered, 4-point, 1e-6..1e6), kernel (gaussian, boxcar, triangular, asymmetric uniform, asymmetric triangular, expon) and scale, the outputs agree with this branch to 1e-10.

Two things fall out for free. The truncation condition becomes `isfinite(u_left) and isfinite(u_right)`, which is provably equivalent to `has_finite_log_support` but easier to check by eye. And computing `k` in float for the shared `max_grid_points` guard fixes the overflow reported on line 358 by construction.

Happy to push the prototype if useful.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+    start = int(offsets[-1])
+    numerator = full[start : start + y.size]
+    denominator = _valid_weight_sums(y.size, offsets, weights)
+
+    out = np.full_like(numerator, np.nan)
+    np.divide(numerator, denominator, out=out, where=denominator > 0.0)
+    return out
+
+
+def _smooth_relative_kernel_on_geomgrid(
```
SimonHeybrock: `scale`, `tail` and `log_spacing` are validated four times on the relative path: in `smooth_relative`, here, and again in `_relative_kernel_weights` — all private, one call path. The fixed-width path has no `_on_uniformgrid` counterpart, which suggests this layer is an artifact rather than a design choice. Validating once at the public boundary would be enough.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+def smooth(
+    x: _ScippArray,
+    y: sc.Variable | None = None,
+    *,
+    scale: sc.Variable,
+    kernel: _Kernel = "gaussian",
+    tail: float = 1e-12,
+    max_grid_points: int = 1_000_000,
+) -> _ScippArray:
+    """Smooth sampled data with a translation-invariant kernel.
```
SimonHeybrock, line 463: Two properties are worth stating explicitly here, because both public functions are normalized weighted averages rather than convolutions:

- A constant is preserved exactly, but a sum is not. Boundary renormalization pulls mass back into the domain, so a Gaussian peak's total changes by ~2% at `scale=1.0` on a 200-point grid. That is the right behavior for the use case in the PR description (smearing an idealized model to compare against data), but it makes the function unsuitable for smearing raw counts — and the tests use `unit='counts'` throughout, which invites exactly that reading.
- The result may contain NaN where no kernel mass is reachable: at the boundary for one-sided kernels, and for the whole array when the kernel has no reachable mass at all (the `nonzero.size == 0` path at line 88). This is deliberate and tested, but neither `Returns` section mentions it.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
...
+            "Smoothing signals with variances is not supported because it would "
+            "introduce correlations between data points."
+        )
+    return x, y, template
+
+
+def _scipp_output(
+    template: sc.DataArray | None, y: sc.Variable, values: _FloatArray
+) -> _ScippArray:
+    data = sc.array(dims=y.dims, values=values, unit=y.unit)
```
SimonHeybrock, line 422: Silent dtype widening: float32 input produces float64 output, and integer data also produces float64. Doubling memory on a large spectrum is a surprise worth either avoiding or documenting.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,672 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2026 Scipp contributors (https://github.com/scipp)
+
+from __future__ import annotations
```
SimonHeybrock, line 15: No module docstring. The geometric-grid trick — that a relative-width kernel is translation invariant in `log(x)`, which is what makes `scipy.signal.convolve` applicable — is the central idea here, and it currently lives only in a PDF attached to a PR comment. That should be in the module docstring so it survives.

tests/smoothing_test.py
```diff
@@ -0,0 +1,790 @@
...
+# Copyright (c) 2026 Scipp contributors (https://github.com/scipp)
+
+import numpy as np
+import pytest
+import scipp as sc
+from scipy.special import ndtr
+from scipy.stats import uniform
+
+import scippneutron as scn
+from scippneutron.smoothing import (
```
SimonHeybrock: Three private functions imported. `test_kernel_stencil_is_bounded_before_allocation`, `test_asymmetric_kernel_convolution_matches_direct_weighted_sum` and `test_kernel_with_no_reachable_mass_returns_nan` can all go through `smooth_relative` on an already-geometric grid, which would also keep them working through the refactor suggested on `smoothing.py:195`.

src/scippneutron/smoothing.py
jokasimr: This is not important. The spectrums are 1D, it's unlikely that doubling their memory usage will cause any issues, and that saves us from the complexity of keeping track of the dtype of the input, returning truncated results etc.

src/scippneutron/smoothing.py
jokasimr: Makes sense.

src/scippneutron/smoothing.py
jokasimr: I don't think this is worth DRYing up by unifying the two cases. It's better to keep them separate because I think that makes it easier to understand, it's already complex enough.

src/scippneutron/smoothing.py
jokasimr: Fixed by allowing intermediate grids to expand 2x compared to the input even if they exceed `max_grid_points`. Also updated the documentation of `max_grid_points` to say that this is the upper bound of intermediate grid sizes that are allowed unconditionally.

tests/smoothing_test.py
jokasimr: I think testing internal functions is fine here. It's simpler and more specific than testing the behavior through the public interface.

src/scippneutron/smoothing.py
jokasimr: I removed the unnecessary `_on_geomgrid` layer.

src/scippneutron/smoothing.py
SimonHeybrock: But less duplication is easier to review.

jokasimr: I don't agree that's obviously the case.
I've already said what I think about it. Do you think I should do the refactor suggested by the model?

SimonHeybrock: yes

jokasimr: I tried a refactor to unify `_relative_kernel_weights` and `_translation_invariant_kernel_weights`, but it did not result in any real code reduction, while being more complex.

It's possible there's a better shape that actually reduces code size and complexity. You are welcome to push such a version to this branch if you prefer it.

src/scippneutron/smoothing.py
jokasimr: I think the first point has to be clarified a bit. Exactly what was the test case and what was the result of that test?

It's not correct to say that the normalization behavior "makes the function unsuitable for smearing raw counts". That is a far to general statement. It is just context dependent.

The smoothing operation needs to have a strategy for handling boundaries. Here the strategy used is to re-normalize by the mass of the kernel that falls inside the boundary. This can be understood as an assumption that the (kernel-)weighted mean of the signal outside the boundary is the same as the weighted mean of the signal inside the boundary.
In practice that is a conservative assumption, we don't expect anything drastic happening to the signal exactly at the boundary.
But of course that assumption will be wrong sometimes, any assumption is.

### How to deal with that as a user

As a user we might know something about the behavior of the signal outside the domain, for example, we might know it is zero, or we know it decays following a certain patterns, or something else.
In almost all such cases the smoothed signal near the boundary will not be what it would be if we had taken our extra knowledge into account properly.

If the user wants perfect boundary behavior the best option for them is to extend the signal that they pass to the smoother with the "tails" that they assume it has outside of the bounds of the signal. Then they can crop out the portion of the smoothed signal that overlaps with the real signal, and that entire portion will have been unaffected by "boundary effects".

Another (simpler) option is to smooth the signal and then cut out a center section of the smoothed signal that is unaffected by boundary effects.

SimonHeybrock:
Pushed `d9110a70`. Fair warning that the line count barely moves (643 → 619, 190 → 177 statements) — you were right that unification is not primarily a size win. The case for it turned out to be elsewhere.

The reduction is exact rather than approximate, which is what makes the shape work: substituting `u(z) = scale*z`, `z(u) = u/scale` and `z_min = -inf` into `_relative_kernel_weights` gives `_translation_invariant_kernel_weights` identically. `norm_mass` collapses to 1, `log1p`/`expm1` collapse to linear, and `has_finite_log_support` collapses to a plain `isfinite` check on the bounds in `u`. So the geometry only has to supply the working coordinate and the map between a displacement and its offset in that coordinate — five small methods, no configuration and no branching inside the shared code.

What convinced me it was worth pushing rather than dropping: the paths had already diverged again. `7a820602` fixed the infinite grid size on the relative path, but the uniform path still raised `OverflowError` for inputs whose spacing ratio overflows rather than merely being large:

```python
smooth(sc.array(dims=['x'], values=np.array([0.0, 5e-324, 1.0]), unit='m'), y, scale=sc.scalar(0.1, unit='m'))
# OverflowError: cannot convert float infinity to integer
```

Same for `[1.0, nextafter(1.0), 1e300]`. Both now raise the intended `ValueError`, and with one path they cannot drift apart again. Added those as cases on the existing test.

Verification, since this is a numerics change and the diff is not small:

- All existing tests pass unchanged apart from the two noted below.
- 896 combinations of grid (uniform, geometric, jittered, 4-point, 2-point, 1e-6..1e6, 1..1e12, clustered) × kernel (gaussian, boxcar, triangular, asymmetric uniform, asymmetric triangular, expon, shifted normal) × scale × tail × both functions, compared against `7a820602`: identical values to 1e-11 and identical exception types, with the `OverflowError` → `ValueError` cases above the only intended differences.
- `mypy` on this file goes from three errors to none. Two were the `spacing` name being reused for both an array and a float; the third was the `isinstance(x, sc.Variable)` guard being statically unreachable, which I fixed by typing `_scipp_input` as taking `object`, since it is the boundary where untyped user input is validated.

Two test changes, both flagged in the review as testing implementation rather than behavior. `test_kernel_stencil_is_bounded_before_allocation` now goes through `smooth_relative` on a grid where the gaussian is flat across the whole input, so it asserts every point averages the entire array — that fails just as loudly if the stencil clamp is removed, without naming a private helper. That removes the last private import from the test module.

Untouched, since they are yours to decide rather than mechanical: the float32 → float64 widening, and documenting that these are normalized weighted averages, so a constant is preserved exactly but a sum is not (~2% on a peak at `scale=1.0`), and that the result can legitimately contain NaN.

SimonHeybrock:
@jokasimr What do you think about the pushed change? I like the `_Geometry` protocol, since it aids understanding.

tests/smoothing_test.py
```diff
@@ -709,20 +717,18 @@ def test_wide_coordinate_range_does_not_overflow_grid_construction():
...
+    # reach the input rather than allocated at its nominal width.
+    size = 21
+    x = np.geomspace(1.0, np.exp((size - 1) * 1e-6), size)
+    y = np.arange(size, dtype=float)
+
+    actual = _smooth_values(smooth_relative, x, y, scale=1.0)
+
+    # The kernel is flat to within 1e-5 over the whole input, so every point
+    # averages the entire array.
+    np.testing.assert_allclose(actual, np.full(size, y.mean()), rtol=1e-4)
```
jokasimr, line 731: I think testing the internal helper was simpler and clearer here. It needed less explanatory comments. I'm also not sure if this tests the same things. 

Doesn't seem worth it to me just to get the private helper out of the tests.

jokasimr:
> @jokasimr What do you think about the pushed change? I like the _Geometry protocol, since it aids understanding.

I think it looks good :+1: 

I'm just a little bit worried about the changed test, it would be good to either re-add that, or test it locally manually once to make sure there's no behavioral change.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+    xp = geometry.points(float(x[0]), float(x[-1]), k)
+    zg = _smooth_with_weights(np.interp(xp, x, y), offsets, weights)
+    return cast(_FloatArray, np.interp(x, xp, zg))
+
+
+def _scipp_input(
+    x: object, y: object
+) -> tuple[sc.Variable, sc.Variable, sc.DataArray | None]:
+    """Validate untyped user input and reduce it to a coordinate/data pair."""
```
SimonHeybrock: Both entry points accept either a `DataArray` or an `(x, y)` pair of Variables. I would suggest dropping the pair form and taking a `DataArray` only, matching `scipp.scipy`: `ndimage.gaussian_filter(x, /, *, sigma=...)` and `signal.sosfiltfilt(obj, dim, *, sos=...)` both take a single data argument rather than a coord/data pair.

The pair form saves the caller a `sc.DataArray(y, coords={'x': x})`. In exchange, this function re-validates dimensionality, matching dims and binned-ness that a `DataArray` enforces structurally; `y` becomes a positional-optional with a "must be omitted" `TypeError`; and the first parameter is named `x` even when it holds the data.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    """
+    dist = _as_kernel_distribution(kernel)
+
+    z_min = geometry.displacement_min(scale)
+    p_min = float(dist.cdf(z_min))
+    reachable_mass = 1.0 - p_min
+    if not np.isfinite(reachable_mass) or reachable_mass <= 0.0:
+        raise ValueError("kernel has no mass in the valid domain for this scale")
+
+    support_min, support_max = _kernel_support(dist)
```
SimonHeybrock: `_kernel_support` has a helpful message for unfrozen distributions:

> kernel must be a fully specified distribution. For distributions with shape parameters, pass a frozen distribution such as triang(c=0.5), not triang.

It is unreachable. `dist.cdf(z_min)` above runs first, so `kernel=triang` gives the user `TypeError: _parse_args() missing 1 required positional argument: 'c'` instead. Distributions without shape parameters (`norm`, `uniform`) pass the CDF call and never reach the branch either, so it never fires at all.

Taking the support first fixes it:

```suggestion
    z_min = geometry.displacement_min(scale)
    support_min, support_max = _kernel_support(dist)

    p_min = float(dist.cdf(z_min))
    reachable_mass = 1.0 - p_min
    if not np.isfinite(reachable_mass) or reachable_mass <= 0.0:
        raise ValueError("kernel has no mass in the valid domain for this scale")
```

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+        y = x.data
+        x = x.coords[x.dim]
+    elif not isinstance(x, sc.Variable) or not isinstance(y, sc.Variable):
+        raise TypeError("expected a DataArray or a pair of Variables")
+
+    if x.ndim != 1 or y.ndim != 1:
+        raise sc.DimensionError("x and y must be one-dimensional")
+    if x.dims != y.dims:
+        raise sc.DimensionError("x and y must have the same dimension")
+    if x.is_binned or y.is_binned:
```
SimonHeybrock: A binned `DataArray` hits `x.coords.is_edges(x.dim)` first and gets `CoordError: the dimension coordinate must not contain bin edges`, which points at the wrong thing.

This check is only reachable through the Variable-pair path, so if that path goes away it becomes dead code unless it moves ahead of the `is_edges` check.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+
+    # Trim zero-only ends, but preserve offset zero for convolution alignment.
+    zero = -offsets[0]
+    first = min(nonzero[0], zero)
+    last = max(nonzero[-1], zero) + 1
+    weights = weights[first:last]
+    return offsets[first:last], weights / weights.sum()
+
+
+class _Geometry(Protocol):
```
SimonHeybrock: This part was implemented by myself, looking at it again now I see that `scale` is threaded through `displacement_min`, `offset` and `displacement`, and is the same value for the whole call. Binding it at construction (`_Geometric(scale)`) would drop the parameter from three signatures, take `_kernel_weights` from six parameters to five, and make a mismatched scale between the `offset`/`displacement` inverse pair impossible to express.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    # gives an infinite bound in u. Truncate at the requested tail instead.
+    if not (np.isfinite(u_left) and np.isfinite(u_right)):
+        probabilities = (
+            p_min + np.array([0.5 * tail, 1.0 - 0.5 * tail]) * reachable_mass
+        )
+        u_left, u_right = (
+            float(value) for value in geometry.offset(scale, dist.ppf(probabilities))
+        )
+        if not np.isfinite(u_right):
+            raise ValueError("right kernel bound is not finite; increase tail")
```
SimonHeybrock: Worth a comment here: there is no matching finiteness check on `u_left`, which instead clamps to `-max_offset` via `np.clip` below. That is right on both sides -- the left bound is bounded by the valid domain, the right one is not, and the geometric case with `scale=1` relies on the clamp -- but the asymmetry reads as an oversight without a line saying why.

Minor, same block: when only one side of the support is infinite, both bounds get recomputed from `ppf`, so `tail/2` is discarded off a side that was already finite. Within budget, just slightly wasteful.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    nonzero = np.flatnonzero(weights > 0.0)
+    if nonzero.size == 0:
+        return offsets, weights
+
+    # Trim zero-only ends, but preserve offset zero for convolution alignment.
+    zero = -offsets[0]
+    first = min(nonzero[0], zero)
+    last = max(nonzero[-1], zero) + 1
+    weights = weights[first:last]
+    return offsets[first:last], weights / weights.sum()
```
SimonHeybrock: `weights.sum()` cancels: `_smooth_with_weights` divides the convolution by `_valid_weight_sums` over the same weights, so any constant factor drops out. Not wrong, but it reads as load-bearing. I would drop the division rather than add a comment saying it does nothing.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+        np.convolve(np.ones(n), w[::-1], mode="full")[start:start+n]
+
+    but O(n), not another full convolution.
+    """
+    i = np.arange(n, dtype=np.int64)
+
+    m_min = int(m[0])
+    m_max = int(m[-1])
+
+    lower = np.maximum(m_min, -i)
```
SimonHeybrock: `lower <= 0 <= upper` always holds, because `_kernel_weights` guarantees `m_min <= 0 <= m_max` -- that is what the offset-zero preservation in `_trim_kernel_weights` is for. It is crucial (an empty range would silently give a zero denominator) and worth a line, since the invariant is established two functions away.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    y: _FloatArray, offsets: _IntArray, weights: _FloatArray
+) -> _FloatArray:
+    # Desired operation:
+    #
+    #   out[i] = sum_m weights[m] * y[i + m]
+    #
+    # scipy convolution reverses the second argument, hence weights[::-1].
+    # FFT convolution would spread a single NaN or infinity over the entire
+    # output. SciPy recommends the direct method for non-finite inputs.
+    method = "auto" if np.all(np.isfinite(y)) else "direct"
```
SimonHeybrock: The reasoning is right, and I confirmed `direct` does localise the contamination -- a single `inf` affects exactly its own window, not the whole array.

The cost is not stated anywhere, though. Same data, one NaN vs none:

| n | scale | clean | one NaN |
|---|---|---|---|
| 10 000 | 500 | 0.0018 s | 0.0134 s |
| 50 000 | 5 000 | 0.014 s | 0.43 s |
| 200 000 | 3 000 | 0.036 s | 1.0 s |

The ratio grows with the stencil width, since `direct` is O(n*m) against FFT's O(n log n) and `m` is clamped only by `k-1`. NaN is fairly ordinary in reduced data (zero monitor bins, masked pixels), so a user can hit a ~30x cliff between two visually identical datasets with nothing indicating why.

You have a far better idea than I do of the typical `n` and kernel widths here -- if this is a regime people actually reach, a sentence in the docstrings would be worth it. Purely as a note for later: a mask-and-renormalise formulation would restore FFT speed and would incidentally subsume `_valid_weight_sums`, since boundary renormalisation is the same operation with an all-ones mask. But it turns NaN from "poisons its window" into "ignored", so that is a design decision rather than a refactor.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,621 @@
...
+    if np.allclose(du, du[0], rtol=1e-7, atol=0.0):
+        k = float(x.size)
+    else:
+        # Coordinates that are distinct but collide, or nearly so, in the
+        # working coordinate give an infinite point count. Let it propagate to
+        # the guard below rather than raising here.
+        with np.errstate(divide="ignore", over="ignore"):
+            k = np.ceil(u_range / np.min(du)) + 1.0
+
+    if not (k <= max_grid_points) and k > 2 * x.size:
```
SimonHeybrock: A grid is rejected only if it exceeds both `max_grid_points` and `2 * x.size`, so the effective cap is `max(max_grid_points, 2 * x.size)` -- the parameter is a floor on what is permitted, not a maximum. A 600k-point input needing a 1.1M-point grid is accepted at the default `max_grid_points=1_000_000`.

Both rules are sensible and I would keep both: the absolute one protects small inputs from pathological spacing, the relative one keeps a merely irregular grid from being punished for being long. The problem is that two policies are exposed through one knob named after one of them, and the error message names only `max_grid_points` when the 2n rule is equally responsible. Either name them separately, or say so in the docstring and in the message so the user can tell which limit they hit.

Aside: `not (k <= max_grid_points)` reads as NaN defence, but `k` is NaN only when `u_range == 0` and `min(du) == 0`, which the `allclose` branch above catches and routes to "grid spacing must be positive" (checked with `x = [1e300, nextafter(1e300), ...]`). `k > max_grid_points` says the same thing.

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+            dims=['x'],
+            values=np.asarray(y),
+            unit='counts',
+        ),
+    )
+
+
+def _smooth_values(smooth, x, y, **kwargs):
+    x, y = _variables(x, y)
+    return smooth(x, y, **kwargs).values
```
SimonHeybrock: `smooth` shadows the imported `smooth`, and the helper name shadows `smoothing._smooth_values`. The parametrize ids already call this one `function`:

```suggestion
def _smooth_values(function, x, y, **kwargs):
    x, y = _variables(x, y)
    return function(x, y, **kwargs).values
```

Renaming the helper itself (`_smoothed_values`?) would clear the second collision.

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+    sizes = np.array([500, 1000, 2000, 4000])
+    errors = np.array(
+        [
+            np.max(np.abs(_fixed_width_quadratic_smoothing_error(size=size, width=0.1)))
+            for size in sizes
+        ]
+    )
+
+    observed_orders = np.log2(errors[:-1] / errors[1:])
+    np.testing.assert_allclose(observed_orders, 2.0, atol=0.06)
```
SimonHeybrock: These are two-sided: +/-3% on a measured convergence order. Only the lower bound carries meaning -- converging faster than second order is not a defect -- and 09e8848b already had to loosen a tolerance once. `assert np.all(observed_orders > 1.9)` keeps the content and drops the flake surface.

Same thought for `reduction_per_decade` on lines 140 and 474, where the upper bound is doing little, and for line 437.

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+        scale=0.1,
+        kernel=uniform(loc=100.0, scale=1.0),
+        tail=1e-12,
+    )
+
+    assert np.all(np.isnan(actual))
+
+
+@pytest.mark.parametrize("scale", [np.nan, np.inf, -np.inf, -1.0])
+def test_rejects_invalid_relative_scale_before_noop_return(scale):
```
SimonHeybrock: `scale=0` is documented in both public docstrings ("Set to zero to return a copy of the input without smoothing") and has no test. It is the one uncovered line in the module that is not an error path (`smoothing.py:341`; 92% statement coverage overall). I checked it does work and returns an unaliased copy.

Also uncovered, all cheap and all verified correct by hand: unknown kernel string, non-monotonic `x`, 2-D input, and non-real `scale` for `smooth_relative`.

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+
+    with pytest.raises(sc.VariancesError, match="scales with variances"):
+        smooth(
+            x,
+            y,
+            scale=sc.scalar(0.1, variance=0.01, unit='m'),
+        )
+
+
+def test_gaussian_smoothing_matches_exact_quadratic_on_geometric_grid():
```
SimonHeybrock:
```suggestion
def test_relative_width_gaussian_matches_exact_quadratic_on_geometric_grid():
```

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+    # The outer cell edges of this grid coincide with the finite integration
+    # domain used by _quadratic_smoothing_error.
+    error = _quadratic_smoothing_error(size=4000, alpha=0.1)
+
+    assert np.max(np.abs(error)) < 5e-7
+    assert np.sqrt(np.mean(error**2)) < 1e-7
+
+
+def test_gaussian_smoothing_error_is_second_order_in_log_grid_spacing():
```
SimonHeybrock:
```suggestion
def test_relative_width_gaussian_error_is_second_order_in_log_grid_spacing():
```

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+        ]
+    )
+
+    # Halving the log-grid spacing should reduce midpoint quadrature error by
+    # four, corresponding to second-order convergence.
+    observed_orders = np.log2(errors[:-1] / errors[1:])
+    np.testing.assert_allclose(observed_orders, 2.0, atol=0.06)
+
+
+def test_gaussian_smoothing_error_scales_with_inverse_kernel_width():
```
SimonHeybrock:
```suggestion
def test_relative_width_gaussian_error_scales_with_inverse_kernel_width():
```

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+    scaled_errors = errors * alphas / log_step**2
+    assert np.all(np.diff(errors) < 0.0)
+    np.testing.assert_allclose(
+        scaled_errors,
+        np.mean(scaled_errors),
+        rtol=0.08,
+    )
+
+
+def test_gaussian_smoothing_error_scales_with_tail_until_grid_error_dominates():
```
SimonHeybrock:
```suggestion
def test_relative_width_gaussian_error_scales_with_tail_until_grid_error_dominates():
```

tests/smoothing_test.py
```diff
@@ -0,0 +1,836 @@
...
+@pytest.mark.parametrize(
+    ("kernel", "relative_variance", "max_error"),
+    [
+        ("boxcar", 1.0 / 3.0, 3e-7),
+        ("triangular", 1.0 / 6.0, 5e-8),
+    ],
+)
+def test_compact_symmetric_kernel_matches_exact_interior_quadratic_moments(
+    kernel, relative_variance, max_error
+):
```
SimonHeybrock:
```suggestion
def test_relative_width_compact_kernel_matches_interior_quadratic_moments(
    kernel, relative_variance, max_error
):
```

SimonHeybrock (commented):
This review was created with the help of AI. If anything below strikes you as unpleasantly formulated (tone, verbosity, ...), as too nit-picky, or as otherwise improvable, please tell me -- I am actively trying to improve this.

The algorithm looks correct to me. I checked the two convolution helpers independently against a brute-force double loop over 300 random asymmetric stencils, and both the alignment (`start = offsets[-1]`) and the O(n) boundary normalisation come out exact.

Nice work on the tests! Comparing against an analytically smoothed quadratic via truncated-normal moments, rather than golden values, means they would catch a wrong answer and not merely a changed one; `test_asymmetric_kernel_convolution_matches_direct_weighted_sum` brute-forces the convolution, which is exactly the test I would have asked for; and the `nextafter` / `5e-324` / `1e300` cases are more thorough than most numerical code gets.

Comments below are mostly small. The one I would like settled before merge is the input API (see the comment on `_scipp_input`), since it is public and awkward to walk back later.

Aside, not for this PR: `docs/user-guide/algorithms-background/` would be a natural home for the figure in the PR description at some point -- it currently only lives here, and it illustrates the behaviour well.

src/scippneutron/smoothing.py
jokasimr: The scale value in the table doesn't say how wide the kernel was, that is, how many neighboring points it takes into account. So it's hard to say if it's realistic or not. The `n` values are in the large end of the spectrum, but they are realistic. 

I think the suggestion to ignore nan values is reasonable.

src/scippneutron/smoothing.py
jokasimr: I think it's better to expose as little as possible of this logic to the user because it just makes the interface more complicated. 

The contract is: if the intermediate grid is smaller than `max_grid_points` then it will not be rejected.

Adding more "knobs" is unnecessary.
If the user hits the grid size limit they increase the `max_grid_points` and then they will not hit the limit any more.

tests/smoothing_test.py
jokasimr: This is a nit

src/scippneutron/smoothing.py
jokasimr: I'll remove the pair-path, so this will also go away

tests/smoothing_test.py
jokasimr: If you think there are missing tests, feel free to add those.

src/scippneutron/smoothing.py
jokasimr: I changed my mind about this. Ignoring nans introduces too many edge cases. For now the user will have to do that manually before smoothing the function. If we figure out later what the perfect behavior is here we can always add it later. For now I'll raise a warning if there are nans in the data, to let the user know that this might impact performance.

src/scippneutron/smoothing.py
jokasimr: I think it's fine to pass scale as an argument only to the methods that needs it.

src/scippneutron/smoothing.py
jokasimr: Added a comment to explain the asymmetry :+1: 

I think the minor point is not important enough to do anything about.

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,616 @@
...
+        Total probability omitted when truncating a kernel with unbounded
+        support. Must be strictly between zero and one.
+    max_grid_points:
+        Intermediate uniform grids no larger than this are always allowed. Larger
+        grids may be rejected to guard against excessive resampling.
+
+    Returns
+    -------
+    :
+        Smoothed data. Coordinates and units are preserved.
```
SimonHeybrock: The `Returns` section does not mention that the result can legitimately contain NaN. It happens in two places a user can reach without doing anything wrong: at the boundary for one-sided kernels, where no kernel mass falls inside the domain, and across the whole array when the kernel has no reachable mass at all. I confirmed both on this branch with `expon()` and `uniform(loc=0.5, scale=1.0)`.

This is deliberate and tested, so it is only the documentation that is missing -- but a silent NaN is the kind of thing worth finding in the docstring rather than in a plot.

```suggestion
    Returns
    -------
    :
        Smoothed data. Coordinates and units are preserved. Points where no
        kernel mass falls inside the input domain are NaN; this can occur at
        the boundaries for one-sided kernels.
```

src/scippneutron/smoothing.py
```diff
@@ -0,0 +1,616 @@
...
+        or support reaching the nonpositive coordinate domain. Must be strictly
+        between zero and one.
+    max_grid_points:
+        Intermediate geometric grids no larger than this are always allowed. Larger
+        grids may be rejected to guard against excessive resampling.
+
+    Returns
+    -------
+    :
+        Smoothed data. Coordinates and units are preserved.
```
SimonHeybrock: Same here, for the same reason as on `smooth`.

```suggestion
    Returns
    -------
    :
        Smoothed data. Coordinates and units are preserved. Points where no
        kernel mass falls inside the input domain are NaN; this can occur at
        the boundaries for one-sided kernels.
```

SimonHeybrock (approved):
This review was created with the help of AI. If anything below strikes you as unpleasantly formulated (tone, verbosity, ...), as too nit-picky, or as otherwise improvable, please tell me -- I am actively trying to improve this.

Approving. Everything substantive from both rounds is addressed, and the pushbacks hold up -- I agree on the dtype widening, on keeping `max_grid_points` a single knob, and that the convergence-order tolerances were a nit.

You were right to object to "unsuitable for smearing raw counts": that was too general a claim from a single peak-integral measurement. Your framing -- renormalisation assumes the kernel-weighted mean outside the domain equals the one inside, and a user who knows better should extend the signal and crop -- is the correct one, and I am glad it is on the thread now.

Two things I want to note, since they are the newest code and were not reviewed:

`1f78d46d` removes the offset-zero invariant rather than documenting it, which is better than what I suggested -- the empty-range case is now handled where it arises instead of resting on a guarantee established two functions away. I brute-forced the result against a direct weighted sum over 400 random stencils including fully one-sided ones: exact agreement, and the previously reported `OverflowError` inputs now all raise the intended `ValueError`.

The NaN warning is a good compromise. One consequence worth knowing: it fires per call, so a loop over spectra that legitimately contain NaN will be noisy. Fine to leave until someone complains.

Only the docstring comments below are left, and neither should hold up the merge.

PR #719, opened 2026-08-06
test: double normalize in rotation creation to reduce flakiness

jokasimr:
Fixes #686

### DIFF
--- a/tests/conversion/beamline_conversions_test.py
+++ b/tests/conversion/beamline_conversions_test.py
@@ -26,6 +26,8 @@ def rotations() -> st.SearchStrategy[sc.Variable]:
     return (
         npst.arrays(dtype=np.float64, shape=4, elements=coefficients)
         .filter(lambda q: np.linalg.norm(q) != 0)
+        .map(lambda q: q / np.linalg.norm(q))
+        # Double normalization is intentional
         .map(lambda q: sc.spatial.rotation(value=q / np.linalg.norm(q)))
     )
 
@@ -213,7 +215,7 @@ def test_two_theta_invariant_under_rotation(rotation: sc.Variable) -> None:
         incident_beam=rotation * incident_beam, scattered_beam=rotation * scattered_beam
     )
 
-    sc.testing.assert_allclose(original, rotated, rtol=sc.scalar(1e-4))
+    sc.testing.assert_allclose(original, rotated, rtol=sc.scalar(1e-8))
 
 
 @given(incident_rotation=rotations(), scattered_rotation=rotations())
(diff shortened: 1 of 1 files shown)


tests/conversion/beamline_conversions_test.py
```diff
@@ -213,7 +213,7 @@ def test_two_theta_invariant_under_rotation(rotation: sc.Variable) -> None:
         incident_beam=rotation * incident_beam, scattered_beam=rotation * scattered_beam
     )
 
-    sc.testing.assert_allclose(original, rotated)
+    sc.testing.assert_allclose(original, rotated, rtol=sc.scalar(1e-4))
```
nvaytet: Strange, I would have hoped that the mathematical operation of a rotation would preserve angles better than 1e-4? Do yo know why it may not?

jokasimr: The issue was not really the rotation operation, it was the normalization step in the `rotations` strategy:

```python
def rotations() -> st.SearchStrategy[sc.Variable]:
    ...
    return (
        npst.arrays(dtype=np.float64, shape=4, elements=coefficients)
        .filter(lambda q: np.linalg.norm(q) != 0)
        .map(lambda q: sc.spatial.rotation(value=q / np.linalg.norm(q)))
```

when the entries of the provided `q` vector is extremely small the normalization can be inaccurate
```
In [1]: import numpy as np

In [2]: a = np.array([4.95652019e-160, 4.95652019e-160, 4.95652019e-160, 4.95652
      ⋮ 019e-160])

In [3]: a / np.linalg.norm(a)
Out[3]: array([0.50000175, 0.50000175, 0.50000175, 0.50000175])

```
and that seems to be what caused the test failures, probably because the rotation operation assumes the quaternions are normalized.

Normalizing twice when constructing the rotations allows lower tolerances to be used.

tests/conversion/beamline_conversions_test.py
```diff
@@ -26,6 +26,8 @@ def rotations() -> st.SearchStrategy[sc.Variable]:
     return (
         npst.arrays(dtype=np.float64, shape=4, elements=coefficients)
         .filter(lambda q: np.linalg.norm(q) != 0)
+        .map(lambda q: q / np.linalg.norm(q))
+        # Double normalization is intentional
```
nvaytet: Can you add a bit more of the description you posted in the PR, so the reader can understand why it is intentional? thanks

nvaytet (approved):

nvaytet:
Can you update the PR title? :pray:

PR #721, opened 2026-08-07
docs: deprecate convert

### LINKED ISSUE #9: Wrappers for coordinate transformations
We want to removed `scippneutron.convert` at some point. But we need to first ensure that we have user friendly interface around `transform_coords`. This likely requires wrappers around it for each technique / instrument.

We should start by refactoring all usages of `convert` to `transform_coords` and then compare workflows to come up with a common structure for a wrapper.

jokasimr:
Fixes https://github.com/scipp/ess/issues/9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,8 @@ addopts = """
 testpaths = "tests"
 filterwarnings = [
     "error",
+    # Legacy conversion tests intentionally exercise this deprecated API.
+    'ignore:scippneutron.convert is deprecated:scipp.VisibleDeprecationWarning',
     'ignore:You are running a "Debug" build of scipp:',
     # from ipywidgets; they are migrating away from ipykernel, this warning should go away
     'ignore:The `ipykernel.comm.Comm` class has been deprecated.:DeprecationWarning',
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -80,6 +80,8 @@ addopts = """
 testpaths = "tests"
 filterwarnings = [
     "error",
+    # Legacy conversion tests intentionally exercise this deprecated API.
+    'ignore:scippneutron.convert is deprecated:scipp.VisibleDeprecationWarning',
```
jl-wynen: Can you suppress this in `convert_test.py` instead of globally? That way, we check that we don't use `convert` elsewhere on accident.

jl-wynen (commented):
Works for me. @SimonHeybrock do you have an opinion? I seem to remember you being in favour of keeping `convert` in the past.

SimonHeybrock:
I suppose the question is: Does it cost us anything to keep it? If we deprecate it, should we still keep it around indefinitely (like some projects do)?

jokasimr:
> I suppose the question is: Does it cost us anything to keep it? If we deprecate it, should we still keep it around indefinitely (like some projects do)?

I'd say it costs very little. In principle the codebase will be more difficult to navigate and understand if we have tons of deprecated functionality around, but we're not anywhere near that.

Eventually I think we should remove this entirely. But imo we can easily wait 6 months or more to do so.

jl-wynen (approved):
I am fine with merging this.

If we deprecate `convert`, we should ultimately remove it. Otherwise we teach our users to ignore deprecation warnings.

